PR #34, opened 2022-03-22
dev create base class for processor tests

ekneg54:
this is the first pull request of a bigger refactoring story to ensure, that all processors are able to accept and process generic and specific rules.


ekneg54:
no need to merge.
can be deleted

PR #44, opened 2022-03-30
Add processing times to the prometheus exporter

dtrai2:
To account for the raising and falling of the metric values the
prometheus exporter was changed to use the Gauge instead of the
Counter metric. Furthermore the exporter now exports all metrics
separately for each process. The status logger configuration now
allows the user to choose if child processes should be aggregated
to single values or not.


ekneg54 (approved):
Fixed some linting and testing issues
looks good to me now

ppcad (approved):

PR #62, opened 2022-04-13
Remove no longer existing labeler exceptions from schema validator

ppcad:
Remove no longer existing labeler exceptions from schema validator

* The schema validation for the labeler should no longer cause import errors.


ekneg54 (changes requested):
the file has zero test coverage. And because of this, this didn't fail in the pipeline. 
I think it is the right time to add some simple first tests to ensure, that this won't happen again, isn't it?

ekneg54 (approved):

ppcad:
> @ppcad can you confirm that #65 reproduces the bug in pipeline? if so, we are now sure that this won't happen again and merge this and close #65

Yes it does. Thank you!

Issue #60, opened 2022-04-11
Refactor Processors to a common interface


ekneg54:
At stand now it is difficult to implement new features shared by all processors as seen in the #35 . As an developer I want to be able to implement such features in a more straight forward process without touching all the written code at once again. We should refactor the processors to a common interface class `Processor`.  And in this Issue we should also address the refactoring of the many factories for each processor to one factory for all processors. In this way we can guarantee that all processors behave the same and it should be easier to implement new processors. 

one first shot could be with `Normalizer` and `Pseudonymizer` as example:

```mermaid
classDiagram
    Processor <-- Normalizer : implements
    Processor <-- Pseudonymizer : implements
    dict --> ProcessorConfiguration : inherits
    Processor *-- ProcessorConfiguration
    ProcessorConfiguration <-- NormalizerConfiguration : implements
    ProcessorConfiguration <-- PseudonymizerConfiguration : implements
    ProcessorTestCase <-- NormalizerTestCase : implements
    ProcessorTestCase <-- PseudonymizerTestCase : implements
    class Processor{
        <<interface>>
        +String name
        +ProcessorConfiguration configuration
        +String describe()
        +add_rules_from_directories()
        +process()
        +apply_rules()*
    }
    class Normalizer{
        +apply_rules()
    }

    class Pseudonymizer{
        +apply_rules()
    }

    class ProcessorConfiguration{
        +__init__(dict)
        +list mandatory_fields()*
        +is_valid()
    }

    class NormalizerConfiguration{
        +list mandatory_fields()
    }

    class PseudonymizerConfiguration{
        +list mandatory_fields()
    }

    class ProcessorFactory{
        -_check_configuration(configuration)
        +Normalizer create_normalizer(name, configuration, logger)$
        +Pseudonymizer create_pseudonymizer(name, configuraiton, logger)$
    }

    class TestProcessorFactory{
        +test_check_configuration()
        +test_create_normalizer()
        +test_create_pseudonymizer()
    }

    class ProcessorTestCase{
        +test_describe()
        +test_add_rules_from_diretories()
        +test_process()
        +test_apply_rules()*
    }

    class NormalizerTestCase{
        +test_apply_rules()
    }

    class PseudonymizerTestCase{
        +test_apply_rules()
    }

```

ekneg54:
in addition we are able to enforce that generic and specific rules are processed in the same order for all processors.

PR #76, opened 2022-04-26
Refactor processor factories to use config check

dtrai2:
The processors datetime extractor, domain label extractor and
labeler weren't using the check_common_configuration method.
This is aligned in this pull request.


ppcad (approved):

ekneg54 (approved):
we should consider to refactor these code duplication to one place. this is addressed in  https://github.com/fkie-cad/Logprep/issues/60

PR #77, opened 2022-04-26
Fix init of the generic adder rule trees

dtrai2:
The tree_config path was given as a positional argument resulting
in being interpreted as a Node instead of the proper config path

### DIFF
--- a/logprep/processor/generic_adder/processor.py
+++ b/logprep/processor/generic_adder/processor.py
@@ -46,8 +46,8 @@ class GenericAdder(RuleBasedProcessor):
         generic_rules_dirs = configuration.get("generic_rules")
         super().__init__(name, tree_config, logger)
         self.ps = ProcessorStats()
-        self._generic_tree = RuleTree(tree_config)
-        self._specific_tree = RuleTree(tree_config)
+        self._generic_tree = RuleTree(config_path=tree_config)
+        self._specific_tree = RuleTree(config_path=tree_config)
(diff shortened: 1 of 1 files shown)


logprep/processor/generic_adder/processor.py
```diff
@@ -46,8 +46,8 @@ def __init__(self, name: str, configuration: dict, logger: Logger):
         generic_rules_dirs = configuration.get("generic_rules")
         super().__init__(name, tree_config, logger)
         self.ps = ProcessorStats()
-        self._generic_tree = RuleTree(tree_config)
-        self._specific_tree = RuleTree(tree_config)
+        self._generic_tree = RuleTree(config_path=tree_config)
+        self._specific_tree = RuleTree(config_path=tree_config)
```
ekneg54, line 50:
```suggestion
```
as you can see [here ](https://github.com/fkie-cad/Logprep/blob/b695bb0e9872b9ea14f3ffa71a1c186e1991e677/logprep/processor/base/processor.py#L128)it is set in the base class and can be deleted

ekneg54 (changes requested):

logprep/processor/generic_adder/processor.py
dtrai2: ah even better, thanks!

ppcad (approved):

ekneg54 (approved):

PR #85, opened 2022-05-02
Fix auto tester without rules

ppcad:
Fix auto tester without rules


ekneg54 (changes requested):
This file has zero test coverage.
In my opinion we should add some tests to ensure, that this does not happen again.

ppcad:
I added basic tests for the test coverage in the auto tester, since this was what caused the crash.
I would like to add further tests to cover other parts of the auto tester in another branch.

ppcad:
Added some more tests and resolved merge conflicts.

ekneg54 (approved):

PR #86, opened 2022-05-03
Add python 3.7 to 3.9 support

dtrai2:
Adjusts python requirements to allow for higher versions,
while still maintaining support for python 3.6. 
With that update also the tox environments such that the tests
can be run against all python versions. And also update the
github actions in order to continuously test logprep against the 
different versions.


ekneg54 (approved):
looks good to me

ppcad (approved):

PR #87, opened 2022-05-04
Fix quickstart config and add regression tests

ekneg54:
Ensure that the quickstart config is valid


quickstart/exampledata/config/pipeline.yml
```diff
@@ -3,47 +3,51 @@ timeout: 0.1
...
-     rules:
-       - quickstart/exampledata/rules/pre_detector/rules
-     pre_detector_topic: sre
-     tree_config: quickstart/exampledata/rules/pre_detector/tree_config.json
-     alert_ip_list: quickstart/exampledata/rules/pre_detector/alert_ips.yml
+      type: pre_detector
+      specific_rules:
+        - quickstart/exampledata/rules/pre_detector/rules
+      generic_rules:
+        - quickstart/exampledata/rules/pre_detector/rules
```
urdit52, line 32: I think that in the predetector, passing identical sets of rules as generic and specific will lead to duplicate detection results from separate trees. This will not cause duplicate events in the pre_det-topic, but it might generate duplicate elements in the list of detection results per processed event.

urdit52 (changes requested):
As mentioned, we should not pass identical rule sets as generic and specific rules in our quickstart pipeline. Could you please add separate generic and specific test rules for the labeller and  predetector?

urdit52 (approved):
Looks good to me.

dtrai2 (approved):

PR #90, opened 2022-05-05
Fix init of prometheus multiprocesscollector

dtrai2:
The multiprocessorcollector was initialized before the directory was
created. That lead to an error if the directory didn't exist already.
Fixed it by switching the statements and adding more tests for that.

### DIFF
--- a/tests/unit/util/test_prometheus_exporter.py
+++ b/tests/unit/util/test_prometheus_exporter.py
@@ -1,6 +1,9 @@
 # pylint: disable=missing-docstring
 # pylint: disable=protected-access
+# pylint: disable=attribute-defined-outside-init
+import os
 from logging import getLogger
+from unittest import mock
 
 from prometheus_client import Gauge, Info, REGISTRY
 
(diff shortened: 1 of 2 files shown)


tests/unit/util/test_prometheus_exporter.py
```diff
@@ -10,9 +15,7 @@
 class TestPrometheusStatsExporter:
     def setup_method(self):
         REGISTRY.__init__()
-
-    def test_correct_setup(self):
-        status_logger_config = {
+        self.status_logger_config = {
```
ekneg54, line 18: this is shared for all tests, isn't it? if yes -> why?

dtrai2: The status_logger_config is used for almost all tests. I consider this now as the base configuration that is used in every test, which focus isn't the configuration. In test cases where the configuration is part of the actual test I configured a separate status_logger_config object.

ppcad (approved):

ekneg54 (approved):

PR #93, opened 2022-05-12
Downgrade hyperscan to 0.1.5 and install only for x86_64 linux

ppcad:
Hyperscan >0.1.5 requires to install hyperscan libraries, but 0.1.5 does not seem run for python >3.8 despite what is said in the requirements.
Furthermore, it is compiled for x86_64 linux.


urdit52 (approved):
I verified that this works as intended on different systems. LGTM.

ekneg54 (approved):

Issue #97, opened 2022-05-16
Collect error messages on if the config is incorrect


ppcad:
**Is your feature request related to a problem? Please describe.**
An incorrect configuration makes logprep fail, but it returns one error message at a time instead of showing everything that is wrong with the configuration.

**Describe the solution you'd like**
Collect error messages and give the user feedback of all incorrect configurations.

**Describe alternatives you've considered**
Fixing the configuration one error at a time is possible, but can be time consuming in certain situations.

Issue #96, opened 2022-05-16
Logprep via pex fails silently if the config is incorrect


ppcad:
Running logprep via a pex file fails silently if the config is incorrect.

**Expected behavior**
A descriptive error message is thrown if the config is incorrect and logprep is started via pex, like it would be without pex.

**Current behavior**
Logprep via pex fails silently if the config is incorrect.

**Steps to reproduce**
Create an incorrect logprep config start via pex. (i.e. remove the type from processor).
There will be no output and logprep will close.

**Environment**

Logprep version: b5c3439e87fd2461418986c2ac82209f259e4e69
Using pex

PR #98, opened 2022-05-16
Add config error handling for pex and collection of exceptions

ppcad:
For issues #96 and #97. Makes logprep print out feedback on configuration errors even if it is executed via pex. Collects multiple errors before printing them.


dtrai2 (changes requested):
I wrote a few small comments for you to check. Maybe you could also add one test that checks the logger output for more than one exception. That way we can be sure that the behavior also always stays the same.

dtrai2 (approved):
Thanks for the changes!

PR #100, opened 2022-05-17
Add semgrep rule repository to pipeline

dtrai2:


ekneg54 (changes requested):
please add one python change to pipeline run to see if the rule run

PR #106, opened 2022-05-23
Delete do nothing processor

ekneg54:
removal of `donothing` processor because it is only used in tests but has to be maintained in future refactoring which should be unnecessary. Replaced by using mock objects


ekneg54:
closed because of manual merge

PR #113, opened 2022-06-04
Dev refactor domain resolver

ekneg54:
Rewrite `DomainResolver` to make it more readable and increase test coverage

### DIFF
--- a/logprep/processor/domain_resolver/processor.py
+++ b/logprep/processor/domain_resolver/processor.py
@@ -139,65 +139,63 @@ class DomainResolver(Processor):
         return self._config.timeout
 
     def _apply_rules(self, event, rule):
-        # TODO rewrite and tests this method
         domain_or_url = rule.source_url_or_domain
-        # new variable: output field
         output_field = rule.output_field
         domain_or_url_str = self._get_dotted_field_value(event, domain_or_url)
-        if domain_or_url_str:
-            domain = self._tld_extractor(domain_or_url_str).fqdn
-            if domain:
-                self.ps.increment_nested(self.name, "total_urls")
-                if self._config.cache_enabled:
-                    try:
-                        hash_string = self._hasher.hash_str(domain, salt=self._config.hash_salt)
-                        requires_storing = self._cache.requires_storing(hash_string)
-                        if requires_storing:
-                            result = self._thread_pool.apply_async(socket.gethostbyname, (domain,))
-                            resolved_ip = result.get(timeout=self._config.timeout)
-                            if len(self._domain_ip_map) >= len(self._cache):
-                                first_hash = next(iter(self._cache.keys()))
-                                del self._domain_ip_map[first_hash]
-                            self._domain_ip_map[hash_string] = resolved_ip
-                            self.ps.increment_nested(self.name, "resolved_new")
+        if not domain_or_url_str:
+            return
+        domain = self._tld_extractor(domain_or_url_str).fqdn
+        if not domain:
+            return
+        self.ps.increment_nested(self.name, "total_urls")
+        if self._config.cache_enabled:
+            try:
+                hash_string = self._hasher.hash_str(domain, salt=self._config.hash_salt)
+                requires_storing = self._cache.requires_storing(hash_string)
+                if requires_storing:
+                    result = self._thread_pool.apply_async(socket.gethostbyname, (domain,))
+                    resolved_ip = result.get(timeout=self._config.timeout)
+                    if len(self._domain_ip_map) >= len(self._cache):
+                        first_hash = next(iter(self._cache.keys()))
+                        del self._domain_ip_map[first_hash]
+                    self._domain_ip_map[hash_string] = resolved_ip
+                    self.ps.increment_nested(self.name, "resolved_new")
 
-                        if self._config.debug_cache:
-                            event["resolved_ip_debug"] = {}
-                            event_dbg = event["resolved_ip_debug"]
-                            if requires_storing:
-                                event_dbg["obtained_from_cache"] = False
-                            else:
-                                event_dbg["obtained_from_cache"] = True
-                            event_dbg["cache_size"] = len(self._domain_ip_map.keys())
+                if self._config.debug_cache:
+                    event["resolved_ip_debug"] = {}
+                    event_dbg = event["resolved_ip_debug"]
+                    if requires_storing:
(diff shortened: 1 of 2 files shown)


logprep/processor/domain_resolver/processor.py
```diff
@@ -54,7 +54,7 @@ class DomainResolverError(BaseException):
     """Base class for DomainResolver related exceptions."""
 
     def __init__(self, name: str, message: str):
-        super().__init__(f"DomainResolver ({name}): {message}")
+        super().__init__(f"DomainResolver ({name}): {message}")  # pragma: no cover
```
dtrai2, line 57: Why are we ignoring this in the tests? Same as line 187. We could add maybe a test which mocks the thread_pool and has an UnicodeError as an 'side_effect'.

logprep/processor/domain_resolver/processor.py
```diff
@@ -134,70 +134,64 @@ def _tld_extractor(self):
...
+        if not domain_or_url_str:
+            return
+        domain = self._tld_extractor(domain_or_url_str).fqdn
+        if not domain:
+            return
+        self.ps.increment_nested(self.name, "total_urls")
+        if self._config.cache_enabled:
+            hash_string = self._hasher.hash_str(domain, salt=self._config.hash_salt)
+            requires_storing = self._cache.requires_storing(hash_string)
+            resolved_ip = None
```
dtrai2, line 150: This line is not needed as the value is set in the if and else block.
```suggestion
```

logprep/processor/domain_resolver/processor.py
```diff
@@ -134,70 +134,64 @@ def _tld_extractor(self):
...
+            if requires_storing:
+                resolved_ip = self._resolve_ip(domain, hash_string)
+                self._domain_ip_map.update({hash_string: resolved_ip})
```
dtrai2, line 153: Why are we storing the resolved ips in a separate dict? Isn't this the job of the Cache? After all, the cache is an OrderedDict so we could use this to store the values and delete the `_domain_ip_map` entirely.

tests/unit/processor/domain_resolver/test_domain_resolver.py
```diff
@@ -28,21 +27,159 @@ class TestDomainResolver(BaseProcessorTestCase):
...
+            "description": "",
+        }
+        self._load_specific_rule(rule)
+        document = {"fqdn": "google.de"}
+        expected = {"fqdn": "google.de", "resolved_ip": "1.2.3.4"}
+        self.object.process(document)
+        assert document == expected
+
+    @mock.patch("socket.gethostbyname", return_value="1.2.3.4")
+    def test_domain_to_ip_resolved_and_added_from_cache(self, _):
```
dtrai2, line 44: How does this test know that it should resolve the domain from the cache? The test is actually exactly the same as the previous one `test_domain_to_ip_resolved_and_added`. Does this test maybe miss a second `self.object.process(document)` call?

tests/unit/processor/domain_resolver/test_domain_resolver.py
```diff
@@ -28,21 +27,159 @@ class TestDomainResolver(BaseProcessorTestCase):
...
+            "description": "",
+        }
+        self._load_specific_rule(rule)
+        document = {"url": "https://www.google.de/something"}
+        expected = {"url": "https://www.google.de/something", "resolved_ip": "1.2.3.4"}
+        self.object.process(document)
+        assert document == expected
 
-        monkeypatch.setattr(socket, "gethostbyname", mockreturn)
+    def test_domain_ip_map_greater_cache(self):
```
dtrai2, line 69: This test doesn't have any assertion anymore. What is it testing for or what behavior is expected?

tests/unit/processor/domain_resolver/test_domain_resolver.py
```diff
@@ -28,21 +27,159 @@ class TestDomainResolver(BaseProcessorTestCase):
...
+            "domain_resolver": {"source_url_or_domain": "url"},
+            "description": "",
+        }
+        self._load_specific_rule(rule)
+        document = {"url": "https://www.google.de/something"}
+        expected = {"url": "https://www.google.de/something", "resolved_ip": "1.2.3.4"}
+        self.object.process(document)
+        assert document == expected
+
+    # @pytest.mark.skipif(not exists(TLD_LIST.split("file://")[-1]), reason="Tld-list required.")
```
dtrai2, line 156: The comment isn't needed anymore.
```suggestion
```

dtrai2 (changes requested):

tests/unit/processor/domain_resolver/test_domain_resolver.py
ekneg54: rewrote test

tests/unit/processor/domain_resolver/test_domain_resolver.py
ekneg54: done

logprep/processor/domain_resolver/processor.py
ekneg54: no... in the cache there are only the hashes cause the cache is used in other modules too.

logprep/processor/domain_resolver/processor.py
ekneg54: I would prefer to delete it, because I have no clue, where it is raised in line 187

ekneg54: deleted

logprep/processor/domain_resolver/processor.py
```diff
@@ -134,70 +134,64 @@ def _tld_extractor(self):
...
-                if self._config.cache_enabled:
-                    try:
-                        hash_string = self._hasher.hash_str(domain, salt=self._config.hash_salt)
-                        requires_storing = self._cache.requires_storing(hash_string)
-                        if requires_storing:
-                            result = self._thread_pool.apply_async(socket.gethostbyname, (domain,))
-                            resolved_ip = result.get(timeout=self._config.timeout)
-                            if len(self._domain_ip_map) >= len(self._cache):
-                                first_hash = next(iter(self._cache.keys()))
-                                del self._domain_ip_map[first_hash]
```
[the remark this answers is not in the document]
ekneg54: removed

dtrai2 (approved):

Issue #116, opened 2022-06-20
Clusterer slowed down after refactoring


ppcad:
The clusterer slowed down extremely after refactoring.
The slowdown increases with the number of rules and the amount of processed documents.
Thus, it might not appear to slow down if only a few rules and documents are being used.

**Expected behavior**
The clusterer processing time is similar to other processors.

**Current behavior**
The clusterer is extremely slow. The more rules it has the slower it becomes.

**Steps to reproduce**
Run the clusterer with many rules and compare it to other processors.

**Environment**

Logprep version: c83199743c362a2e808385f80c51d5160742abbf
Python version: 3.6-3.9

PR #117, opened 2022-06-20
Ensure matching rules are cleared for clusterer on processing

ppcad:
See #116

### DIFF
--- a/logprep/processor/clusterer/processor.py
+++ b/logprep/processor/clusterer/processor.py
@@ -59,6 +59,7 @@ class Clusterer(Processor):
         self.has_custom_tests = True
 
     def process(self, event: dict):
+        self.matching_rules = []
         super().process(event)
         if self._is_clusterable(event):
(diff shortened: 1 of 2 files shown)


logprep/processor/clusterer/processor.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, name: str, configuration: Processor.Config, logger: Logger):
         self.has_custom_tests = True
 
     def process(self, event: dict):
+        self.matching_rules = []
```
ekneg54, line 62: it is declared as attribute and set in `__init__`. Am I right if I say, that one of them is redundant?

ppcad: The one in process must exist. The one in __init__ has been kept for readability and better unit tests.

ekneg54 (approved):

urdit52 (approved):
looks good to me.

PR #122, opened 2022-06-21
remove wrong if condition in github action

dtrai2:
Also remove debug information as the github.ref
really does have the correct tag of the published
release.


dtrai2:
@ekneg54 I know it's only one commit but still please merge without squash because of the set tag.

ekneg54 (approved):

PR #126, opened 2022-06-22
Use abstract dependencies for distribution

dtrai2:
To simplify the installation via `pip install logprep` only the abstract 
requirements should be given. That way pip can itself look, what 
versions suit best to the system of the installing client.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,12 +5,9 @@ from setuptools import setup, find_packages
 
 import versioneer
 
-with open("requirements.txt", encoding="utf-8") as f:
+with open("requirements.in", encoding="utf-8") as f:
     requirements = f.read().splitlines()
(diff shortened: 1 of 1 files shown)


ekneg54 (commented):
A little bit more control of the installed dependencies should be safer, shouldn't it?

dtrai2:
> A little bit more control of the installed dependencies should be safer, shouldn't it?

I actually read about it, that the requirements.txt should be very specific to be able to produce the same venv for all developers, but for distribution abstract dependencies are fine. Most clients have slightly different setups and that way pip can try to find the best fit for that client. 

But we can actually get rid of this pull-request, if the current `pip install logprep` works without any issues. I opened this pull request as a fix or debug, but right now it seems to work fine without it as well.

ekneg54 (approved):

PR #129, opened 2022-06-22
some inspiration for dev-gen-mysql

ekneg54:
only here to get some inspiration on using `pymysql`.
the sql injection topic is not solved with this, but as I investigated it should not be a problem because parameters are not strings.

but on my journey I discovered that `pymysql` has a `DictCursor` which should simplify things and make them much more readable. Give it a try.

please close this, if not needed


ekneg54:
closed, because it won`t be merged.

Issue #132, opened 2022-06-23
Pseudonymizer has error with tld list


ppcad:
The pseudonymizer crashes if a tld list is configured.

**Expected behavior**
The pseudonymizer loads the tld list.

**Current behavior**
The pseudonymizer crashes if a tld list is configured.

**Steps to reproduce**

1. Configure a tld list for the pseudonymizer
2. Run logprep
3. The pseudonymizer will crash, because the variable self._config.tld_list does not exist

**Environment**

Logprep version: e67d0ed3bc23fdf53a448ff97fe75b1ee4ed6aa7
Python version: 3.6-3.9

**Possible implementation**
Rename the variable self._config.tld_list to self._config.tld_lists.

PR #133, opened 2022-06-23
Fix tld variable in pseudonymizer

ppcad:
See #132

### DIFF
--- a/logprep/processor/pseudonymizer/processor.py
+++ b/logprep/processor/pseudonymizer/processor.py
@@ -24,7 +24,8 @@ Example
         regex_mapping: /path/to/regex_mapping.json
         max_cached_pseudonyms: 1000000
         max_caching_days: 1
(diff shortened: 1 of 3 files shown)


dtrai2 (changes requested):
Could you also check the lines 27 and 119 please, seems there are also occurrences of `tld_list`. Funny thing is also, this file is completely tested, except two lines. And the initialization of the tld extractor with tld_list is one of them. If you have time, could you check if you could write a quick test for that so we cover this as well?

ppcad:
> Could you also check the lines 27 and 119 please, seems there are also occurrences of `tld_list`. Funny thing is also, this file is completely tested, except two lines. And the initialization of the tld extractor with tld_list is one of them. If you have time, could you check if you could write a quick test for that so we cover this as well?

I have changed it.

dtrai2 (approved):

Issue #131, opened 2022-06-23
Auto tester does not work for some processors after refactoring


ppcad:
The auto tester does not work for the pseudonymizer and the list comparison processor after some recent refactoring.
This could also affect future processors.

**Expected behavior**
The auto tester works for all processors. Rules are being initialized correctly.

**Current behavior**
The auto tester does not work for the pseudonymizer and the list comparison processor. Rules are not being initialized correctly.

**Steps to reproduce**

1. Create rule passing auto tests for the pseudonymizer (with regex list keywords) or for the list comparison processor
2. Run the auto tests
3. The tests will fail

**Environment**

Logprep version: e67d0ed3bc23fdf53a448ff97fe75b1ee4ed6aa7
Python version: 3.6-3.9

**Possible solution**
Perform additional initialization in the auto tester.

PR #134, opened 2022-06-23
Handle processor specific rule initialization in auto tester

ppcad:
See #131

### DIFF
--- a/logprep/processor/list_comparison/processor.py
+++ b/logprep/processor/list_comparison/processor.py
@@ -65,6 +65,9 @@ class ListComparison(Processor):
 
     def __init__(self, name: str, configuration: "Processor.Config", logger: Logger):
         super().__init__(name, configuration, logger)
+        self._init_rules_list_comparison()
+
+    def _init_rules_list_comparison(self):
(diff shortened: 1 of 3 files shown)


dtrai2 (commented):
The auto-rule tester didn't get lot's of attention from us lately, especially after all the refactoring. Maybe it is worth having a look at trying to initialize the processors in the auto-rule tester with the new processor factory? I assume this would make the processor specific setup unnecessary. 
Sadly the auto-rule tester is also not really tested (coverage of 39%), maybe you could add one test for it, so we can improve it step by step every time someone is touching that file.

ppcad:
> The auto-rule tester didn't get lot's of attention from us lately, especially after all the refactoring. Maybe it is worth having a look at trying to initialize the processors in the auto-rule tester with the new processor factory? I assume this would make the processor specific setup unnecessary. Sadly the auto-rule tester is also not really tested (coverage of 39%), maybe you could add one test for it, so we can improve it step by step every time someone is touching that file.

I will write a test, but I would not like to rework the auto tester right now, since the tester is broken and needs to be fixed asap.

dtrai2 (approved):

PR #136, opened 2022-07-01
Print logprep version via CLI

dtrai2:
Adds functionality to print the logprep version as well as 
the configuration version. The configuration version is
only complete if a configuration file was found and also 
has the version key.


CHANGELOG.md
```diff
@@ -17,6 +19,11 @@
 * Fix processor initialization in auto rule tester
 * Fix generation of RST-Docs
 
+### Breaking
+
+* The configuration argument `--config` is not a positional argument anymore. 
+With that it has to be used when passing a configuration to logprep
```
urdit52, line 25: This is misleading. How about:
```suggestion
This makes it necessary to explicitly use the argument --config when a config is to be passed to logprep.
```

tests/unit/test_run_logprep.py
```diff
@@ -34,8 +45,61 @@ def test_quickstart_rules_are_valid(self):
...
             "--disable-logging",
             "--validate-rules",
+            "--config",
             "quickstart/exampledata/config/pipeline.yml",
         ]
         with pytest.raises(SystemExit) as e_info:
             run_logprep.main()
         assert e_info.value.code == 0
+
+    def test_version_arg_prints_logprep(self, capsys):
```
urdit52, line 55: this also tests for the case that no config version is specified, which should be reflected in the test name.

urdit52 (changes requested):
I have requested and suggested 2 minor things.

urdit52 (approved):
thanks for fixing this.

ppcad:
Useful changes! I also like that you can version the config now.

If you change the config parameter to be like below, then you could  still use _--version_ without having to specify a config and it would not be e breaking change.

```
argument_parser.add_argument(
        "config", nargs='?',
        help=f"Path to configuration file, if not given then "
        f"the default path '{DEFAULT_LOCATION_CONFIG}' is used",
        default=DEFAULT_LOCATION_CONFIG,
    )
```

You would also have to change some of the feedback messages that were written with _--config_ in mind.

What do you think? Am I missing something? I would like to avoid breaking changes if possible.

ppcad (approved):
Thanks!

Issue #138, opened 2022-07-06
MySQL generic adder crash on missing config but existing rule


ppcad:
The generic adder crashes if the sql_config is disabled/commented out and there exist SQL rules for the generic adder.
It does not properly detect that sql was deactivated and crashes during runtime.

**Expected behavior**
The generic adder does not crash if the sql_config is disabled/commented out and there exist SQL rules for the generic adder.

**Current behavior**
The generic adder crashes if the sql_config is disabled/commented out and there exist SQL rules for the generic adder.

**Steps to reproduce**

1. Configure the genric adder to not use sql.
2. Add a SQL rule to the rule path.
3. Run logprep
4. Pass a document to logprep that would trigger a rule with SQL
5. Watch it crash during runtime with a message like "'member_descriptor' object has no attribute 'get'"

**Environment**

Logprep version: d4b03ff4238905de7907cb82743cbc314e8e79b0
Python version: 3.6-3.9

PR #139, opened 2022-07-07
Fix crash from not initialized class variable in sql adder

ppcad:
Not initializing "_db_table" in the processor of the generic adder has led to some unexpected crashes if the SQL adder was disabled, but rules that utilize SQL where processed.

It was changed so that it's initialized with "None" again and it was changed to be an instance variable.
Furthermore, some unit tests that prevent this issue from happening again where added.

See #138.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 * Fix processor initialization in auto rule tester
 * Fix generation of RST-Docs
+* Fix crashes of the generic adder if sql-based rules are processed without having set the sql config 
 
 ### Breaking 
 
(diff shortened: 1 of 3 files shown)


ppcad:
> You can initialize it in **init**. There is no need to remove it from slots. If you do it in the class it will be a class variable and is possibly shared by all objects of that class.

The original idea for it was to be a class variable to reduce the amount of times a table is loaded into logprep.
However, I see that this can lead to problems with certain configurations.
I will change it as you suggest and think of ways to improve it later.

ekneg54 (approved):

dtrai2 (approved):

PR #149, opened 2022-07-15
Add logprep and configuration version to event

dtrai2:
This enables the pipeline to automatically add version information
to all events independent of the pipeline configuration.
This pull request also refactors the initialization of pipeline and 
multiprocessing pipeline objects. Furthermore it introduces a 
`preprocessing` config field to the connector consumer options. 
The already existing hmac options were moved under this new 
field while maintaining backwards capabilities. Meaning the hmac
can now be configured in two places. The old place under the
field `consumer` will lead to a deprecation warning, indicating a 
removal in the next major release. 

In an additional pull request the hmac implementation should also
move away from the connector and into the pipeline such that it is
an connector independent feature.

### DIFF
--- a/doc/source/user_manual/configuration/connector.rst
+++ b/doc/source/user_manual/configuration/connector.rst
@@ -43,10 +43,20 @@ This object configures how log messages are being fetched from Kafka.
 - **offset_reset_policy**: Corresponds to the Kafka configuration parameter `auto.offset.reset <https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md>`_. This parameter influences from which offset the Kafka consumer starts to fetch log messages from an assigned partition. The values *latest/earliest/none* are possible. With a value of *none* Logprep must manage the offset by itself. However, this is not supported by Logprep, since it is not relevant for our use-case. If the value is set to *latest/largest*, the Kafka consumer starts by reading the newest log messages of a partition if a valid offset is missing. Thus, old log messages from that partition will not be processed. This setting can therefore lead to a loss of log messages. A value of *earliest/smallest* causes the Kafka consumer to read all log messages from a partition, which can lead to a duplication of log messages. Currently, the deprecated value *smallest* is used, which should be later changed to *earliest*. The default value of librdkafka is *largest*.
 - **enable_auto_offset_store**: Corresponds to the Kafka configuration parameter `enable.auto.offset.store <https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md>`_. This parameter defines if the offset is automatically updated in memory. Disabling this allows Logprep to update the offset more accurately. The default value in librdkafka it is *true*.
 
-Additionally to the previous configurations it is possible to automatically attach an HMAC to an incoming log message.
-If it is required to do so the following options should be appended to the general consumer options under a new
-field :code:`hmac`. This field is completely optional and can also be omitted if no hmac is needed. An example with
-hmac configuration is given at the end of this page.
+preprocessing
+^^^^^^^^^^^^^
+
+It is possible to activate simple preprocessors in the consumer.
+The following will describe the available preprocessors and what they do.
+They each have to be defined under the key :code:`preprocessor`.
+
+**hmac**
+
+If required it is possible to automatically attach an HMAC to incoming log messages.
+To activate this preprocessor the following options should be appended to the preprocessor options
+under a new field :code:`hmac`.
+This field is completely optional and can also be omitted if no hmac is needed.
+An example with hmac configuration is given at the end of this page.
 
 - **target**: Defines a field inside the log message which should be used for the hmac calculation. If the target field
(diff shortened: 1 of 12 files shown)


doc/source/user_manual/configuration/connector.rst
```diff
@@ -103,10 +140,12 @@ Example
         auto_commit: on
         session_timeout: 6000
         offset_reset_policy: smallest
-        hmac:
-          target: <RAW_MSG>
-          key: secret-key
-          output_field: Hmac
+        preprocessing:
```
ppcad, line 143: The consumer is specific to the confluent-kafka connector. Preprocessors will be useful for other inputs as well. Do you think placing it under 'consumer' is the best choice?

ppcad (changes requested):

doc/source/user_manual/configuration/connector.rst
dtrai2: I completely agree with you that's why I already implemented appending the version info in the pipeline.py instead of the confluent_kafka.py, which currently has the hmac functionality implemented. I just left the hmac calculation where it is for now as we were already talking about doing a refactoring of the inputs and outputs in general. If I would have already moved the hmac functionality away from the confluent_kafka then this merge request would have been much larger. 

I'd say we leave it as is for now an, discuss what the best solution for inputs, outputs and preprocessing is and then start working on how to implement it. Especially if we plan to add more preprocessors we should find a common interface for them before they get too messy.

dtrai2:
@ppcad I just moved the Improvements from the Generic Adder in the Changelog.md to the Next Release. The Release v3.1.0 is already 12 days old and I think the Generic Adder optimizations are not part yet of the 3.1.0 release.

ppcad (approved):

urdit52 (changes requested):
Review of the tests still needs to be done, here are a few change requests for the code.

logprep/connector/confluent_kafka.py
```diff
@@ -211,42 +215,62 @@ def set_option(self, new_options: dict):
...
-            self._add_hmac = True
+    def _validate_hmac_options(self, hmac_options):
+        valid_hmac_options_keys = set(
+            self._config.get("consumer", {}).get("preprocessing", {}).get("hmac", {}).keys()
+        )
+
+        new_hmac_options_keys = set(hmac_options)
+        unknown_options = new_hmac_options_keys.difference(valid_hmac_options_keys)
+        if unknown_options:
+            raise UnknownOptionError(f"Unknown Hmac Options: {unknown_options}")
```
urdit52, line 253: This seems to be untested.

logprep/connector/confluent_kafka.py
```diff
@@ -211,42 +215,62 @@ def set_option(self, new_options: dict):
             Raises if an option is invalid.
 
         """
+        # DEPRECATION: HMAC-Option: Remove this with next major version update, check also the
+        # self._config dict in this class's init method
+        consumer_options = new_options.get("consumer")
+        if consumer_options and "preprocessing" not in consumer_options:
+            consumer_options["preprocessing"] = {}
+        if new_options.get("consumer", {}).get("hmac"):
+            consumer_options["preprocessing"]["hmac"] = new_options["consumer"]["hmac"]
```
urdit52, line 224: please add tests to test_confluent_kafka.py

logprep/connector/confluent_kafka.py
```diff
@@ -211,42 +215,62 @@ def set_option(self, new_options: dict):
...
-                if not isinstance(config_value, str):
-                    raise InvalidConfigurationError(
-                        f"Hmac option '{key}' has wrong type: '{type(config_value)}', "
-                        f"expected 'str'"
-                    )
-                if len(config_value) == 0:
-                    raise InvalidConfigurationError(
-                        f"Hmac option '{key}' is empty: '{config_value}'"
-                    )
+    def update_default_configuration(self, default_options, user_options):
```
urdit52, line 233: This method is not called is not tested explicitly.

urdit52 (changes requested):
3 things concerning test coverage.

tests/unit/connector/test_confluent_kafka.py
```diff
@@ -702,3 +694,84 @@ def test_get_next_without_hmac(self):
...
+            "option": {"with": {"multiple": "layers", "foo": "bar"}},
+            "another": "option",
+        }
+        user_config = {}
+        config = deepcopy(TestConfluentKafkaFactory.valid_configuration)
+        kafka = ConfluentKafkaFactory.create_from_configuration(config)
+        new_config = kafka.update_default_configuration(default_config, user_config)
+        assert new_config == default_config
+
+    def test_update_default_configuration_raises_error_on_overwriting_all_subfields(self):
```
urdit52, line 750: This also seems to test for the type of field value to be replaced (as does the next test). Do we really need both?

urdit52 (changes requested):
Thank you. One minor thing regarding a single test.

PR #150, opened 2022-07-19
Dev generic adder sql optimizations

ppcad:
Make the generic adder write the SQL table to a file and load it from there instead of loading it 
from the database for every process of the multiprocessing pipeline.
Furthermore, only connect to the SQL database on checking if the database table has changed and the file is stale.
This reduces the SQL connections.
Before, there was permanently one connection per multiprocessing pipeline active and now there is only one connection per Logprep instance active when accessing the database.


logprep/processor/generic_adder/mysql_connector.py
```diff
@@ -47,25 +53,51 @@ def __init__(self, sql_config: dict, logger: Logger):
...
+        bool
+            True if a check should be performed, False otherwise.
+
+        """
+        check_change = time.time() - self._last_check >= self._db_check_interval
+        if check_change:
+            self._last_check = time.time()
+        return check_change
+
     def has_changed(self) -> bool:
```
urdit52, line 101: In my opinion, this function could be simplified. If the last checksum is None, it is also not equal to the current checksum and therefore should just return True. My suggestion would be to simplify in line with your doctoring by putting the False case first and in any other case return True and set the self._last_table_checksum = checksum.

logprep/processor/generic_adder/mysql_connector.py
```diff
@@ -78,16 +110,13 @@ def has_changed(self) -> bool:
...
-            checksum = self._get_checksum()
-            if self._last_table_checksum is None:
-                self._last_table_checksum = checksum
-                return True
-            if self._last_table_checksum == checksum:
-                return False
+        checksum = self._get_checksum()
+        if self._last_table_checksum is None:
+            self._last_table_checksum = checksum
             return True
```
urdit52, line 116:
```suggestion
        if self._last_table_checksum == checksum:
            return False
```

logprep/processor/generic_adder/mysql_connector.py
```diff
@@ -78,16 +110,13 @@ def has_changed(self) -> bool:
...
-        return False
+        if self._last_table_checksum == checksum:
+            return False
+        return True
```
urdit52, line 119:
```suggestion
        else:
            self._last_table_checksum = checksum
            return True
```

tests/unit/processor/generic_adder/test_generic_adder.py
```diff
@@ -270,13 +296,41 @@ class TestGenericAdderProcessorSQLWithoutAddedTarget(BaseProcessorTestCase):
...
+        stale_after_interval = self.object._check_if_file_not_exists_or_stale(now)
+        os.utime(db_file_path, (now, now))
+        stale_after_change = self.object._check_if_file_not_exists_or_stale(now)
+        os.remove(db_file_path)
+        stale_without_file = self.object._check_if_file_not_exists_or_stale(now)
+
+        assert stale_before_interval is False
+        assert stale_after_interval is True
+        assert stale_after_change is False
+        assert stale_without_file is True
```
urdit52, line 333: In my opinion, this test is not understandable at first glance which might warrant a doctoring even though it is a test.

urdit52 (changes requested):
Thank you for committing this.

I requested a few minor things.

tests/unit/processor/generic_adder/test_generic_adder.py
ppcad: I have split it into smaller tests so that it is easier to understand it.

logprep/processor/generic_adder/mysql_connector.py
ppcad: Thanks, I have changed it.

urdit52 (approved):
Thanks a lot for the changes.

PR #152, opened 2022-08-04
Fix selective extractor output

dtrai2:
The internal extracted list wasn't cleared between each event,
leading to duplication in the output of the processor. Now the
events are cleared such that only the result of the current event is
returned.


ekneg54 (approved):
good job

PR #153, opened 2022-08-04
Fix pipeline config attribute name

dtrai2:
The logprep configuration was overwritten by the config of the
multiprocessing Process. This python class BaseProcess also contains
an attribute that's called '_config'. This resulted in a naming
conflict. Renaming the logprep config to _logprep_config fixed this
issue.


ppcad (approved):
Thanks, it works!

ekneg54 (approved):
looks good to me

PR #154, opened 2022-08-05
Dev improve dry run

### LINKED ISSUE #130: DryRun with input type json results in KeyError 0
If the dry run is started with an input type `json` it results in an `KeyError: 0`.

**Expected behavior**
The dry-run should be executed on one json in the file. 

**Current behavior**
A stacktrace with the error message `KeyError: 0` is given.

**Steps to reproduce**
1. Delete all but one json in the file `quickstart/exampledata/input_logdata/test_input.jsonl`
2. (optional) Format the json into multiple lines
3. Start logprep with `logprep quickstart/exampledata/config/pipeline.yml --dry-run quickstart/exampledata/input_logdata/test_input.jsonl --dry-run-input-type json --dry-run-full-output`

**Environment**

Logprep version: 2.0.1
Python version: 3.9.13

dtrai2:
The issue #130 expected a bug which turned out to be no bug. 
The input file format was just wrong. This Pull-Request implements
an improvement such that the expected format from the issue is 
possible now too. This makes the usability of logprep's dry-run easier.

Before, the expected input for a single json still had to be in brackets `[`
and `]`. Now a single json can also be used inside a file without these.

Furthermore a few smoke tests for the dry run were added.

Closes #130


logprep/util/json_handling.py
```diff
@@ -84,11 +85,14 @@ def parse_json(json_path):
...
-        The dictionary representing the json.
+    list
+        A list of dictionary representing the json/jsons.
     """
     with open(json_path, "r", encoding="utf8") as json_file:
-        return json.load(json_file)
+        parsed_json = json.load(json_file)
+        if isinstance(parsed_json, dict):
+            parsed_json = [parsed_json]
+        return parsed_json
```
ekneg54, line 95: is see, that this is tested implicit but could you add some explicit tests for this unit please?

logprep/input/jsonl_input.py
```diff
@@ -1,9 +1,8 @@
 """This module contains a dummy input that can be used for testing purposes."""
```
ekneg54, line 1: if we compare this file with json_input.py we will see, that they only differ in one letter in line 23.
how about to merge the two input connectors into one connector? or alternativly use the magic of inheritance for this?

logprep/util/rule_dry_runner.py
```diff
@@ -22,8 +22,8 @@
 def get_test_output_multi(config_path, logger):
```
ekneg54, line 22: missing function or method docstring

logprep/util/rule_dry_runner.py
```diff
@@ -54,8 +54,8 @@ def get_patched_runner(config_path, logger):
 class DryRunner:
```
ekneg54, line 54: this is the dark side of logprep. We need to refactor this class. Could you spend some time to increase test coverage for the `run` method?

tests/unit/util/test_rule_dry_runner.py
```diff
@@ -0,0 +1,101 @@
...
+import json
+import logging
+import os
+
+from logprep.util.rule_dry_runner import DryRunner
+
+
+class TestRunLogprep:
+    def test_dry_run_accepts_json_as_input(self, tmp_path, capsys):
+        config = "tests/testdata/config/config-dry-run.yml"
```
ekneg54, line 12: could we get this config to this file to be aware what this config is without searching the config file?

ekneg54 (changes requested):

logprep/util/json_handling.py
dtrai2: done

logprep/input/jsonl_input.py
dtrai2: That is indeed a reasonable change that should be done. I don't see it as a part of this pull-request though. This pull-request started as a bug-fix and the resulting improvements here far exceed that initial goal already. Merging these two json input types would end in a refactoring inside the logprep connectors and as we are planning to do this anyways I suggest we do it when we touch the connectors itself.

logprep/util/rule_dry_runner.py
dtrai2: I started to do some refactorings here. It is still not perfect but I think its an improvement already. Similar to my other remark I would prefer though not to spend much more time on this as it started as a bugfix and wasn't intended as a huge refactoring.

ekneg54: good job, but 100% test coverage  before doing refactorings should be the better way.
I keep my fingers crossed 🤞

ekneg54 (approved):

PR #158, opened 2022-08-12
Fix preprocessor config validation

dtrai2:
The unit tests for the preprocessor which adds version information to
an event didn't cover the config validation. This pull-requests adds
acceptances tests to test a full message pass with the preprocessor
activated and fixes a config validation error.

### DIFF
--- a/logprep/framework/pipeline.py
+++ b/logprep/framework/pipeline.py
@@ -256,7 +256,7 @@ class Pipeline:
     def _preprocess_event(self, event):
         consumer_config = self._logprep_config.get("connector").get("consumer", {})
         preprocessing_config = consumer_config.get("preprocessing", {})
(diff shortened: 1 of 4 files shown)


logprep/framework/pipeline.py
```diff
@@ -256,7 +256,7 @@ def _retrieve_and_process_data(self):
     def _preprocess_event(self, event):
         consumer_config = self._logprep_config.get("connector").get("consumer", {})
         preprocessing_config = consumer_config.get("preprocessing", {})
-        if preprocessing_config.get("version_info_target_field"):
+        if preprocessing_config.get("version_info_target_field", "") != "":
```
urdit52, line 259:
```suggestion
        if preprocessing_config.get("version_info_target_field"):
```

urdit52 (changes requested):
one little detail.

PR #167, opened 2022-08-25
Refactor delete processor -> deleter

saegel:
This PR renames the `delete` Processor to `deleter` to be more consistent with the names of the other processors.


ekneg54:
you are right. it is indeed more consistent. But it is also a breaking change because all user have to rebuild their configs. I`m not sure it is worth it only for the pending `r`

logprep/processor/processor_registry.py
```diff
@@ -27,7 +27,7 @@ class ProcessorRegistry:
     mapping = {
         "clusterer": Clusterer,
         "datetime_extractor": DatetimeExtractor,
-        "delete": Delete,
+        "deleter": Deleter,
```
saegel, line 30: This might be slightly hacky but would adding an extra mapping for the delete class prevent the breaking change?

```suggestion
        "deleter": Deleter,
        "delete": Deleter,
```

dtrai2: Well we could do it the same way with the HMAC-Option. Meaning we can introduce the inconsistency in the mapping while at the same time marking the `delete` processor as deprecated, which will then be removed in the next mayor release. That way we can now introduce the change, but have one mayor update with all the breaking changes included.

ekneg54: lets do it

urdit52: agreed. however, we also need a changelog for this PR.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 * Removal of Deprecated Feature: HMAC-Options in the connector consumer options have to be 
 under the subkey `preprocessing`
 
+* Removal of Deprecated Feature: `delete`processor was renamed to `deleter`
```
dtrai2, line 8:
```suggestion
* Removal of Deprecated Feature: `delete` processor was renamed to `deleter`
```

logprep/processor/processor_registry.py
```diff
@@ -58,4 +61,9 @@ def get_processor_class(cls, processor_type):
         _type_
             _description_
         """
+        if processor_type == "delete":
+            warnings.warn(
+                "delete processor is deprecated and will be removed in a future release. Please use deleter instead.",
+                DeprecationWarning,
+            )
```
dtrai2, line 68: Does this warning actually work? When I tried implementing the first deprecation the warning was never thrown, so I used the regular logging. Maybe you can write a test to check if it actually works.

saegel: I found this bit in the [Python Docs](https://docs.python.org/3/library/warnings.html#warning-categories):

DeprecationWarning:
> Base category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in __main__).

This might explain why deprecation warnings sometimes do not show up.
I now changed the warning type to `UserWarning` which should definitely show up and also added a test.

UserWarnings also make sense in this case, since it's mainly users of Logprep we want to warn about the breaking change.

dtrai2: Okay thanks, that's good to know. I will try this out with the next deprecation I'll write.

dtrai2 (approved):

tests/unit/processor/delete/test_delete.py
```diff
@@ -6,9 +6,9 @@
 class TestDelete(BaseProcessorTestCase):
```
ekneg54, line 6:
```suggestion
class TestDeleter(BaseProcessorTestCase):
```
the file module be `.deleter.test_deleter`

tests/unit/processor/test_processor_registry.py
```diff
@@ -0,0 +1,13 @@
...
+
+def test_get_processor_class_throws_deprecation_warning():
+    with warnings.catch_warnings(record=True) as w:
+        ProcessorRegistry.get_processor_class("delete")
+        assert len(w) == 1
+        assert isinstance(w[0], warnings.WarningMessage)
+        assert (
+            str(w[0].message)
+            == "delete processor is deprecated and will be removed in a future release. Please use deleter instead."
+        )
```
ekneg54, line 13:
```suggestion
    with warnings.catch_warnings(record=True) as warning_messages:
        ProcessorRegistry.get_processor_class("delete")
        assert len(warning_messages) == 1
        assert isinstance(warning_messages[0], warnings.WarningMessage)
        assert (
            str(warning_messages[0].message)
            == "delete processor is deprecated and will be removed in a future release. Please use deleter instead."
        )
```

ekneg54 (changes requested):
please rename:

`tests.unit.processor.delete.test_delete` to `tests.unit.processor.deleter.test_deleter`.
`tests.unit.processor.delete.test_delete_rule` to `tests.unit.processor.deleter.test_deleter_rule`

ekneg54:
Please resolve merge conflicts so we could merge it

ekneg54 (approved):

saegel:
> Please resolve merge conflicts so we could merge it

Issue #166, opened 2022-08-24
Logprep applies some rules multiple times if they match


ppcad:
Logprep applies some rules multiple times if they matches.
It's one time for each matching disjunction a rule filter contains.

**Expected behavior**
Logprep applies a rule only once if it matches.

**Current behavior**
Logprep applies a rule multiple times for each matching disjunction it's rule filter contains.

**Steps to reproduce**

1. Create a rule with a fiter like "term_1 OR term_2", where both terms are true for some event X. This is most easily tested with a processor that does not allow duplication (i.e. the domain resolver).
2. Run Logprpe and send event X.
3. See the duplication warning in the console.

**Environment**

Logprep version:
Python version: >=3.6

**Possible solution**
Adapt the rule tree to not match a rule multiple times or remove duplicates after the matching process if possible.

ekneg54:
in every processor rule we define, how equality is set for the rule. I do not think, that this happens to all processors, but we have to check that.
It could be possible, that we forget about this behavior by defining equality in rules.

ekneg54:
especially  for the domain_resolver we have no rule tests and the equality of two domain_resolver rules is never testet. :-/

ekneg54:
we could dissect a rule with a logical OR expression in filter statement into two rules where the dissected rules have one part of the filter expression as filter and the same other fields. so these rules are not processed multiple times for one event?

ekneg54:
is it as simple as I think, that we should return a `set` here?:

https://github.com/fkie-cad/Logprep/blob/6ddbe4875998ed00e3065e3aaa083befebd3db6a/logprep/framework/rule_tree/rule_tree.py#L178

ppcad:
Thanks, but I have already fixed it. I just need to do some more testing.
There was no check for duplicates in the rule tree and the equality check was not done correctly for some rules types.
I removed some duplication that happened already at the time of rule tree creation (in add_rule) and then ensured that there is no duplication in get_matching_rules as well.

One thing I'm still questioning is how to handle rules with different definitions but equal results. There are some rules that have different rule definitions, but end up having equal effects. Like for the generic adder with "add_from_file". There it is possible that the results from the file are the same as if it was added via a different configuration. I decided to make rules equal if the results of their configuration is equal.

ppcad:
Ok, a duplicate check in add_rule can reduce duplicates, but a check/set in get_matching_rules needs to handle the rest.
[...]

PR #168, opened 2022-08-25
Fix duplication in rule tree and adapt rule equalities

ppcad:
See issue #166


logprep/processor/domain_resolver/rule.py
```diff
@@ -34,8 +34,10 @@ def __init__(self, filter_rule: FilterExpression, domain_resolver_cfg: dict):
             self._output_field = "resolved_ip"
 
     def __eq__(self, other: "DomainResolverRule") -> bool:
-        return (other.filter == self._filter) and (
-            self._source_url_or_domain == other.source_url_or_domain
+        return (
+            (other.filter == self._filter)
+            and (self._source_url_or_domain == other.source_url_or_domain)
+            and (self._output_field == other.output_field)
```
ekneg54, line 40: please use the builtin `all` function for performance reasons. it is implemented in C

logprep/framework/rule_tree/rule_tree.py
```diff
@@ -212,8 +213,9 @@ def get_matching_rules(
             if child.does_match(event):
                 current_node = child
 
-                if current_node.matching_rules:
-                    matches += child.matching_rules
+                for matching_rule in current_node.matching_rules:
+                    if matching_rule not in matches:
+                        matches.append(matching_rule)
```
ekneg54, line 218: not needed if we declare `matches` as `set`  in line 210

ppcad: This makes the rule application order random (since matches are being processed in a random order), which is no big deal in itself, but it can make problems with some tests, since they do sometimes succeed and sometimes they don't.
An example for this would be the selective extractor, since it adds labels into a list, which might then be inserted in a different order.
I imagine similar problems could occur with auto-tests. Fixing it in a unit test is relatively easy, but I'm not sure about auto-tests.
Sets might be slightly faster in this situation however.
Not sure what is more important though and if it's worth the hassle.

ekneg54, line 218: if `matches` would be declared as a `set` this wouldn't be needed.

ppcad: You said that already and I commented it. If there is something I'm missing then please tell me.

ekneg54: then the tests have be fixed. the implementation should not rely on the order in a list. I think we should do it.
In tests we have this nice `load_specific_rule` method, which ensures that only this rule is loaded. this should help.

ekneg54: sorry... didn't see your comment

ppcad: I've changed it to use set again. We have to be more careful when writing tests now. Some tests don't load rules directly, but load them when creating a processor. Maybe they can be rewritten at some point to all use `load_specific_rule` instead. Auto-tests are unaffected, since they always apply one rule at a time. I will need to test this more though before I open the pull request.

ekneg54: thanks.... and I agree, that we should rewrite the tests to use the `load_specific_rule` helper method, because it increases readability of the tests a lot. but we do it, when we see, that something does not work reliable.

ekneg54 (changes requested):
the equality for some rules changes with this PR but we do not have tests for this.
could you add tests for the rules where euality changed, please?

a good template would be the equality check for the `selective_extractor` here:
https://github.com/fkie-cad/Logprep/blob/74d9436bc44610b27e004ccdd07da3429d3111d8/tests/unit/processor/selective_extractor/test_selective_extractor_rule.py#L200

ppcad:
> the equality for some rules changes with this PR but we do not have tests for this. could you add tests for the rules where euality changed, please?
> 
> a good template would be the equality check for the `selective_extractor` here:
> 
> https://github.com/fkie-cad/Logprep/blob/74d9436bc44610b27e004ccdd07da3429d3111d8/tests/unit/processor/selective_extractor/test_selective_extractor_rule.py#L200

I changed it. The normalizer equality had to be changed for grok, since it did not work properly.

ekneg54 (approved):
thx for your work

Issue #160, opened 2022-08-16
Split output and input connectors


ekneg54:
**Is your feature request related to a problem? Please describe.**

As a user of logprep I need to have independent input and output connectors to orchestrate my input and output connectors via configuration in `pipeline.yml`

**Additional context**

As now it is not possible to combine to different connectors via pipeline.yml and there is a need to implement them hard in code. with the above flexibility a user could orchestrate the input and output connectors by his own.

PR #171, opened 2022-08-31
Add option for log arrival time and arrival delta

ppcad:
Add timestamp when events arrive in Logprep and calculate the delta time with the event generation to be able to see the delay between event generation and processing by Logprep. See #172.


ekneg54:
Pay attention, because currently we refactor the connector code in a fundamental way.

ekneg54:
we have implementet this in #160 to avoid you from being blockt with integration in the new interface of connectors.
a deep dive introduction of the new connector interfase will follow

ppcad:
> we have implementet this in #160 to avoid you from being blockt with integration in the new interface of connectors. a deep dive introduction of the new connector interface will follow

Ok, thank you.

PR #183, opened 2022-09-23
Make kafka max poll interval configurable

ppcad:


ppcad:
No longer needed. Configuration solved by #357.

PR #187, opened 2022-10-09
Add Dockerfile, .dockerignore and update README

ekneg54:
This adds a production ready `Dockerfile` to have the ability to adhoc run logprep from a specific commit.


README.md
```diff
@@ -228,7 +231,7 @@ Details about the rule language and how to write rules for the processors can be
 ### Installation
 
 Python should be present on the system, currently supported are the versions 3.6 - 3.9.
```
dtrai2, line 233: Could you also make the following changes in the `gettings_started.rst` file, please?

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN python -m pip install --upgrade pip && python -m pip install wheel
+RUN python setup.py sdist bdist_wheel
+
+
+FROM python:3.9-slim as prod
+RUN useradd -s /bin/sh -m -c "logprep user" logprep
+USER logprep
+ENV PATH=/home/logprep/.local/bin:$PATH
+COPY --from=build /logprep/dist/logprep-0+unknown-py3-none-any.whl /
+RUN pip install logprep-0+unknown-py3-none-any.whl
```
dtrai2, line 14: As this should be a production ready docker image maybe it is worth it to reduce the image size by removing obsolete python files. Some libraries like `numpy` create some c files during the `pip install` that are not really needed anymore afterwards. An initial idea could be:

```Dockerfile

RUN find /usr/local/lib/python3.9 -name '*.c' -delete
RUN find /usr/local/lib/python3.9 -name '*.pxd' -delete
RUN find /usr/local/lib/python3.9 -name '*.pyd' -delete
RUN find /usr/local/lib/python3.9 -name '__pycache__' | xargs rm -r
```

dtrai2 (changes requested):

Dockerfile
ekneg54: this doesn't change image size, instead it increases it. every RUN statement is an additional layer in die image and the layer with the files persist ;-)

dtrai2: Well I am sure you can also make this in one RUN statement ;) But okay, I accept that.

README.md
ekneg54: DONE

dtrai2 (approved):

ppcad (approved):

PR #188, opened 2022-10-10
Fix picklingerror in debug

ppcad:
Fix pickling error in debug, but also remove some lazy logging, since the aggregator aggregates the strings ("... %s ...") before they get replaced with variables if it's being used.


ekneg54 (approved):
thanks ... this solves the problem.

PR #191, opened 2022-10-13
Fix time measurement for get_next input connector

### LINKED ISSUE #189: Error with time measurement in input
The measure_time in abc/input.py causes an error if metrics are enabled, since it tries to use the timeout as an event.

**Expected behavior**
Logprep runs without error if metrics are activated.

**Current behavior**
Logprep has error "AttributeError: 'int' object has no attribute 'get'" if metrics are activated.

**Steps to reproduce**
Activate metrics, run Logprep and put events into the pipeline.

**Environment**

Logprep version: 84c027dc1b3c89e155dfa1ed983265e4aaf62fa1
Python version: 3.8

**Possible solution**
Remove measure_time from "get_next" in "abc/input.py".

dtrai2:
The `TimeMeasurement.measure_time` decorator was initially designed for processing times of the processors, but was recently also used for the input connectors. If the time measurement was enabled and was configured to append the result to the event it resulted in an error as the input connector method `get_next` does not receive an event as an argument, instead it receives a timeout argument. 

This fix introduces a check if the given argument is a dictionary or not.
Alternative solution would be to check if the `caller` is of type `Input`.

closes #189

### DIFF
--- a/logprep/util/time_measurement.py
+++ b/logprep/util/time_measurement.py
@@ -28,7 +28,7 @@ class TimeMeasurement:
             def inner(*args, **kwargs):  # nosemgrep
                 if TimeMeasurement.TIME_MEASUREMENT_ENABLED:
                     caller = args[0]
-                    event = args[1]
+                    first_argument = args[1]
                     begin = time()
                     result = func(*args, **kwargs)
                     end = time()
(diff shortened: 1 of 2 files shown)


ppcad (approved):
Thank you, it works.

PR #194, opened 2022-10-16
fix keyerror for message with setted _op_type field

ekneg54:
this fixes a keyerror if the document has a special _op_type field, which we are using to write to data_streams.
the key for the error info will become `create` instead of `index` if you are using a field `"op_type": "create"` in your document

### DIFF
--- a/logprep/connector/elasticsearch/output.py
+++ b/logprep/connector/elasticsearch/output.py
@@ -225,9 +225,11 @@ class ElasticsearchOutput(Output):
         """
         error_documents = []
         for bulk_error in error.errors:
-            error_info = bulk_error.get("index", {})
-            data = error_info.get("data")
-            reason = f'{error_info["error"]["type"]}: {error_info["error"]["reason"]}'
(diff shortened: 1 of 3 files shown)


dtrai2 (commented):
Does this also apply to the elastic search connector? I have seen there are the exact same error handling lines. If so it'd be great if you could fix it there as well.

ekneg54:
@dtrai2 I have no clue... I do not have an elastic stack to test this. But it could be possible.

dtrai2:
Is the answer maybe hidden in the [elasticsearch-py tests](https://github.com/elastic/elasticsearch-py/blob/421b286f1aed0af2a198d8c18955d6321aa324cf/test_elasticsearch/test_server/test_helpers.py)?

ekneg54:
it could be... perhaps ... possible... but I would prefer to wait for the bug report before changing the behavior of the elesticsearch_output

ekneg54:
on the other hand... what could go wrong? :-D

dtrai2 (approved):

PR #196, opened 2022-10-18
add jupyther notebooks as additional documentation

ekneg54:
I think these notebooks could be helpful for future demos and in addition to our documentation.
I think we should provide them as we create them.


ekneg54:
does not make sense

PR #204, opened 2022-10-25
dev-add-keychecker-processor

DogchampDiego:
The processor gets a list of event fields and compares these fields with the fields of a given event. If fields from the List are missing in the event, the processor creates a "missing_fields" field in the Event and writes all missing fields in this new field.

### DIFF
--- a/logprep/processor/key_checker/rule.py
+++ b/logprep/processor/key_checker/rule.py
@@ -59,14 +59,15 @@ class KeyCheckerRule(Rule):
     class Config:
         """key_checker rule config"""
 
-        key_list: list = field(
+        key_list: set = field(
             validator=[
                 validators.deep_iterable(
                     member_validator=validators.instance_of(str),
-                    iterable_validator=validators.instance_of(list),
+                    iterable_validator=validators.instance_of(set),
                 ),
                 partial(min_len_validator, min_length=1),
-            ]
+            ],
+            converter=set,
         )
 
         output_field: str = field(validator=validators.instance_of(str))
--- a/logprep/processor/key_checker/rule.py
+++ b/logprep/processor/key_checker/rule.py
@@ -34,7 +34,7 @@ class KeyCheckerRule(Rule):
             ]
         )
 
-        error_field: str = field(validator=validators.instance_of(str))
+        output_field: str = field(validator=validators.instance_of(str))
 
     def __eq__(self, other: "KeyCheckerRule") -> bool:
         return all([other.filter == self._filter, other._config == self._config])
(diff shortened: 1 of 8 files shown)


logprep/processor/key_checker/rule.py
```diff
@@ -0,0 +1,96 @@
...
+
+        key_list: list = field(
+            validator=[
+                validators.deep_iterable(
+                    member_validator=validators.instance_of(str),
+                    iterable_validator=validators.instance_of(list),
+                ),
+                partial(min_len_validator, min_length=1),
+            ]
+        )
```
ekneg54, line 70: try this to ensure, no key is inserted twice and to get rid of order dependency in equal checks
```suggestion
        key_list: set= field(
            validator=[
                validators.deep_iterable(
                    member_validator=validators.instance_of(str),
                    iterable_validator=validators.instance_of(list),
                ),
                partial(min_len_validator, min_length=1),
            ],
            converter=set
        )
```

tests/unit/processor/key_checker/test_key_checker_rule.py
```diff
@@ -0,0 +1,167 @@
...
+    def test_create_from_dict_validates_config(self, rule, error, message):
+        if error:
+            with pytest.raises(error, match=message):
+                KeyCheckerRule._create_from_dict(rule)
+        else:
+            keychecker_rule = KeyCheckerRule._create_from_dict(rule)
+            assert hasattr(keychecker_rule, "_config")
+            for key, value in rule.get("key_checker").items():
+                assert hasattr(keychecker_rule._config, key)
+                sad = type(value)
```
ekneg54, line 94:
```suggestion
```
not needed?

tests/unit/processor/key_checker/test_key_checker_rule.py
```diff
@@ -0,0 +1,167 @@
...
+                },
+                {
+                    "filter": "*",
+                    "key_checker": {
+                        "key_list": ["key1"],
+                        "output_field": "missing_fields",
+                    },
+                },
+                False,
+            ),
```
ekneg54, line 161:
```suggestion
            ),
                        (
                "should be equal, because unique keys are the same",
                {
                    "filter": "*",
                    "key_checker": {
                        "key_list": ["key1", "key2", "key2"],
                        "output_field": "missing_fields",
                    },
                },
                {
                    "filter": "*",
                    "key_checker": {
                        "key_list": ["key2", "key1", "key1"],
                        "output_field": "missing_fields",
                    },
                },
                True,
            ),
```
please add this test

ekneg54 (changes requested):
good job so far!
but some minor changes should be added.

ekneg54 (approved):

logprep/processor/key_checker/rule.py
```diff
@@ -0,0 +1,97 @@
+"""
+KeyCheckerRule
+------------
+
+The `key_checker` processor needs a list with atleast one element in it.
+The Rule contains this list and it also contains a missing field where the processor
+can store all missing keys.
```
urdit52, line 7:
```suggestion
The `key_checker` processor needs a list with at least one element in it.
The Rule contains this list and it also contains a custom field where the processor
can store all missing keys.
```

urdit52 (changes requested):
Just a few bits of documentation remarks, thank you.

ekneg54:
merged because tests ran on my machine

PR #215, opened 2022-11-11
Fix error message for processing error in field_manager

ekneg54:
fixes the missing_fields output for error warning


logprep/processor/field_manager/processor.py
```diff
@@ -85,9 +85,7 @@ def _get_field_values(self, event, rule):
         return [get_dotted_field_value(event, source_field) for source_field in rule.source_fields]
 
     def _get_missing_fields_error(self, source_fields, field_values):
-        missing_fields = [
-            source_fields[index] for value, index in enumerate(field_values) if value is None
```
dtrai2, line 89: Ah I also always mix up the values of enumerate, you could have just flipped them (-> `... for index, value in ...`), but the `zip` method works also just fine :+1:

dtrai2 (approved):

PR #218, opened 2022-11-15
Fix prometheus_exporter reset on failed pipelines

dtrai2:
In case no `metric_targets` are configured at all logprep also shouldn't try to access the `prometheus_target`. This revises also the init of the `metric_targets`. This prevents initializing the targets per pipeline and instead only initializes them once for the whole logprep instance.


ekneg54 (approved):
looks good to me

PR #219, opened 2022-11-15
add new representation for rules

ekneg54:
* this helps on debugging because you are able to see the rule config in debugger as repr

### DIFF
--- a/tests/testdata/unit/domain_resolver/rules/generic/domain_resolver.yml
+++ b/tests/testdata/unit/domain_resolver/rules/generic/domain_resolver.yml
@@ -26,10 +26,3 @@ filter: client_2
 domain_resolver:
   source_url_or_domain: client_2
 description: ""
----
-# Equal to rule above
-filter: client_2
-domain_resolver:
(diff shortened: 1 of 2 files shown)


tests/testdata/unit/domain_resolver/rules/generic/domain_resolver.yml
```diff
@@ -26,10 +26,3 @@ filter: client_2
 domain_resolver:
   source_url_or_domain: client_2
 description: ""
----
```
dtrai2, line 29: Why does the `repr` method alone make some tests fail? And how come just removing this is the right way to solve the problem?

ekneg54: because we had a bug so the domain_resolver in this case loads same rules twice?

dtrai2: okay

dtrai2 (approved):

PR #222, opened 2022-11-19
Reimplement flipping aggregation test

ekneg54:
* reimplemented using a mock of time.time
* removed unused code
* increased test coverage

### DIFF
--- a/logprep/util/log_aggregator.py
+++ b/logprep/util/log_aggregator.py
@@ -39,11 +39,6 @@ class Aggregator:
         cls.timer_thread.start()
 
     @classmethod
-    def stop_timer(cls):
-        """Stop repeating timer for aggregation."""
-        cls.timer_thread.cancel()
-
-    @classmethod
     def _aggregate(cls, record: LogRecord) -> bool:
         log_id = "{0[levelname]}:{0[name]}:{0[msg]}".format(record.__dict__)
         if log_id not in cls.logs:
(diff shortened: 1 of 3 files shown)


ppcad:
This tests if the aggregator keeps aggregating in consecutive aggregations even if the aggregation threshold was not exceeded.

Example:
Aggragate after 3 logs within 0.25 seconds. The following should happen:

- First 0.25 sec period if #logs>3: Print the first 3 logs as seperate logs and then start aggregating the rest for this period.
  - A. Following 0.25 sec period if #logs>3: Don't print the first 3 logs despite the threshold being not yet exceeded, since the aggregations are consecutive. Aggregate everything for those periods.
  - B. Following 0.25 sec period if #logs<= 3: Print the first 3 logs normally. The next aggregation won't be consecutive, thus 3 logs will be printed again.

This means, the first 3 logs should not be printed for every consecutive aggregation period in which the threshold has been exceeded.

We should get something like this:
"Some log message"
"Some log message"
"Some log message"
"Some log message (6 in ~0.25 sek)"
"Some log message (4 in ~0.25 sek)"
"Some log message (5 in ~0.25 sek)"
(>0.25 sec passed)
"Some log message"
(>0.25 sec passed)
"Some log message"
"Some log message"
"Some log message"
"Some log message (5 in ~0.25 sek)"
"Some log message (7 in ~0.25 sek)"

This has to be testet somewhere.

ekneg54:
ahh... ok... right... Thank you for the explanation...  I will do my best, to implement it in a testable way or rewrite a stable non flipping test for it.

ppcad:
> ahh... ok... right... Thank you for the explanation... I will do my best, to implement it in a testable way or rewrite a stable non flipping test for it.

Thank you, that would be great!

logprep/util/log_aggregator.py
```diff
@@ -99,14 +94,3 @@ def _perform_logging_if_possible(cls):
             else:
                 if time() - cls.logs[log_id]["first_record"].created >= cls.log_period:
                     cls.logs[log_id]["aggregate"] = False
-
-    @staticmethod
```
ppcad, line 103: Logprep will crash without "filter". It is required to perform the aggregation with the logger.
The remaining methods that you removed did not seem to be called anywhere.

ppcad (changes requested):

ppcad (approved):
Thanks for the fix!

PR #226, opened 2022-11-24
Add requester processor

ekneg54:
A processor to invoke http requests. Could be usefull to enrich events from an external api or
to trigger external systems by and with event field values.
Parses the response to enrich event data.

### DIFF
--- a/logprep/processor/requester/processor.py
+++ b/logprep/processor/requester/processor.py
@@ -22,5 +22,8 @@ class Requester(Processor):
     rule_class = RequesterRule
 
     def _apply_rules(self, event, rule):
-        response = requests.request(url=rule.url, method=rule.method, **rule.kwargs)
-        response.raise_for_status()
+        try:
+            response = requests.request(url=rule.url, method=rule.method, **rule.kwargs)
+            response.raise_for_status()
+        except requests.HTTPError as error:
+            self._handle_warning_error(event, rule, error)
--- a/logprep/processor/requester/processor.py
+++ b/logprep/processor/requester/processor.py
@@ -1,6 +1,10 @@
 """
 Requester
-============
+=========
+
+A processor to invoke http requests. Could be usefull to enrich events from an external api or
+to trigger external systems by and with event field values.
+
 """
 from typing import List, Tuple, Any
 
--- a/logprep/processor/requester/processor.py
+++ b/logprep/processor/requester/processor.py
@@ -38,10 +38,7 @@ class Requester(Processor):
         except requests.exceptions.HTTPError as error:
             self._handle_warning_error(event, rule, error)
         if rule.target_field:
-            try:
-                result = json.loads(rsp.content)
-            except json.JSONDecodeError:
-                result = rsp.content.decode("utf-8")
+            result = self._get_result(rsp)
             successful = add_field_to(
                 event,
                 rule.target_field,
@@ -52,6 +49,28 @@ class Requester(Processor):
             if not successful:
                 error = DuplicationError(self.name, [rule.target_field])
                 self._handle_warning_error(event, rule, error)
+        if rule.target_field_mapping:
+            result = self._get_result(rsp)
+            for source_field, target_field in rule.target_field_mapping.items():
+                source_field_value = get_dotted_field_value(result, source_field)
+                successful = add_field_to(
+                    event,
+                    target_field,
+                    source_field_value,
+                    rule.extend_target_list,
+                    rule.overwrite_target,
+                )
+                if not successful:
+                    error = DuplicationError(self.name, [rule.target_field])
+                    self._handle_warning_error(event, rule, error)
+
+    @staticmethod
+    def _get_result(rsp):
+        try:
+            result = json.loads(rsp.content)
+        except json.JSONDecodeError:
+            result = rsp.content.decode("utf-8")
+        return result
 
     def _template_kwargs(self, kwargs: dict, source: dict):
         for key, value in kwargs.items():
--- a/logprep/processor/requester/processor.py
+++ b/logprep/processor/requester/processor.py
@@ -2,7 +2,7 @@
 Requester
 =========
 
-A processor to invoke http requests. Could be usefull to enrich events from an external api or
+A processor to invoke http requests. Can be used to enrich events from an external api or
 to trigger external systems by and with event field values.
 
 Example
--- a/logprep/processor/requester/processor.py
+++ b/logprep/processor/requester/processor.py
@@ -6,6 +6,7 @@ A processor to invoke http requests. Could be usefull to enrich events from an e
 to trigger external systems by and with event field values.
 
 """
+from functools import partial
 import re
 import requests
 from typing import List, Tuple, Any
@@ -24,8 +25,11 @@ class Requester(Processor):
     rule_class = RequesterRule
 
     def _apply_rules(self, event, rule):
-        url = rule.url
-
+        source_fields = rule.source_fields
+        source_field_values = map(partial(get_dotted_field_value, event), source_fields)
+        source_field_dict = dict(zip(source_fields, source_field_values))
+        self._check_for_missing_values(event, rule, source_field_dict)
+        url = self._template(rule.url, source_field_dict)
         try:
             response = requests.request(url=url, method=rule.method, **rule.kwargs)
             response.raise_for_status()
@@ -39,3 +43,11 @@ class Requester(Processor):
             pattern = r"\$\{(" + rf"{key}" + r")\}"
             string = re.sub(pattern, str(value), string)
         return string
+
+    def _check_for_missing_values(self, event, rule, source_field_dict):
+        missing_fields = list(
+            dict(filter(lambda x: x[1] in [None, ""], source_field_dict.items())).keys()
+        )
+        if missing_fields:
+            error = BaseException(f"{self.name}: no value for fields: {missing_fields}")
+            self._handle_warning_error(event, rule, error)
--- a/logprep/processor/requester/processor.py
+++ b/logprep/processor/requester/processor.py
@@ -5,6 +5,18 @@ Requester
 A processor to invoke http requests. Could be usefull to enrich events from an external api or
 to trigger external systems by and with event field values.
 
+Example
+^^^^^^^
+..  code-block:: yaml
+    :linenos:
+
+    - requestername:
+        type: requester
+        specific_rules:
+            - tests/testdata/rules/specific/
+        generic_rules:
+            - tests/testdata/rules/generic/
+
 """
 import json
 import re
--- a/logprep/processor/requester/processor.py
+++ b/logprep/processor/requester/processor.py
@@ -5,6 +5,8 @@ Requester
 A processor to invoke http requests. Can be used to enrich events from an external api or
 to trigger external systems by and with event field values.
 
+For further information for the rule language see: :ref:`requester_rule`
+
 Example
 ^^^^^^^
 ..  code-block:: yaml
--- a/logprep/processor/requester/processor.py
+++ b/logprep/processor/requester/processor.py
@@ -32,11 +32,10 @@ class Requester(Processor):
         source_field_dict = dict(zip(source_fields, source_field_values))
         self._check_for_missing_values(event, rule, source_field_dict)
         kwargs = self._template_kwargs(rule.kwargs, source_field_dict)
-        try:
-            rsp = requests.request(**kwargs)
-            rsp.raise_for_status()
-        except requests.exceptions.HTTPError as error:
-            self._handle_warning_error(event, rule, error)
+        rsp = self._request(event, rule, kwargs)
+        self._handle_response(event, rule, rsp)
+
+    def _handle_response(self, event, rule, rsp):
         if rule.target_field:
             result = self._get_result(rsp)
             successful = add_field_to(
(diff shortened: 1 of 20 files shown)


logprep/processor/requester/processor.py
```diff
@@ -0,0 +1,112 @@
...
+            string = re.sub(pattern, str(value), string)
+        return string
+
+    def _check_for_missing_values(self, event, rule, source_field_dict):
+        missing_fields = list(
+            dict(filter(lambda x: x[1] in [None, ""], source_field_dict.items())).keys()
+        )
+        if missing_fields:
+            error = BaseException(f"{self.name}: no value for fields: {missing_fields}")
+            self._handle_warning_error(event, rule, error)
```
dtrai2, line 112: If you wait for the `timestamp_differ` pull request you can make use of this method by calling it from the parent class.

logprep/processor/requester/processor.py
```diff
@@ -0,0 +1,112 @@
...
+    def _apply_rules(self, event, rule):
+        source_field_dict = get_source_fields_dict(event, rule)
+        self._check_for_missing_values(event, rule, source_field_dict)
+        kwargs = self._template_kwargs(rule.kwargs, source_field_dict)
+        rsp = self._request(event, rule, kwargs)
+        self._handle_response(event, rule, rsp)
+
+    def _handle_response(self, event, rule, rsp):
+        if rule.target_field:
+            result = self._get_result(rsp)
```
dtrai2, line 48: both if cases make use of this line, maybe we can do it once in the first line of this method?

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,186 @@
...
+Requester
+=========
+
+The Requester is configured by the keyword :code:`requester`.
+It can be used to trigger external systems via web request or enrich eventdata by external
+apis.
+
+A speaking example for event enrichment via external api:
+
+..  code-block:: yaml
```
dtrai2, line 11: Can you add to this example how the raw response from the api call would look like? I think that could enrich the understanding of this example.

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,186 @@
...
+        url: str = field(
+            validator=[
+                validators.instance_of(str),
+                validators.matches_re(rf"^({URL_REGEX_PATTERN})|({FIELD_PATTERN}.*)"),
+            ]
+        )
+        """The url for the request. You can use dissect pattern language to add field values"""
+        json: dict = field(validator=validators.instance_of(dict), factory=dict)
+        """ (Optional) The json payload. Can be templated by using the pattern
+        :code:`${the.dotted.field}` somewhere in the key or value all elements.
```
dtrai2, line 111: I would write something like (relevant also for the other docs below): 
```suggestion
        """ (Optional) The json payload. Can be enriched with event data by using the pattern
        :code:`${the.dotted.field}` to retrieve nested field values. 
```

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,186 @@
...
+
+    {"message": {"hostname": "BB37293hhj"}}
+
+..  code-block:: json
+    :linenos:
+    :caption: Processed event
+
+    {"message": {"hostname": "BB37293hhj"},
+     "cmdb": {
+         "locaction": {
```
dtrai2, line 39:
```suggestion
         "location": {
```

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,186 @@
...
+
+parameter_keys = inspect.signature(requests.Request).parameters.keys()
+REQUEST_CONFIG_KEYS = [
+    parameter for parameter in parameter_keys if parameter not in ["hooks", "cookies", "files"]
+] + ["timeout", "proxies", "verify", "cert"]
+
+URL_REGEX_PATTERN = r"(http|https):\/\/.+"
+
+
+HTTP_METHODS = ["GET", "OPTIONS", "HEAD", "POST", "PUT", "PATCH", "DELETE"]
```
dtrai2, line 66:
```suggestion
URL_REGEX_PATTERN = r"(http|https):\/\/.+"
HTTP_METHODS = ["GET", "OPTIONS", "HEAD", "POST", "PUT", "PATCH", "DELETE"]
```

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class RequesterRule(FieldManagerRule):
+    """Interface for a simple Rule with source_fields and target_field"""
```
dtrai2, line 70: Copy paste error, please fix it.

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,186 @@
...
+        auth: tuple = field(
+            validator=[validators.instance_of(tuple)],
+            converter=tuple,
+            factory=tuple,
+        )
+        """ (Optional) The authentication tuple. Defined as list. Will be converted to tuple"""
+        timeout: float = field(validator=validators.instance_of(float), converter=float, default=2)
+        """ (Optional) The timeout in seconds as float for the request. Defaults to 2 seconds"""
+        verify: bool = field(validator=validators.instance_of(bool), default=True)
+        """ (Optional) Wether or not verify the ssl context. Defaults to :code:`True`"""
```
dtrai2, line 148:
```suggestion
        """ (Optional) Whether or not verify the ssl context. Defaults to :code:`True`."""
```

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,186 @@
...
+    }
+"""
+import inspect
+import json
+import re
+
+import requests
+from attrs import define, field, validators
+
+from logprep.processor.calculator.rule import FIELD_PATTERN
```
dtrai2, line 55: I'm not sure if I like this calculator dependency. Maybe we can move the `FIELD_PATTERN` up into the parent class? As I do use a similar pattern also in the `timestamp_differ`. Maybe we can align them and make it more general.

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,186 @@
...
+        to be used on the request"""
+        cert: str = field(validator=validators.instance_of(str), default="")
+        """(Optional) SSL client certificate as path to ssl client cert file (.pem)."""
+
+        def __attrs_post_init__(self):
+            url_fields = re.findall(FIELD_PATTERN, self.url)
+            json_fields = re.findall(FIELD_PATTERN, json.dumps(self.json))
+            data_fields = re.findall(FIELD_PATTERN, self.data)
+            params_fields = re.findall(FIELD_PATTERN, json.dumps(self.params))
+            self.source_fields = list({*url_fields, *json_fields, *data_fields, *params_fields})
```
dtrai2, line 170: I think that's the same issue here you remarked in the `timestamp_differ`. Doesn't this break the `delete_source_field` option (in case the `delete_source_field` option makes sense here)? So same remark here could you investigate with a test what happens if someone adds this option to their rule. You can also wait again for the `timestamp_differ` as I have revised the `_apply_rules_wrapper` there.

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # pylint: disable=missing-docstring
+    @property
+    def kwargs(self):
+        kwargs = {
+            key: getattr(self._config, key)
+            for key in REQUEST_CONFIG_KEYS
+            if getattr(self._config, key)
```
dtrai2, line 178: Probably it won't make a difference, but couldn't you use here `hasattr`?

requirements_dev.txt
```diff
@@ -235,7 +238,7 @@ ruamel-yaml-clib==0.2.7
     # via
     #   -r ./requirements.txt
     #   ruamel-yaml
-semgrep==0.122.0
+semgrep==1.0.0
```
dtrai2, line 241: As it appears semgrep recently had it's `1.0` release. Maybe you can check update the `attrs` requirements now as we have fixed it. If we can update it we can finally get rid of our custom `min_len_validator`.

tests/unit/processor/requester/test_requester.py
```diff
@@ -0,0 +1,274 @@
...
+        "type": "requester",
+        "specific_rules": ["tests/testdata/unit/requester/specific_rules"],
+        "generic_rules": ["tests/testdata/unit/requester/generic_rules"],
+    }
+
+    @responses.activate
+    @pytest.mark.parametrize("testcase, rule, event, expected, response_kwargs", test_cases)
+    def test_testcases(
+        self, testcase, rule, event, expected, response_kwargs
+    ):  # pylint: disable=unused-argument
```
dtrai2, line 247: If you use the `testcase` as a argument for the `assert` statement you can get rid of this pylint disable.  See line 263.

tests/unit/processor/requester/test_requester.py
```diff
@@ -0,0 +1,274 @@
...
+    ):
+        if response_kwargs:
+            responses.add(responses.Response(**response_kwargs))
+        self._load_specific_rule(rule)
+        with pytest.raises(ProcessingWarning):
+            self.object.process(event)
+        assert event == expected, testcase
+
+    @responses.activate
+    def test_process(self):
```
dtrai2, line 266: What is this test case for? There is only one assert that should be always true and after the other calls there is nothing checked again.

tests/unit/processor/requester/test_requester_rule.py
```diff
@@ -0,0 +1,257 @@
...
+                    "filter": "message",
+                    "requester": {
+                        "method": "GET",
+                        "url": "http://the-api/endpoint",
+                        "params": '{"key": "value"}',
+                    },
+                },
+                TypeError,
+                r"must be <class \\\'dict\\\'>",
+            ),
```
dtrai2, line 129: I think that's a duplicate from line 94 to 105.

dtrai2 (changes requested):

tests/unit/processor/requester/test_requester.py
ekneg54: I had to override this from the base class, to decorate it with the responses fixture

logprep/processor/requester/rule.py
ekneg54: done

logprep/processor/requester/rule.py
ekneg54: moved FIELD_PATTERN to field_manager and refactored all three processors to this

logprep/processor/requester/rule.py
ekneg54: done

logprep/processor/requester/rule.py
ekneg54: it is not the same issue, because it leads to a flat list with strings of source fields gotten from the parameters.
but you are right and I added a test for delete_source_field and overwrite target field

logprep/processor/requester/rule.py
ekneg54: no, because I only want to add if I get a truthy value back. I do not want to add empty fields

requirements_dev.txt
ekneg54: not in this PR, but soon we will try it

tests/unit/processor/requester/test_requester.py
ekneg54: done

tests/unit/processor/requester/test_requester_rule.py
ekneg54: done

logprep/processor/requester/processor.py
```diff
@@ -0,0 +1,106 @@
+"""
+Requester
+=========
+
+A processor to invoke http requests. Can be used to enrich events from an external api or
+to trigger external systems by and with event field values.
+
+For further information for the rule language see: :ref:`requester_rule`
```
dtrai2, line 8:
```suggestion
For further information for the rule language see: :ref:`requester_rule`.
```

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,195 @@
...
+    description: '...'
+
+..  code-block:: json
+    :linenos:
+    :caption: Incoming event
+
+    {"message": {"hostname": "BB37293hhj"}}
+
+.. code-block:: json
+    :lineos:
```
dtrai2, line 36:
```suggestion
..  code-block:: json
    :linenos:
```

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,195 @@
...
+        url: str = field(
+            validator=[
+                validators.instance_of(str),
+                validators.matches_re(rf"^({URL_REGEX_PATTERN})|({FIELD_PATTERN}.*)"),
+            ]
+        )
+        """The url for the request. You can use dissect pattern language to add field values"""
+        json: dict = field(validator=validators.instance_of(dict), factory=dict)
+        """ (Optional) The json payload. Can be enriched with event data by using the pattern
+        :code:`${the.dotted.field}` to retrieve nested field values.
```
dtrai2, line 120:
```suggestion
        :code:`${the.dotted.field}` to retrieve nested field values.
```

logprep/processor/requester/rule.py
```diff
@@ -0,0 +1,195 @@
...
+
+..  code-block:: json
+    :linenos:
+    :caption: Incoming event
+
+    {"message": {"hostname": "BB37293hhj"}}
+
+.. code-block:: json
+    :lineos:
+    :caption: raw response json data given from the api
```
dtrai2, line 37:
```suggestion
    :caption: Raw response json data given from the api
```

dtrai2 (changes requested):
Thanks for the changes, here some last minor docu change requests.

logprep/processor/requester/rule.py
dtrai2: You only applied this one to the json data, could you also changes this for the `data` and `params` options please?

dtrai2 (changes requested):

ekneg54:
done

dtrai2 (approved):

PR #237, opened 2022-12-01
add notebooks

ekneg54:
* we have a need to share these notebooks with the community

### DIFF
--- a/notebooks/concatenator.ipynb
+++ b/notebooks/concatenator.ipynb
@@ -1,0 +1,218 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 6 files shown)


dtrai2:
I just tried out some notebooks and not all of them are working. The `dissector` for example still has the typo and is called `dissecter` and some miss also the line `sys.path.append("..")` befor the `Factory` import.

ekneg54:
I like it.
Removed the performance analysis, because it was broken.
need an approve.

dtrai2 (approved):

PR #245, opened 2022-12-06
log python and logprep version on startup 

ekneg54:


logprep/run_logprep.py
```diff
@@ -141,6 +141,8 @@ def main():
     try:
         AggregatingLogger.setup(config, logger_disabled=args.disable_logging)
         logger = AggregatingLogger.create("Logprep")
+        logger.info(f"python version: {sys.version}")
+        logger.info(f"logprep version: {get_versions()}")
```
dtrai2, line 145: If you are at it already, how about we log also the found configuration version?

logprep/run_logprep.py
```diff
@@ -195,4 +197,5 @@ def main():
 
 
 if __name__ == "__main__":
+
```
dtrai2, line 200: and this one I think you can remove again.

dtrai2 (changes requested):

logprep/run_logprep.py
ekneg54: done

logprep/run_logprep.py
ekneg54: done

dtrai2:
I just replaced the `\t` with a fixed padding. This bothered me for a while already as the `\t` was rendered differently on different system. Hope this fixed padding fixes that.

dtrai2 (approved):

PR #251, opened 2022-12-09
refactor pipeline

ekneg54:


logprep/framework/pipeline.py
```diff
@@ -158,24 +158,23 @@ def _create_metrics(self):
...
+        self._logger.debug(f"Building '{current_process().name}'")
+        self._pipeline = tuple(
+            (self._create_processor(entry) for entry in self._logprep_config.get("pipeline"))
+        )
+        self._logger.debug(f"Finished building pipeline ({current_process().name})")
+
+    def _create_processor(self, entry: dict) -> "Processor":
+        processor_name = list(entry.keys())[0]
+        entry[processor_name]["metric_labels"] = self._metric_labels
+        processor = Factory.create(entry, self._logger)
```
dtrai2, line 170: I know it's only indirectly related with this refactoring, but could you fix the type-hint of the factory? It currently has the return type `Connector` which conflicts with the type-hint of this method. So it should have a type like `Union[Connector, Processor]`.

logprep/framework/pipeline.py
```diff
@@ -251,10 +233,34 @@ def _iterate(self):
     def _enable_iteration(self):
         self._continue_iterating = True
 
-    def _retrieve_and_process_data(self):
-        event = {}
+    def _process_pipeline(self):
```
dtrai2, line 236:
```suggestion
    def _apply_pipeline(self):
```

dtrai2 (changes requested):

logprep/framework/pipeline.py
ekneg54: done

logprep/framework/pipeline.py
ekneg54: noooooo.... I think apply is not the right verb... but for sure with process_pipeline I have problems too... 

but not as much as with apply :-D

dtrai2 (approved):

PR #254, opened 2022-12-12
add build image

ekneg54:
tried it in branch pipeline and it should work:

revise by `docker pull ghcr.io/fkie-cad/logprep:latest`


dtrai2:
I checked the CI pipeline and saw that we have three tags there. So which Logprep Versions are in each docker image? I also did the `docker pull ghcr.io/fkie-cad/logprep:latest` followed by `docker run ghcr.io/fkie-cad/logprep:latest --version`. Sadly I don't get a correct logprep version info, instead I get the following: 

```bash
python version:          3.11.1
logprep version:         0+unknown
configuration version:   no configuration found in '/etc/logprep/pipeline.yml'
```
Could you please check why we don't have a version info there. It would be desirable to see something like: `v4.0.0+14`.

ekneg54:
I fixed it. Tryout with `docker run ghcr.io/fkie-cad/logprep:py3.9-latest --version`
is built together with releasing now
and images are built for all supported python versions now
the release step we have to monitor on next release, that the version is set the right way

dtrai2 (changes requested):
Thanks for fixing the logprep version. Sadly the python version is not correct now. When I run `docker run ghcr.io/fkie-cad/logprep:py3.11-latest --version` then I get the following output:
```bash
python version:          3.9.16
logprep version:         4.0.0+31.g3567d27
configuration version:   no configuration found in '/etc/logprep/pipeline.yml'
```
So for some reason there is still python 3.9 being used, even though I chose the image with py3.11.

ekneg54:
now it works.... test with: `docker run --pull always ghcr.io/fkie-cad/logprep:py3.10-latest --version`

dtrai2 (approved):
nice, thanks for the fix :+1:

PR #258, opened 2022-12-16
fix version output

ekneg54:
* version output could potentially expose secret data


urdit52 (approved):
good solution. Thx for fixing this.

PR #260, opened 2022-12-19
add session handling to http getter

ekneg54:


tests/unit/test_run_logprep.py
```diff
@@ -57,12 +57,12 @@ def test_raises_getter_error_for_not_existing_protocol(self):
...
 
-    @mock.patch("requests.get")
-    def test_gets_config_from_https(self, mock_request):
+    @responses.activate
+    def test_gets_config_from_https(self):
         pipeline_config = Path("quickstart/exampledata/config/pipeline.yml").read_text(
             encoding="utf8"
         )
-        mock_request.return_value.text = pipeline_config
+        responses.add(responses.GET, "https://does.not.exits/pipline.yml", pipeline_config)
```
urdit52, line 65: Even though this is just a test, can we change the following, please?

```suggestion
        responses.add(responses.GET, "https://does.not.exist/pipeline.yml", pipeline_config)
```

urdit52: This also needs to be changed below in order to be consistent in this test, of course. (l. 70)

urdit52 (changes requested):
Minor suggestion in one test file.

urdit52 (approved):
thy for changing this.

PR #261, opened 2022-12-19
add get_raw to getter and make it the abstract method

ekneg54:


urdit52 (approved):
based on feedback for #260, this one looks good to me. Thx for committing.

PR #263, opened 2022-12-21
fix artifacts load during process time

ekneg54:
This PR fixes a bug, that the domain_label_extractor and domain_resolver download the tldlists during processing time. 
With this fix the tld lists were downloaded during setup time.
Additionally the lists were saved per process and per processor instance to avoid problems with getting a file lock on the downloaded artifacts.


urdit52 (approved):
Good solution. Thx for committing this.

PR #268, opened 2022-12-21
resolve deprecation warnings in test for labeler

ekneg54:


urdit52 (approved):
these were a lot, thank you for updating them.

PR #270, opened 2022-12-21
Fix deprecation warnings pseudonymizer

ekneg54:
This fixes a bug in pseudonymizer for new rule structure. Additionally I had to resolve some circular import issues.


dtrai2 (approved):

dtrai2:
> This fixes a bug in predetector for new rule structure. Additionally I had to resolve some circular import issues.

Just for documentation reasons, I think you mean pseudonymizer and not predetector, right?

PR #272, opened 2022-12-22
fix minor issues

ekneg54:


urdit52 (approved):
thanks a lot for fixing the documentation, as this makes things work on Ubuntu.

PR #275, opened 2023-01-03
Add config refresh interval

ekneg54:


dtrai2 (changes requested):
Feedback so far, more once it is no draft anymore.

logprep/runner.py
```diff
@@ -245,8 +260,13 @@ def stop(self):
...
     def _keep_iterating(self):
-        with self._continue_iterating.get_lock():
-            return self._continue_iterating.value
+        """generator function"""
+        while True:
+            with self._continue_iterating.get_lock():
+                iterate = self._continue_iterating.value
+                if not iterate:
+                    return
+                yield iterate
```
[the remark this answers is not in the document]
ekneg54: with this change we could use a for loop instead of a while loop in the method `start()`. This leads to easier tests, because we have the control over how often the loop will happen by patching `runner._keep_iterating` with our own generator function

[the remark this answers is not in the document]
ekneg54: and we get rid of the `StoppinIteration` error because the return in a generator function raises the `StopIteration` error. It is a more pythonic approach

logprep/runner.py
```diff
@@ -230,6 +249,15 @@ def reload_configuration(self):
                 + str(error)
             )
 
+    def _schedule_config_refresh_job(self):
+        refresh_interval = self._configuration.get("config_refresh_interval")
```
dtrai2, line 253: In an error cases you enforce a min interval of 5 seconds. Maybe we should do the same on the initial start up of logprep. Just in case someone should be so creative to use a `config_refresh_interval` of 0 seconds. At the moment this would mean that logprep might try to reload the configuration many times per second.

logprep/runner.py
```diff
@@ -230,6 +249,15 @@ def reload_configuration(self):
                 + str(error)
```
dtrai2, line 249: I am not sure if you have to add the original error message here, as the verification of the configuration already logs the errors as a batch. The only down side that might could use some refactoring is that the error messages from the verification are not logged but just printed in red.

logprep/runner.py
```diff
@@ -210,18 +210,37 @@ def reload_configuration(self):
         """
         if self._configuration is None:
             raise CannotReloadWhenConfigIsUnsetError
-
-        new_configuration = Configuration.create_from_yaml(self._yaml_path)
-
+        try:
+            new_configuration = Configuration.create_from_yaml(self._yaml_path)
+        except (requests.RequestException) as error:
```
dtrai2, line 215: What about `No such file or directory` error? If I run the quickstart config, not with a SimpleHTTPServer but really just the plain old config file I get the error `No such file or directory` if the original config file should get lost before the next reload attempt. Logprep then goes into a state of shutting down, but not doing it completely.

dtrai2 (changes requested):

logprep/runner.py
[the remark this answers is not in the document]
dtrai2: Ah okay, the for loop I have noticed but the advantage of the tests I didn't see yet. And also a neat behavior of a generator. Thanks for the explanation :+1:

logprep/runner.py
ekneg54: I would leave it as is, because I didn't wrote it and didn't changed it

logprep/runner.py
ekneg54: good point -> DONE

tests/unit/test_run_logprep.py
```diff
@@ -175,13 +204,65 @@ def test_no_config_error_is_printed_if_no_config_was_arg_was_given(self, capsys)
...
+    def test_logprep_exits_if_logger_can_not_be_created(self):
+        with mock.patch("logprep.run_logprep.AggregatingLogger.create") as mock_create:
+            mock_create.side_effect = BaseException
+            config_path = "quickstart/exampledata/config/pipeline.yml"
+            with mock.patch("sys.argv", ["logprep", config_path]):
+                with pytest.raises(SystemExit):
+                    run_logprep.main()
+
+    @mock.patch("logprep.util.log_aggregator.Aggregator")
+    def test_logprep_exits_on_invalid_configuration(self, _):
```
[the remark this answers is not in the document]
ekneg54: copy paste :-D -> DONE

tests/unit/test_run_logprep.py
```diff
@@ -175,13 +204,65 @@ def test_no_config_error_is_printed_if_no_config_was_arg_was_given(self, capsys)
...
+        with mock.patch("logprep.util.configuration.Configuration.verify") as mock_verify:
+            mock_verify.side_effect = Exception
+            config_path = "quickstart/exampledata/config/pipeline.yml"
+            with mock.patch("sys.argv", ["logprep", config_path]):
+                with pytest.raises(SystemExit):
+                    run_logprep.main()
+
+    @mock.patch("logprep.util.log_aggregator.Aggregator")
+    def test_logprep_exits_on_request_exception(self, _):
+        with mock.patch("logprep.util.configuration.Configuration.verify") as mock_verify:
```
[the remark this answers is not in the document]
ekneg54: good point -> DONE

logprep/runner.py
ekneg54: as you command... -> DONE :-D

dtrai2 (approved):

PR #282, opened 2023-01-10
Dev ducplication warnings and generic adder overwrite

ppcad:
Replace duplicarion errors with duplication warnings.
Allow overwrite and extend via FieldManager for the GenericAdder.
Make tests using "/usr/bin/ls" use "/bin/ls" if the former is not available.
Slightly increase timeout for config_refresh acceptance test.


ekneg54:
we have the `_apply_rules_wrapper` in the class `logprep.abc.processor.Processor`. Is it possible to handle the `ProcessingWarnings` there, isn`t it?

Somthing like this:

```python
    def _apply_rules_wrapper(self, event, rule):
        try:
            self._apply_rules(event, rule)
        except ProcessingWarning as error:
            self._handle_warning_error(event, rule, error)
        if not hasattr(rule, "delete_source_fields"):
            return
        if rule.delete_source_fields:
            for dotted_field in rule.source_fields:
                pop_dotted_field_value(event, dotted_field)
```

If so, I guess we could undo half the code changes here.

ppcad:
> we have the `_apply_rules_wrapper` in the class `logprep.abc.processor.Processor`. Is it possible to handle the `ProcessingWarnings` there, isn`t it?
> 
> Somthing like this:
> 
> ```python
>     def _apply_rules_wrapper(self, event, rule):
>         try:
>             self._apply_rules(event, rule)
>         except ProcessingWarning as error:
>             self._handle_warning_error(event, rule, error)
>         if not hasattr(rule, "delete_source_fields"):
>             return
>         if rule.delete_source_fields:
>             for dotted_field in rule.source_fields:
>                 pop_dotted_field_value(event, dotted_field)
> ```
> 
> If so, I guess we could undo half the code changes here.

Most processors did not raise `ProcessingWarning` but `DuplicationError`, so additionaly `ProcessingWarning` would have to be raised instead of calling `_handle_warning_error` within the processors.
Furthermore, `_handle_warning_error` raises another `ProcessingWarning` from the input error. Thus raising a `ProcesisngWarning` form a `ProcessingWarning`.

Did you mean `except DuplicationError`? There exist different versions of `DuplicationError` for some processors. This would have to be unified first.

ekneg54:
I think `DuplicationError` should be a `ProcessingError` because it does not Stop logprep. It should inherit from ProcessingError or could be deleted in favor of using `ProcessingError`

ekneg54:
but we could do the refactoring in a separate Task in the near future.

ppcad:
I think inheriting would be the better option, since the error message would include `ProcessingWarning` twice if it was a `ProcessingWarning` (Something like `... ProcessingWarning: ... ProcessingWarning: ...`).
I will see later if I can do it easily. Otherwise it could be refactored later like you said.

ppcad:
I have refactored it.

ekneg54 (approved):
thank you for your awesome work

PR #283, opened 2023-01-10
Dev substitute from environment in config and rules

ekneg54:
* this adds the functionality to use environment variables in rules and config (pipeline.yml)

### DIFF
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -22,7 +22,11 @@ class Getter(ABC):
     def get(self) -> str:
         """calls the get_raw method and returns the decoded content"""
         content = self.get_raw().decode("utf8")
-        return Template(content).substitute(**os.environ)
+        try:
+            content = Template(content).substitute(**os.environ)
+        except (KeyError, ValueError):
+            pass
+        return content
 
     def get_yaml(self) -> Union[Dict, List]:
         """gets and parses the raw content to yaml"""
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -1,5 +1,7 @@
 """Module for getter interface"""
 from abc import ABC, abstractmethod
+import os
+from string import Template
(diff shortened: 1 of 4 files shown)


quickstart/exampledata/config/pipeline.yml
```diff
@@ -57,7 +58,7 @@ input:
   kafka:
     type: confluentkafka_input
     bootstrapservers:
-      - 172.21.0.5:9092
+      - 127.0.0.1:9092
```
dtrai2, line 61: Can you double check these changes please. I think the IP `172.21.0.5` is necessary for the quickstart setup.

logprep/abc/getter.py
```diff
@@ -19,7 +21,12 @@ class Getter(ABC):
 
     def get(self) -> str:
         """calls the get_raw method and returns the decoded content"""
-        return self.get_raw().decode("utf8")
+        content = self.get_raw().decode("utf8")
+        try:
+            content = Template(content).substitute(**os.environ)
+        except (KeyError, ValueError):
+            pass
+        return content
```
dtrai2, line 29: It's not really a necessary change, but I think this is a nice case where we could make use of the `finally` clause:
```suggestion
        try:
            content = Template(content).substitute(**os.environ)
        finally:
            return content
```

dtrai2 (changes requested):

quickstart/exampledata/config/pipeline.yml
ekneg54: reverted changes on pipeline.yml

logprep/abc/getter.py
ekneg54: this could swallow exceptions, but I only want to handle explicitly these exceptions. All other exceptions should raise without returning content

dtrai2 (approved):

PR #290, opened 2023-01-17
add append without separator to dissector

### LINKED ISSUE #289: Dissector append value without separator
**Is your feature request related to a problem? Please describe.**
I have a log source which timestamp has a #-separated timezone, e.g. `2022 12 06 15:12:30:534#+0100`. I would like to use the dissector to split these fields and append them together without a delimiter. So the end result should look like this: `2022 12 06 15:12:30:534+0100`.

**Describe the solution you'd like**
It should be possible to specify in the dissect append pattern if I want to use the default separator, some other separator or no separator at all. 

**Describe alternatives you've considered**
/

**Additional context**
Following is a test case example with the desired behavior, that could be used in the `test_dissector.py` file. With the current implementation this test fails, because it uses the following separator as delimiter, such that the field `date` will look like: `2022 12 06 15:12:30:534#MOREINFO+0100`.

```python3
test_case = [ 
    (
        "Appending without separator",
        {
            "filter": "message",
            "dissector": {
                "mapping": {
                    "message": "INFO#%{date}#%{+date}#MOREINFO%{}"
                }
            },
        },
        {
            "message": "INFO#2022 12 06 15:12:30:534#+0100#MOREINFO",
        },
        {
            "message": "INFO#2022 12 06 15:12:30:534#+0100#MOREINFO",
            "date": "2022 12 06 15:12:30:534+0100"
        },
    )
]
```

ekneg54:
closes #289

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 `customize_target_subfields`
 * Add a `timestamp_differ` processor that can parse two timestamps and calculate their respective time delta.
 * Add `config_refresh_interval` configuration option to refresh the configuration on a given timedelta
+* Add a feature to `dissector` processor to enable appending to strings without a separator
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Breaking
 
 * drop support for python `3.6`, `3.7`, `3.8`
+* change default prefix behavior on appending to strings of `dissector`
 
 ### Features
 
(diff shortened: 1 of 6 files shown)


sashs (changes requested):
I would also add a test for syntactical correctness, e.g. does all parentheses (open, close) exist

ekneg54:
@sashs  I added a stricter validation of the dissect language. please have a look

sashs (approved):

PR #292, opened 2023-01-18
fix quickstart docker-compose setup

ekneg54:
* this fixes the quickstart and deletes redundant Dockerfile


herrfeder (approved):
Yes, I checked the build process for the Logprep Container on my own machine. Actually it's good that you removed the Layers that requires BuildKit as it isn't necessary here in my opinion. Well, newer versions of docker already come with builtin BuildKit support, therefore it doesn't matter at all :)

Additionally I would recommend to clean the apt cache after the last apt call to have minor improvements in terms of image size.

PR #295, opened 2023-01-22
refactor pipeline.py

ekneg54:


logprep/framework/pipeline.py
```diff
@@ -120,100 +200,106 @@ def __init__(
...
+    def _output(self):
         output_connector_config = self._logprep_config.get("output")
         connector_name = list(output_connector_config.keys())[0]
         output_connector_config[connector_name]["metric_labels"] = self._metric_labels
-        self._output = Factory.create(output_connector_config, self._logger)
+        return Factory.create(output_connector_config, self._logger)
 
-    def _create_input_connector(self):
+    @cached_property
+    def _input(self):
```
dtrai2, line 260: Can you type the cached properties please? My IDE is giving me now 23 more warnings than the current main-state. I think many of these warnings can be prevented if the typing would be more strict. Most of the times those warnings are something like `Unresolved attribute reference 'store' for class 'Component' ` for the line `self._output.store(event)`. Yes the output does have a `store` method, and output is a subclass of `Component` but if we strictly specify that the attribute `_output` is of type `Output` then we can prevent these warnings.

tests/unit/framework/test_pipeline.py
```diff
@@ -106,13 +104,6 @@ def test_shut_down_calls_shut_down_on_pipeline_processors(self, _):
         for processor in processors:
             processor.shut_down.assert_called()
 
-    def test_setup_creates_connectors(self, _):
```
dtrai2, line 109: Couldn't you refactor this test such that it also checks the properties `_input` and `_output`?

tests/unit/framework/test_pipeline.py
```diff
@@ -508,19 +496,16 @@ def test_pipeline_metrics_number_of_errors_counts_errors_of_all_processor_metric
         self.pipeline.metrics.pipeline = [mock_metrics_one, mock_metrics_two]
         assert self.pipeline.metrics.number_of_errors == 2
 
-    def test_create_connector_adds_versions_information_to_input_connector_config(
-        self, mock_create
-    ):
-        self.pipeline._create_logger()
-        self.pipeline._create_connectors()
+    def test_setup_adds_versions_information_to_input_connector_config(self, mock_create):
+        self.pipeline._setup()
```
dtrai2, line 500: This test could also be more specific by referencing the `_input` property directly instead of the `_setup` method, as the setup method doesn't really add the version information.

logprep/framework/pipeline.py
```diff
@@ -358,76 +444,17 @@ def _shut_down(self):
...
-                        msg = (
-                            f"Processed events per {self._period / 60.0:.2f} minutes: "
-                            f"{self._val.value}"
-                        )
-                    if self._logger:
-                        self._logger.info(msg)
-                    self._val.value = 0
-                    self._timer.value = time() + self._period
-
-
```
dtrai2, line 430: It's not part of your changes here, but could you please also fix the wrong type in the MultiprocessingPipeline init?

`used_server_ports: list,` should be `used_server_ports: dict,`.

dtrai2 (changes requested):

logprep/framework/pipeline.py
ekneg54: DONE

logprep/framework/pipeline.py
ekneg54: sure -> done

tests/unit/framework/test_pipeline.py
ekneg54: not needed because it would test python itself...

Issue #298, opened 2023-01-25
List as response type of the requester processor


DFromPMC:
It is not possible to get specific key-value pairs from the http response if the response is of the data type list and not dict

ekneg54:
done with merge request #299

PR #299, opened 2023-01-25
add list access to get_dotted_field_value

ekneg54:
* this adds simple list access to `get_dotted_field_value` helper function
* slicing is not supportet yet

### DIFF
--- a/logprep/util/helper.py
+++ b/logprep/util/helper.py
@@ -104,6 +104,13 @@ def add_field_to(event, output_field, content, extends_lists=False, overwrite_ou
     return False
 
 
+def _get_item(items, item):
+    try:
+        return dict.__getitem__(items, item)
+    except TypeError:
+        return list.__getitem__(items, int(item))
+
+
 def get_dotted_field_value(event: dict, dotted_field: str) -> Optional[Union[dict, list, str]]:
     """
     Returns the value of a requested dotted_field by iterating over the event dictionary until the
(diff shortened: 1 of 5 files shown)


ekneg54:
closes #298

logprep/util/helper.py
```diff
@@ -104,6 +104,13 @@ def add_field_to(event, output_field, content, extends_lists=False, overwrite_ou
     return False
 
 
+def _get_item(items, item):
+    try:
+        return dict.__getitem__(items, item)
+    except TypeError:
+        return list.__getitem__(items, int(item))
```
dtrai2, line 111: looks good to me. I am not sure if there is a need for it, but it is fairly simple to extend this to get characters from a string with `str.__getitem__(items, int(item))`. As one single character might be uninteresting though you could also extend it with the slicing functionality. For that you "just" have to check if the `item` string has a `:` colon in it, split the string and pass it into the `slice(start, stop, step)` method.

logprep/util/helper.py
```diff
@@ -124,9 +131,11 @@ def get_dotted_field_value(event: dict, dotted_field: str) -> Optional[Union[dic
...
     fields = [event, *dotted_field.split(".")]
     try:
-        return reduce(dict.__getitem__, fields)
+        return reduce(_get_item, fields)
     except KeyError:
         return None
+    except ValueError:
+        return None
     except TypeError:
         return None
```
dtrai2, line 140: Here you also have to check for the `IndexError` and write in the `test_helper.py` file a corresponding test. Just for the cases where someone tries to access an element, that is outside of the boundaries of the collection.

dtrai2 (changes requested):

dtrai2 (approved):
I just added some documentation and more test cases, as well as the case for IndexErrors. On my side it can be merged now, but please take a look at my changes first before doing so.

PR #301, opened 2023-01-26
Expand configuration verification

ppcad:
This makes the verification of the configuration handle more errors and reformats it's output for easier debugging.


logprep/factory_error.py
```diff
@@ -16,6 +16,13 @@ def __init__(self):
         super().__init__("There must be exactly one definition per pipeline entry.")
 
 
+class NotExactlyOneEntryInConnectorConfigurationError(InvalidConfigurationError):
+    """Raise if there is not exactly one definition per connector."""
+
+    def __init__(self):
+        super().__init__("There must be exactly one definition for input and output connectors.")
+
+
```
dtrai2, line 25: Can you double check again if we actually need this error? I didn't find any location where it is used.

logprep/factory.py
```diff
@@ -19,22 +18,27 @@ class Factory:
 
     @classmethod
     def create(cls, configuration: dict, logger: "Logger") -> Component:
-        """Create connector."""
+        """Create component."""
         if not configuration:
-            raise NotExactlyOneEntryInConfigurationError()
+            raise InvalidConfigurationError("The component definition is empty.")
+        if not isinstance(configuration, dict):
+            raise InvalidConfigSpecificationError()
```
dtrai2, line 25: Can you add a test case for the raise please?

logprep/factory.py
```diff
@@ -19,22 +18,27 @@ class Factory:
...
             raise InvalidConfigurationError(
-                "There must be exactly one definition per pipeline entry."
+                f"Found multiple component definitions ({', '.join(configuration.keys())}),"
+                + " but there must be exactly one."
             )
-        for connector_name, connector_configuration_dict in configuration.items():
-            if not isinstance(connector_configuration_dict, dict):
+        for component_name, component_configuration_dict in configuration.items():
+            if component_configuration_dict is None:
+                raise InvalidConfigurationError("The component definition is empty.")
```
dtrai2, line 33: Same as above, I don't think this case has a test.

logprep/util/configuration.py
```diff
@@ -103,7 +119,11 @@ def create_from_yaml(path: str) -> "Configuration":
         try:
             configuration = config_getter.get_json()
         except ValueError:
-            configuration = config_getter.get_yaml()
+            try:
+                configuration = config_getter.get_yaml()
+            except ScannerError as error:
+                print_fcolor(Fore.RED, f"Error parsing YAML file: {path}\n{error}")
+                sys.exit(1)
```
dtrai2, line 126: Can you also add a test case for this except block please? So far I don't think that it is tested.

logprep/util/configuration.py
```diff
@@ -195,36 +215,73 @@ def _verify_input(self, logger):
...
 
     def _verify_output(self, logger):
         try:
             _ = Factory.create(self["output"], logger)
         except FactoryError as error:
-            raise InvalidConnectorConfigurationError(str(error)) from error
+            raise InvalidOutputConnectorConfigurationError(str(error)) from error
+        except TypeError as error:
+            msg = self._format_type_error(error)
+            raise InvalidOutputConnectorConfigurationError(msg) from error
```
dtrai2, line 232: Here both TypeError cases need a test case as well :)

logprep/util/configuration.py
```diff
@@ -195,36 +215,73 @@ def _verify_input(self, logger):
...
-    def _verify_pipeline(self, logger: Logger):
+    @staticmethod
+    def _format_type_error(error: TypeError) -> str:
+        msg = str(error)
+        if "missing" in str(error):
+            parameters = re.split(r"argument(s)?:", str(error))[-1].strip()
+            if len(re.split("', '| and ", parameters)) > 1:
+                msg = f"Required options are missing: {parameters}."
+            else:
+                msg = f"Required option is missing: {parameters}."
```
dtrai2, line 244: If I see it correctly then the only difference is the letter `s` here. How about we just simplify this like this:
```suggestion
            msg = f"Required option(s) are missing: {parameters}."
```
That way you also don't have to write an additional test just for the plural version ^^

dtrai2 (changes requested):

ppcad:
Thanks For the review. I have incorporated it.

tests/unit/test_factory.py
```diff
@@ -46,6 +46,42 @@ def test_create_fails_if_config_does_not_contain_type():
...
+
+@mark.parametrize("component", ["foo", 1, True, ["foo"], [], "", 0, False])
+def test_create_fails_if_config_sub_component_is_no_dict(component):
+    with raises(
+        InvalidConfigurationError,
+        match=r'The configuration for component "foo" must be specified as an object\.',
+    ):
+        Factory.create({"foo": component}, logger)
+
+
```
dtrai2, line 84: Thanks for implementing these test cases. :+1: I think you could simplify it a bit or merge them together though. All these tests expect an exception and are only calling the factory create method. If you consider the Error and the error-match-string as parameters as well, then I think this could be written in one parameterized test. You can have a look at the file `test_field_manager_rule.py`, there is a similar implementation.

tests/testdata/config/config-yml-invalid.yml
```diff
@@ -0,0 +1,2 @@
+process_count: 5
+invalid_yml
```
dtrai2, line 2: Can you try to get around this two line file? I think as an alternative you could just mock this by writing an arbitrary string to the test temp directory and just use this file path for creating the configuration. I write a comment at the corresponding test as well suggesting an implementation.

tests/unit/util/test_configuration.py
```diff
@@ -41,6 +44,14 @@ def assert_fails_when_replacing_key_with_value(self, key, value, expected_messag
         with pytest.raises(InvalidConfigurationError, match=expected_message):
             config.verify(logger)
 
+    @mock.patch("logprep.util.configuration.print_fcolor")
+    def test_invalid_yml_prints_formatted_error(self, mock_print_fcolor):
+        with pytest.raises(SystemExit, match="1"):
+            Configuration.create_from_yaml(path_to_invalid_yml_config)
```
dtrai2, line 50:
```suggestion
    def test_invalid_yml_prints_formatted_error(self, mock_print_fcolor, tmp_path):
        broken_config_path = Path(tmp_path/"test_config")
        broken_config_path.write_text("process_count: 5\ninvalid_yaml", encoding="utf8")
        with pytest.raises(SystemExit, match="1"):
            Configuration.create_from_yaml(str(broken_config_path))
```

dtrai2 (changes requested):
Thanks for incorporating my suggestions, I have here two small new ones. After that we can merge it. :)

tests/unit/util/test_configuration.py
ppcad: I have changed it.

dtrai2 (approved):
Thank you :+1:

PR #306, opened 2023-01-31
add support for multiple output connectors

### LINKED ISSUE #178: Add support for multiple input and output connectors
**Is your feature request related to a problem? Please describe.**

with #160 the connectors were split into seperate input and output sections. These sections are dictionaries as implemented. 

Now I want to declare the connectors as a list to support multiple inputs and outputs. 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

All outputs should provide a config field to set it as the default output.  
The goal should be to let logprep ship all messages to all default output but enable the `predetector` and the `selective_extractor` to ship to a different output targets with `store_custom`

The inputs should all be processed without any priority.

ekneg54:
* Add support for multiple output connectors
* changes rule language of `selective_extractor`, `pseudonymizer`, `pre_detector` to support multiple outputs

TODO:
- [x] validation of output_names in `selective_extractor`, `pseudonymizer`, `pre_detector` against defined outputs
- [x] add storing extra_data with multiple outputs
- [x] add documentation for callback behavior if two default outputs are defined

ToDo but NotNow:
- add optional overwrite of outputs in `pseudonymizer` and `pre_detector` -> #321 
- add optional default outputs to `selective_extractor` -> #322
- refactor and harmonize code of `pseudonymizer`, `pre_detector` and `selective_extractor` -> #323


ppcad:
This is a very nice feature, but would it be possible to have multiple outputs for extra data as well?
Furthermore, it is not verified if the output name for custom data is valid. Logprep will therefore run and throw a KeyError as soon as custom data is written.

ppcad (changes requested):
This is a very nice feature, but would it be possible to have multiple outputs for extra data as well?
Furthermore, it is not verified if the output name for custom data is valid. Logprep will therefore run and throw a KeyError as soon as custom data is written.

ekneg54:
Sure... I will give it a try.👍😁

ppcad:
Thank you!

ekneg54:
For the selective_extractor storing extra_output to multiple outputs should yet be possible by defining different rules for different outputs or topics.

we could make the same in the pseudonymizer and the pre_detector, that you are able to overwrite the the output and topic defined in the processor configuration on rule level.

this would be the most useful solution in my opinion.

ppcad:
If the goal is to write extra data for the same event into multiple outputs, then having multiple rules with different outputs would have the problem that events need to be processed by each rule variant. This would multiply the rule count by the number of outputs, which would multiply the processing time of that processor and make it hard to maintain the rules.
Depending on the processor it might not work at all, since it could also change the event by a random value (hash/uuid) or it can't overwritte etc.

ekneg54:
ok... good point... then I have to dig a lot deeper keeping my fingers crossed to not get to a really ugly solution :-D

ppcad:
Thank you very much!

herrfeder (changes requested):
I tested locally with individiual `selective_extractor` configs on my host and the FileInputConnector.

### Incompability with console_output

The output type `console_output` doesn't work with the new implementation at all as it will raise the following error:

```
2023-02-16 09:11:08,030 Pipeline ERROR   : A critical error occurred for processor SelectiveExtractor (selective_extractor) when processing an event, processing was aborted: (AttributeError: module 'sys' has no attribute 'first_topic') (59 in ~29.0 sek)
```

Apparently this output connector doesn't like the topic attribute but ommiting `target_topic` isn't possible which is also confusing when using file outputs.

### Confusion about multi output selective extraction

According to my review comments I don't understand the actual intended test behaviour as I would assume with individual `selective_extractor` the affected outputs should be different.

### Suggestion for optimized dealing with multiple outputs through consecutive pipeline processors

And I guess it touches somehow @ppcad comments I don't see the possibility to create individual streams when using multi-outputs which leads to the described behaviour that for each additional output all rules will be executed again.
An easy but not perfect possibility to deal with it in my opinion would be to introduce (just like in Logstash) Meta-Fields that can be filtered. When addressing multiple or even new outputs in a multi-output aware processor a meta-field could be added `{"output":"output_stream_123"}` which can be filtered in the following processors in the pipeline. This would still touch the filter of each following processor but doesn't execute the full processor behaviour.

ppcad:
> I tested locally with individiual `selective_extractor` configs on my host and the FileInputConnector.
> 
> ### Incompability with console_output
> The output type `console_output` doesn't work with the new implementation at all as it will raise the following error:
> 
> ```
> 2023-02-16 09:11:08,030 Pipeline ERROR   : A critical error occurred for processor SelectiveExtractor (selective_extractor) when processing an event, processing was aborted: (AttributeError: module 'sys' has no attribute 'first_topic') (59 in ~29.0 sek)
> ```
> 
> Apparently this output connector doesn't like the topic attribute but ommiting `target_topic` isn't possible which is also confusing when using file outputs.

Yes, I have also noticed that. It works if `target_topic` is set to `stdout`.

> 
> ### Confusion about multi output selective extraction
> According to my review comments I don't understand the actual intended test behaviour as I would assume with individual `selective_extractor` the affected outputs should be different.
> 
> ### Suggestion for optimized dealing with multiple outputs through consecutive pipeline processors
> And I guess it touches somehow @ppcad comments I don't see the possibility to create individual streams when using multi-outputs which leads to the described behaviour that for each additional output all rules will be executed again. An easy but not perfect possibility to deal with it in my opinion would be to introduce (just like in Logstash) Meta-Fields that can be filtered. When addressing multiple or even new outputs in a multi-output aware processor a meta-field could be added `{"output":"output_stream_123"}` which can be filtered in the following processors in the pipeline. This would still touch the filter of each following processor but doesn't execute the full processor behaviour.

Extra data doesn't get passed to the next processor, so it needs to be generated again for each output if the goal is to send the same extra data to different outputs. To generate it the whole event needs to be processed again.

ekneg54:
It is a draft... I am not ready

ekneg54:
@ppcad @herrfeder  ... I'm ready now... please have a look

herrfeder (changes requested):
As we finished the review with at least 6 eyes gracefully, I can finish this review with minor request changes.

herrfeder (approved):

ppcad (approved):
Thank you for the changes!

ekneg54:
thank you for your approve. I will merge this, if we are ready. Because it will break our configuration.

PR #311, opened 2023-02-06
fix workflow file

ekneg54:


urdit52 (approved):
thx for fixing the variable name.

PR #312, opened 2023-02-09
Fix auto rule tester to call all processor setup methods

dtrai2:
The auto-rule-tester didn't load the tldextract lists properly because the processor setup method wasn't called in the auto-rule-tester. This pull requests fixes it by calling all setup methods and adjusting the list_comparision processor as well as the pseudonymizer to make proper use of the setup method.

### DIFF
--- a/logprep/processor/pseudonymizer/processor.py
+++ b/logprep/processor/pseudonymizer/processor.py
@@ -140,7 +140,6 @@ class Pseudonymizer(Processor):
         self.pseudonyms = []
         self.pseudonymized_fields = set()
         self.setup()
-        self._replace_regex_keywords_by_regex_expression()
 
     @cached_property
     def _url_extractor(self):
@@ -165,6 +164,7 @@ class Pseudonymizer(Processor):
         )
         self._init_tld_extractor()
         self._load_regex_mapping(self._config.regex_mapping)
+        self._replace_regex_keywords_by_regex_expression()
 
     def _init_tld_extractor(self):
         if self._config.tld_lists is not None:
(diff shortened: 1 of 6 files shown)


logprep/processor/pseudonymizer/processor.py
```diff
@@ -346,10 +346,12 @@ def _pseudonymize_value(self, value: str, pseudonyms: List[dict]) -> str:
     def _replace_regex_keywords_by_regex_expression(self):
         for rule in self._specific_rules:
             for dotted_field, regex_keyword in rule.pseudonyms.items():
-                rule.pseudonyms[dotted_field] = self._regex_mapping[regex_keyword]
+                if regex_keyword in self._regex_mapping:
```
herrfeder, line 349: For the sake of readability consider to explicitly check the keys of `self._regex_mapping` like this

```python
                if regex_keyword in self._regex_mapping.keys():

```

Otherwise if this is a suggested and faster way of doing this then ignore my suggestion.

logprep/processor/pseudonymizer/processor.py
```diff
@@ -346,10 +346,12 @@ def _pseudonymize_value(self, value: str, pseudonyms: List[dict]) -> str:
...
+                    rule.pseudonyms[dotted_field] = self._regex_mapping[regex_keyword]
         for rule in self._generic_rules:
             for dotted_field, regex_keyword in rule.pseudonyms.items():
-                rule.pseudonyms[dotted_field] = self._regex_mapping[regex_keyword]
+                if regex_keyword in self._regex_mapping:
```
herrfeder, line 353: For the sake of readability consider to explicitly check the keys of `self._regex_mapping` like this

```python
                if regex_keyword in self._regex_mapping.keys():

```

Otherwise if this is a suggested and faster way of doing this then ignore my suggestion.

herrfeder (changes requested):
All tests passed successfully while checking your branch.
I have two little suggestions for you.

logprep/processor/pseudonymizer/processor.py
ekneg54: the second suggestion will result in a pylint warning, because the first suggestion is the way to go ;-)

herrfeder: okay, sometimes I do not agree with pylint but that's the way the cookie crumbles

PR #313, opened 2023-02-09
Fix graceful shutdown of multiprocessing pipeline

dtrai2:


ekneg54 (changes requested):
you fixed something, but I see no tests to ensure, that it does not break again

dtrai2:
Yeah I am not sure yet how to test it, as the exception is printed by a python multiprocessing module instead of an actual raise. All I could do is check for sys out prints to ensure that the error is not there anymore. Will try that tomorrow.

ekneg54 (approved):
Thank you for the tests

PR #314, opened 2023-02-09
Dev speedup startup

ppcad:
Speedup start of Logprep

### DIFF
--- a/logprep/filter/lucene_filter.py
+++ b/logprep/filter/lucene_filter.py
@@ -110,6 +110,9 @@ class LuceneFilterError(BaseException):
 class LuceneFilter:
     """A filter that allows using lucene query strings."""
 
+    find_escaping_pattern = re.compile(r'(?:\\)+"')
+    perform_escaping_pattern = re.compile(r'((?:\\)+"[\s\)]*(?:AND|OR|NOT|$))')
+
     @staticmethod
     def create(query_string: str, special_fields: dict = None) -> FilterExpression:
         """Create a FilterExpression from a lucene query string.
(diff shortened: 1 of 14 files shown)


ppcad:
The failed test is from main and not from this pull-request.

ekneg54:
we added the `msgspec` dependency for improving input connector performance. It has yaml conversion too. Could we use this for performance reasons? So we could have a chance to use one and only one external dependency for loading json AND yaml in the near future.

ppcad:
> we added the `msgspec` dependency for improving input connector performance. It has yaml conversion too. Could we use this for performance reasons? So we could have a chance to use one and only one external dependency for loading json AND yaml in the near future.

I've just tested it. It is roughly as fast as the `impure_yaml`, but it doesn't seem to be able to load files that include multiple YAML documents.

dtrai2:
As I am not available in the next week I remove myself from this review. Don't think that I can do it in a timely manner. @ekneg54 you seem to have looked into it anyways already, hope you can take over.

ppcad:
> As I am not available in the next week I remove myself from this review. Don't think that I can do it in a timely manner. @ekneg54 you seem to have looked into it anyways already, hope you can take over.

Thank you for informing us.

logprep/filter/lucene_filter.py
```diff
@@ -132,30 +136,76 @@ def create(query_string: str, special_fields: dict = None) -> FilterExpression:
...
         try:
-            tree = parser.parse(query_string)
+            tree = parser.parse(escaped_string)
             transformer = LuceneTransformer(tree, special_fields)
-        except (ParseSyntaxError, IllegalCharacterError) as error:
-            raise LuceneFilterError(error)
+        except ParseSyntaxError as error:
+            msg = (
+                f"{error} Expression: '{escaped_string}'"
+                + " - Ensure that the expression is escaped correctly!"
```
ekneg54, line 147: Why so angry? ;-)
```suggestion
                + " - expression not escaped correctly."
```

logprep/filter/lucene_filter.py
```diff
@@ -132,30 +136,76 @@ def create(query_string: str, special_fields: dict = None) -> FilterExpression:
...
+            )
+            raise LuceneFilterError(msg) from error
+        except IllegalCharacterError as error:
+            msg = (
+                f"{error} in '{escaped_string}'"
+                + " - Ensure that the expression is escaped correctly!"
```
ekneg54, line 153: Again so angry :-D
```suggestion
                + " - expression not escaped correctly"
```

ekneg54 (approved):
only two minor remarks

PR #318, opened 2023-02-15
Apply processors multiple times

dtrai2:
It was previously not possible to apply the same processor to the current event more than one time. The processor strategy extracted all matching rules and applied them. If one rule enabled the execution of another rule it wasn't executed anymore as the retrieval of matching rules was only retrieved once. This pull request changes this and enables it to apply rules on results of previous rules.

One example could look like this:
Incoming event:
```json
{"message": "time [proto col] url"}
```
Dissector rule one:
```yaml
filter: message
dissector:
    mapping:
        message: "%{time} [%{protocol}] %{url}" 
```
Dissector rule two:
```yaml
filter: protocol
dissector:
    mapping:
        protocol: "%{proto} %{col}"
```
(note the filter of the second rule is set to an output of the first rule)
Expected result:
```json
{
    "message": "time [proto col] url",
    "proto": "proto",
    "col": "col",
    "protocol": "proto col",
    "time": "time",
    "url": "url",
}
```

Furthermore in doing so it was necessary to adjust the list_comparission tests. These tests weren't completely encapsulated because they had side effects of rules from other tests.


ekneg54:
isn't this testcase the needed behavior, isn't it?

```
    (
        "use dissected field in mapping",
        {
            "filter": "message",
            "dissector": {
                "mapping": {
                    "message": "%{time} [%{protocol}] %{url}",
                    "protocol": "%{proto} %{col}",
                },
            },
        },
        {"message": "time [proto col] url"},
        {
            "message": "time [proto col] url",
            "proto": "proto",
            "col": "col",
            "protocol": "proto col",
            "time": "time",
            "url": "url",
        },
    ),
```

this would be resolveable in the `dissector` processor itself instead of changing the whole processing strategy for all processors

dtrai2:
Well the dissector was obviously a bad example :smile: but the case is still true for processors like the normalizer or the requester. The requester could actually be an interesting case, because it is then possible to have chained requests based on the result of the previous one. Or execute another request in case the first one results in an error (tag). Or the StringSplitter, here you could split one string based on multiple delimiters after one another. So there are use cases for it, though I agree just for the dissector it seems not needed.

herrfeder:
Sorry for my late review. I never had a look at these parts of the codebase of Logprep and therefore I need some time to get into it and much more important to test it for myself. To be honest, I don't fully understand `RuleTree` yet. 
I checked your branch locally with the file input against my `/var/log/auth.log` and it has buggy behaviour as the "second level" dissector rule will also match the full message string which isn't desired I assume?! I checked it with two rules inside one dissector processor and isolating the rules into seperate dissector processors:
Filetree:
```
pipe_double_dissector
├── dissector_1
│   ├── generic
│   │   └── generic.yml
│   ├── regex_mapping.yml
│   └── specific
├── dissector_2
│   ├── generic
│   │   └── timestamp.yml
│   ├── regex_mapping.yml
│   └── specific
├── pipeline.yml
└── README.md
```

Config:
```yml
version: 1
process_count: 2
timeout: 0.1
logger:
  level: INFO

pipeline:
  - dissector1:
      type: dissector
      specific_rules:
        - logprep_pipes/pipe_double_dissector/dissector_1/specific
      generic_rules:
        - logprep_pipes/pipe_double_dissector/dissector_1/generic/
  - dissector2:
      type: dissector
      specific_rules:
        - logprep_pipes/pipe_double_dissector/dissector_2/specific
      generic_rules:
        - logprep_pipes/pipe_double_dissector/dissector_2/generic/

input:
  blah:
    type: file_input
    logfile_path: /var/log/auth.log
    interval: 1

output:
  first:
    type: console_output

```


Message:
```json
{"message": "Mar  9 12:30:01 dev-davlas CRON[886268]: pam_unix(cron:session): session closed for user root"}
```

First rule:
```yml
filter: "message"
dissector:
  mapping:      
    message: "%{timestamp}  %{+(\ )timestamp} %{+(\ )timestamp} %{hostname} %{process}: %{event}"
```

Second rule:
```yml
filter: "timestamp"
dissector:
  mapping:      
    message: "%{time}  %{stamp}"
```

Output:
```yml
{'event': 'pam_unix(cron:session): session closed for user root',
 'hostname': 'dev-davlas',
 'message': 'Mar  9 12:30:01 dev-davlas CRON[886268]: pam_unix(cron:session): session closed for user root',
 'process': 'CRON[886268]',
 'stamp': '9 12:30:01 dev-davlas CRON[886268]: pam_unix(cron:session): session closed for user root',
 'time': 'Mar',
 'timestamp': 'Mar 9 12:30:01'}
```

As you can see in the output. The `stamp` matches the full event again. And this does happen for multiple rules in one processor or splitting it into two processors like in the provided config.

logprep/processor/processor_strategy.py
```diff
@@ -64,10 +71,17 @@ def _process_generic(
...
-        for rule in generic_tree.get_matching_rules(event):
-            begin = time()
-            callback(event, rule)
-            processing_time = time() - begin
-            rule.metrics._number_of_matches += 1
-            rule.metrics.update_mean_processing_time(processing_time)
-            processor_metrics.update_mean_processing_time_per_event(processing_time)
+        applied_rules = set()
+        matching_rules = set(generic_tree.get_matching_rules(event))
+        while True:
```
herrfeder, line 76: As I stated in my comment, I don't fully understand the logic inside `RuleTree` yet but independently from the functionality, is it really necessary to have the exact same line of codes a second time? I see the `tree` dependent looping, but this could be an own `loop` function which will get only the reference to each tree.

herrfeder (changes requested):
The code comment is only related to code quality but please see my comment with testing of your branch. This seems to be more buggy than desired.

dtrai2:
> As you can see in the output. The `stamp` matches the full event again. And this does happen for multiple rules in one processor or splitting it into two processors like in the provided config.

Well if you look at your second dissector rule you are pointing it also to the `message` field. I think what you were planning to do was to point it just to the `timestamp` field like this:
```yaml
filter: "timestamp"
dissector:
  mapping:      
    timestamp: "%{time}  %{stamp}"
```
right? Can you maybe validate if this will lead to your expected behavior? 

And maybe as a general hint to the idea behind this pull-request. I saw that you added two dissectors into the pipeline. The idea of this change here is that you don't need two processors but can still apply multiple rules inside this one processor. So sometimes it is common to write a multi-rule in one file, like: 
```yaml
filter: "message"
dissector:
  mapping:      
    message: "%{timestamp}  %{+(\ )timestamp} %{+(\ )timestamp} %{hostname} %{process}: %{event}"
---
filter: "timestamp"
dissector:
  mapping:      
    timestamp: "%{time}  %{stamp}"
```
In the current logprep behavior the second rule wouldn't be applied as the field `timestamp` doesn't exist yet in the event. To make it work it would be needed to add a second dissector processor to the pipeline as you just did in your example. To prevent adding a second processor I opened this pull-request such that it is possible to process both rules in one run.

herrfeder:
yes, you're so right. My fault, works like a charm if rules configured correctly. Sorry I'm still learning while reviewing:

```yml
{
 'event': 'pam_unix(cron:session): session closed for user root',
 'hostname': 'dev-davlas',
 'message': 'Mar 10 08:30:01 dev-davlas CRON[3220988]: pam_unix(cron:session): session closed for user root',
 'process': 'CRON[3220988]',
 'stamp': '10 08:30:01',
 'time': 'Mar',
 'timestamp': 'Mar 10 08:30:01'
}
```

logprep/processor/processor_strategy.py
dtrai2: I think the reason we did this is so we can explicitly test the call order of the two functions `process_specific` and `process_generic`. But you are right it is not really nice to have the code exactly twice. I could do something like this:

```python
def _process_specific(
    self,
    event: dict,
    specific_tree: "RuleTree",
    callback: Callable,
    processor_metrics: "Processor.ProcessorMetrics",
):
    self._process_rule_tree(event, specific_tree, callback, processor_metrics)

def _process_generic(
    self,
    event: dict,
    generic_tree: "RuleTree",
    callback: Callable,
    processor_metrics: "Processor.ProcessorMetrics",
):
    self._process_rule_tree(event, generic_tree, callback, processor_metrics)

def _process_rule_tree(
    self,
    event: dict,
    tree: "RuleTree",
    callback: Callable,
    processor_metrics: "Processor.ProcessorMetrics",
):
    # tree looping stuff comes here
```
That way we have minimized the redundant code and we still have two explicit methods we can test for.

herrfeder: :+1: yes I prefer and like this much more as it increases readability also.

herrfeder (approved):

dtrai2:
Hey @herrfeder, I still have your approved, but I also made some more changes to the `pre_detector_acceptance` test as it wasn't really reliable anymore. The order of the pre_detector extra outputs flipped with every execution. I revised the whole test, could you have a look at my last three commits and check it again before we merge this pull-request.

PR #320, opened 2023-02-17
Fix typing in ip_informer processor

saegel:
When running the logprep auto tests, I get the following error:

```
File "logprep/processor/ip_informer/processor.py", line36, in IpInformer _processing_warnings: list[tuple[str, Exception]]
TypeError: 'type' object is not subscriptable
```

This PR fixes this error and allows the auto tests to run.

### DIFF
--- a/logprep/processor/ip_informer/processor.py
+++ b/logprep/processor/ip_informer/processor.py
@@ -20,7 +20,7 @@ Example
 from functools import partial
 import ipaddress
 from itertools import chain
-from typing import Iterable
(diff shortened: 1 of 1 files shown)


ekneg54:
I wonder for this error. Because the code is tested very well. So let me ask you which python version do you have?

saegel:
I was running Python 3.8.

PEP 585 explains the behavior.
I'll try to get my Python version updated.

ekneg54:
closed -> won`t fix

PR #325, opened 2023-02-27
make properties configurable in ip_informer rule

ekneg54:


tests/unit/processor/ip_informer/test_ip_informer.py
```diff
@@ -183,6 +184,46 @@
...
+                "target_field": "result",
+                "properties": ["is_loopback"],
+            },
+        },
+        {"ip": "192.168.5.1"},
+        {
+            "ip": "192.168.5.1",
+            "result": {
+                "192.168.5.1": {
+                    "is_loopback": False,
```
urdit52, line 202: would it make more sense to test for a property that is != False in order to differentiate from the second test case?

urdit52 (changes requested):
2 remarks.

tests/unit/processor/ip_informer/test_ip_informer.py
ekneg54: this testcase tests that it should be possible to only extract a single property. 
the second testscase tests, that it should be possible to extract a properte which does not exist for a ipv4address.
the value of the fields does not matter for the first testcase.

urdit52: fair enough.

urdit52 (approved):
thank you for the clarification and modification. looks good to me.

PR #328, opened 2023-02-28
Add shorthand to dissect pattern for converting datatypes

ekneg54:
* adds datatype conversion shorthand syntax to dissect pattern language
* adds bool datatype conversion that translates to bool by meaning

### DIFF
--- a/logprep/processor/dissector/rule.py
+++ b/logprep/processor/dissector/rule.py
@@ -60,6 +60,11 @@ It is possible to capture the target field name from the source field value with
 this can be referred to with the notation :code:`%{&<the reference>}` (e.g. :code:`%{&key1}`).
 References can be combined with the append operator.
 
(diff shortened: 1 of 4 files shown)


logprep/processor/dissector/rule.py
```diff
@@ -84,25 +89,41 @@
...
+)
 
 
 def _do_nothing(*_):
     return
 
 
+def str_to_bool(input_str: str) -> bool:
+    """converts an input string to bool by meaning"""
+    if input_str.lower() in ("yes", "true", "on", "y"):
```
sashs, line 118: I would add a check for None to ensure that the object input_str is not None before calling lower

ekneg54: indeed, but I resolved it by asking for forgiveness, because the case should occur rarely often.

sashs (approved):

PR #334, opened 2023-03-06
fix opensearch-output-connector-raises-to-late-if-unauthenticated 

ekneg54:
* adds blacklist variable names to enrichment from environment, because we use a special variable LOGPREP_LIST in list_comparison
* fixes docker build for branch and not for main for PRs
* adds connection check for opensearch and elasticsearch output connectors

### DIFF
--- a/doc/source/user_manual/configuration/configurationdata.rst
+++ b/doc/source/user_manual/configuration/configurationdata.rst
@@ -22,7 +22,8 @@ Details for configuring connectors are described in :ref:`output` and :ref:`inpu
 General information about the configuration of the pipeline can be found in :ref:`pipeline_config` .
 
 It is possible to use environment variables in all configuration and rules files in all places.
-Environment variables have to be set in uppercase. Lowercase variables are ignored.
+Environment variables have to be set in uppercase. Lowercase variables are ignored. Forbidden
+variable names are: :code:`["LOGPREP_LIST"]`
 
 The following config file will be valid by setting the given environment variables:
 
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -60,7 +60,7 @@ class Getter(ABC):
             var
             for var in {*used_env_vars, *used_braced_env_vars}
             if var not in BLOCKLIST_VARIABLE_NAMES
-        ]  # deduplicate and clean blacklisted values
+        ]  # deduplicate and clean blocklisted values
 
     def get_yaml(self) -> Union[Dict, List]:
         """gets and parses the raw content to yaml"""
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -10,6 +10,12 @@ from attrs import define, field, validators
 
 yaml = YAML(typ="safe", pure=True)
 
+BLACKLIST_VARIABLE_NAMES = [
+    "",
+    " ",
+    "LOGPREP_LIST",  # used by list_comparison processor
+]
+
 
 @define(kw_only=True)
 class Getter(ABC):
@@ -51,8 +57,10 @@ class Getter(ABC):
         used_env_vars = map(lambda x: x[1], found_variables)
         used_braced_env_vars = map(lambda x: x[2], found_variables)
         return [
-            var for var in {*used_env_vars, *used_braced_env_vars} if var
-        ]  # deduplicate and clean out whitespace
+            var
+            for var in {*used_env_vars, *used_braced_env_vars}
+            if var not in BLACKLIST_VARIABLE_NAMES
+        ]  # deduplicate and clean blacklisted values
 
     def get_yaml(self) -> Union[Dict, List]:
         """gets and parses the raw content to yaml"""
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -10,7 +10,7 @@ from attrs import define, field, validators
 
 yaml = YAML(typ="safe", pure=True)
 
-BLACKLIST_VARIABLE_NAMES = [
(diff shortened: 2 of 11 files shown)


doc/source/user_manual/configuration/configurationdata.rst
```diff
@@ -22,7 +22,8 @@ Details for configuring connectors are described in :ref:`output` and :ref:`inpu
 General information about the configuration of the pipeline can be found in :ref:`pipeline_config` .
 
 It is possible to use environment variables in all configuration and rules files in all places.
-Environment variables have to be set in uppercase. Lowercase variables are ignored.
+Environment variables have to be set in uppercase. Lowercase variables are ignored. Forbidden
+variable names are: :code:`["LOGPREP_LIST"]`
```
dtrai2, line 26: We never really discussed a pattern for the envs but should we maybe rename this to something like `LOGPREP_LISTCOMPARISON_LIST`? Ah though I am realizing now that this would be a breaking change...

logprep/abc/getter.py
```diff
@@ -10,6 +10,12 @@
 
 yaml = YAML(typ="safe", pure=True)
 
+BLACKLIST_VARIABLE_NAMES = [
```
dtrai2, line 13: Could you name this to `BLOCKLIST`?

tests/unit/util/test_getter.py
```diff
@@ -108,6 +108,14 @@ def test_getter_expands_only_uppercase_variable_names_with_braced_variables(self
         my_getter = GetterFactory.from_string(str(testfile))
         assert my_getter.get() == "this is my mytoken, and this is my ${not_a_token}"
 
+    def test_getter_ignores_list_comparison_logprep_list_variable(self, tmp_path):
+        os.environ.update({"PYTEST_TEST_TOKEN": "mytoken"})
```
dtrai2, line 112: Shouldn't the environment have the variable `LOGPREP_LIST` for this test? After all, you want to test here that the blocklist is working, don't you?

dtrai2 (changes requested):

doc/source/user_manual/configuration/configurationdata.rst
ekneg54: yes... it would be a breaking change... and I would not touch it here

logprep/abc/getter.py
ekneg54: why? it behaves like a blacklist. so I called it blacklist :-)

tests/unit/util/test_getter.py
ekneg54: yes... that is what I test. And for that, the LOGPREP_LIST should not be defined in the actual environment. so that it persists in the output to ensure this is handled by the list_comparison

logprep/abc/getter.py
ekneg54: done

tests/unit/util/test_getter.py
dtrai2: ah I still don't fully understand this. But what if a user tries to set the env variable `LOGPREP_LIST` for some reason?

logprep/abc/getter.py
```diff
@@ -51,8 +57,10 @@ def _get_used_env_vars(self, content, template):
         used_env_vars = map(lambda x: x[1], found_variables)
         used_braced_env_vars = map(lambda x: x[2], found_variables)
         return [
-            var for var in {*used_env_vars, *used_braced_env_vars} if var
-        ]  # deduplicate and clean out whitespace
+            var
+            for var in {*used_env_vars, *used_braced_env_vars}
+            if var not in BLOCKLIST_VARIABLE_NAMES
+        ]  # deduplicate and clean blacklisted values
```
dtrai2, line 63:
```suggestion
        ]  # deduplicate and clean blocklisted values
```

tests/unit/util/test_getter.py
ekneg54: then it is ignored as stated in the documentation

dtrai2: but it's not :D If I add the `LOGPREP_LIST` variable into the test like:
```
os.environ.update({"PYTEST_TEST_TOKEN": "mytoken", "LOGPREP_LIST": "foo"})
```
Then test fails and I get the following error: 
```
AssertionError: assert 'this is my mytoken, and this is my foo' == 'this is my mytoken, and this is my ${LOGPREP_LIST}'
```

ekneg54: you are right... I have to fix it

ekneg54: done

tests/unit/util/test_getter.py
```diff
@@ -108,6 +108,23 @@ def test_getter_expands_only_uppercase_variable_names_with_braced_variables(self
...
+        testfile = tmp_path / "test_getter.json"
+        testfile.write_text("this is my ${PYTEST_TEST_TOKEN}, and this is my ${LOGPREP_LIST}")
+        my_getter = GetterFactory.from_string(str(testfile))
+        assert my_getter.get() == "this is my mytoken, and this is my ${LOGPREP_LIST}"
+        assert len(my_getter.missing_env_vars) == 0
+
+    def test_getter_ignores_list_comparison_logprep_list_variable_if_setted(self, tmp_path):
```
dtrai2, line 119:
```suggestion
    def test_getter_ignores_list_comparison_logprep_list_variable_if_set(self, tmp_path):
```

dtrai2 (approved):
Just one small typo left, but looks good now. Thanks for fixing it.

PR #337, opened 2023-03-08
remove verification of configuration in pipeline manager and run_logprep

ekneg54:
the configuration is verified already here:

https://github.com/fkie-cad/Logprep/blob/49303b87ee622502f90a9a220f0d3f52653317ac/logprep/runner.py#L133-L157

on reload the configuration is verified again here to prevent faulty config from loading:

https://github.com/fkie-cad/Logprep/blob/49303b87ee622502f90a9a220f0d3f52653317ac/logprep/runner.py#L236-L256

I think there is no need for this configuration verification in the pipeline manager. this will improve logprep start times

additionally the config was checked on logprep start in `run_logprep` so the config was checked 3 times for one start and in every verification the whole pipeline was instantiated and all rules were loaded.

### DIFF
--- a/logprep/framework/pipeline_manager.py
+++ b/logprep/framework/pipeline_manager.py
@@ -38,8 +38,7 @@ class PipelineManager:
         self._used_server_ports = None
 
     def set_configuration(self, configuration: Configuration):
-        """Verify the configuration and set it in the pipeline manager."""
-        configuration.verify(self._logger)
+        """set the verified config"""
         self._configuration = configuration
 
         manager = Manager()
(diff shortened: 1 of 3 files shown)


ppcad (changes requested):
Thank you, it's a lot faster now.
However, `--verify-config` does now succeed even if the config is invalid.

ekneg54:
I reimplemented the missing feature and add tests, to ensure this never happens again

ppcad (approved):
Thank you for the changes

PR #343, opened 2023-03-10
Fix catched exception in tree parser that led to ignoring of rules

ppcad:
Sometimes the tree parser tries to remove to many elements from the expression when parsing.
In that case it prints a warning and ignores that rule. However, if started with the auto-tester it doesn't even print a warning.
This can be reproduced by having a conjunction of at least four disjunctions.

This pull request prevents the crash that leads to the warning so that the rules won't be ignored.


logprep/framework/rule_tree/rule_parser.py
ekneg54, line 1: we should really rewrite this to make it readable for others. I have no clue, what is happening.

ekneg54 (changes requested):
could you please rewrite this parser? it is one of the worst readable code fragments in logprep and it would be a real improvement, if someone rewrites it so it is more readable.

logprep/framework/rule_tree/rule_parser.py
ppcad: This will be a bigger project, this pull request is just meant to fix the bug.

logprep/filter/expression/filter_expression.py
```diff
@@ -84,7 +84,20 @@ def __eq__(self, other):
...
+        key_list : List[str]
+            List of keys.
+
+        Returns
+        -------
+        str
+            Returns dotted string.
+
+        """
         return ".".join([str(i) for i in key_list])
```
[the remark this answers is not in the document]
ppcad: It's probably not needed then. It's not related to this pull-request and I would rather not change to many thing at once in the tree-parser in this pull-request.

logprep/framework/rule_tree/rule_parser.py
ekneg54: I added a story for that: #376

ekneg54 (approved):

PR #346, opened 2023-03-15
fix non-deterministic behavior and add test

dtrai2:
The processor strategy applied rules in a random order due to using `set` at the wrong locations. This pull request moves the `set` operation to where it is really needed.

### DIFF
--- a/tests/unit/processor/test_processor_strategy.py
+++ b/tests/unit/processor/test_processor_strategy.py
@@ -8,9 +8,8 @@ import pytest
 
 from logprep.abc.processor import Processor
 from logprep.factory import Factory
-from logprep.framework.rule_tree.rule_tree import RuleTree
 from logprep.processor.dissector.rule import DissectorRule
-from logprep.processor.normalizer.rule import NormalizerRule
+from logprep.processor.generic_adder.rule import GenericAdderRule
 from logprep.processor.processor_strategy import SpecificGenericProcessStrategy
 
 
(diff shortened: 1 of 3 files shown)


tests/unit/processor/test_processor_strategy.py
```diff
@@ -89,3 +93,40 @@ def test_apply_processor_multiple_times_until_no_new_rule_matches(self):
...
         )
         assert expected_event == event
+
+    @pytest.mark.parametrize("execution_number", range(5))  # repeat test to ensure determinism
+    def test_strategy_applies_rules_in_deterministic_order(self, execution_number):
+        generic_tree = RuleTree()
+        specific_tree = RuleTree()
+        rule_one_dict = {"filter": "field", "normalize": {"field.one": "copy.one"}}
+        rule_two_dict = {"filter": "NOT field.two", "normalize": {"field.one": "field.two"}}
+        rule_one = NormalizerRule._create_from_dict(rule_one_dict)
```
ekneg54, line 103: please do not use the Normalizer rule here. It is replaced in near future

tests/unit/processor/test_processor_strategy.py
```diff
@@ -89,3 +93,40 @@ def test_apply_processor_multiple_times_until_no_new_rule_matches(self):
             processor_metrics=mock_metrics,
         )
         assert expected_event == event
+
+    @pytest.mark.parametrize("execution_number", range(5))  # repeat test to ensure determinism
+    def test_strategy_applies_rules_in_deterministic_order(self, execution_number):
+        generic_tree = RuleTree()
+        specific_tree = RuleTree()
```
ekneg54, line 100: the processor has already his trees

ekneg54 (changes requested):

tests/unit/processor/test_processor_strategy.py
dtrai2: I changed the test to make use of the generic_adder this led to some problems with the comparison of the rule due to the `base/rule.py` hash and repr functions. I had to add the objects `id` to the hash function to make a unique hash per object. Without that either all test repetitions are always green or red (in the negative case when the bugfix is not present in the processor_strategy).

tests/unit/processor/test_processor_strategy.py
dtrai2: I removed them, I did this also in the neighboring test and revised it a bit.

ekneg54 (approved):

PR #349, opened 2023-03-16
only expand valid prefixed variables

ekneg54:


ppcad (changes requested):
Thanks for the quick fix, it works.
It does however reduce $ to $$ even if it is not a valid environment variable. So if one happens to have two or more consecutive $ it will always remove one, even if they are free standing.
Furthermore, I suggest to add _ to each valid prefix. So that it would use $CI_FOO instead of $CIFOO. Escpecially, since many variables could start with CI.

ekneg54:
I fixed it. But it was not possible to completely disable the escape sequence. Instead I set it to "$$$". I know it is a strange workaround, but we could give it a try :-)

PR #355, opened 2023-03-24
Make multiple applications of rules by the same processor optional

ppcad:
Applying rules multiple times requires to match the rules at least 2 times.
Once for the 1. match and once to check if another match is necessary.
This can reduce the performance of Logprep significantly.
Therefore, this feature is made optional with this pull-request.
It is disabled per default and can be enabled on a processor basis by setting `apply_rules_multiple_times` to true.

### DIFF
--- a/logprep/processor/processor_strategy.py
+++ b/logprep/processor/processor_strategy.py
@@ -5,6 +5,7 @@ processor strategies are used to implement in one point how rules are processed 
 this could be the order of specific or generic rules
 """
 from abc import ABC, abstractmethod
+from functools import reduce
 from time import time
 from typing import Callable, TYPE_CHECKING
 
@@ -31,8 +32,8 @@ class SpecificGenericProcessStrategy(ProcessStrategy):
     specific_rules >> generic_rules
     """
 
-    def __init__(self, apply_rules_multiple_times=False):
-        self._apply_rules_multiple_times = apply_rules_multiple_times
+    def __init__(self, apply_multiple_times=False):
+        self._apply_multiple_times = apply_multiple_times
 
     def process(self, event: dict, **kwargs):
         specific_tree = kwargs.get("specific_tree")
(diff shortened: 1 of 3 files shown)


ekneg54:
I think, we should remove this option completely, because it has a big impact on performance. I'm not sure if someone wants this, if it is ever faster to use a processor multiple times in the pipeline configuration.

@dtrai2 what do you think?

logprep/processor/processor_strategy.py
ekneg54, line 1: this was the fastest implementation I come along with:

```python

    def _process_rule_tree(
        self,
        event: dict,
        tree: "RuleTree",
        callback: Callable,
        processor_metrics: "Processor.ProcessorMetrics",
    ):
        applied_rules = set()

        def _process_rule(event, rule):
            begin = time()
            callback(event, rule)
            processing_time = time() - begin
            rule.metrics._number_of_matches += 1
            rule.metrics.update_mean_processing_time(processing_time)
            processor_metrics.update_mean_processing_time_per_event(processing_time)
            applied_rules.add(rule)
            return event

        matching_rules = tree.get_matching_rules(event)
        while matching_rules:
            reduce(_process_rule, (event, *matching_rules))
            if not self._apply_rules_multiple_times:
                break
            matching_rules = set(tree.get_matching_rules(event)).difference(applied_rules)

```

ekneg54: testcode:

```python
from timeit import timeit

setup_code = """
import json
from pathlib import Path
from logprep.framework.pipeline import Pipeline
from logprep.util.configuration import Configuration

input_data = Path("quickstart/exampledata/input_logdata/test_input.jsonl").read_text().splitlines()
input_data = [json.loads(item) for item in input_data]
config = Configuration.create_from_yaml("quickstart/exampledata/config/pipeline.yml")
del config["output"]
config.update({"input": {"dummy": {"type": "dummy_input", "documents": input_data}}})
pipeline = Pipeline(config=config)
"""

profile_code = """
for _ in input_data:
    parsed_event, extra_outputs = pipeline.process_pipeline()
"""

timeit(stmt=profile_code, setup=setup_code, number=100000000)
```

dtrai2:
Well I did implement it for a reason after all so I would like to keep it. But I like both suggestions. So I will add a flag such that the end user can enable/disable it whenever they need it and I will also consider the performance boost. Another idea I have right now is to implement it as a completely different strategy. After all we did choose the pattern of strategies, but so far we only have one. So we could use this also as an option to implement a second strategy such that there is a global config where the user could change the complete strategy. If I'd do it like this though I think it doesn't make sense anymore to have a per-processor switch then.  
@ppcad @ekneg54 what do you guys think? Add as a second strategy or extend the current strategy with a per-processor based on/off switch?

ppcad:
I like boot solutions proposed by @ekneg54, but I would chose the fastest.
@dtrai2 I would make it per processor, since the performance can vary greatly depending on the rule set a processor is using and there is no reason to activate it if you know that your rules never build upon each-other for one processor.
I would however like to merge the existing solution by @ekneg54 for now, so that logprep is in a performant state as soon as possible.

ekneg54:
I would implement a second strategy and make it configurable per processor and globally, by which the nearest configuration wins. 
for example, if you configure a strategy globally, you can overwrite it for each processor.
As default I would prefer the the old behavior, before we implemented running rules multiple times.

ekneg54:
hint: non of my suggestions boosts performance significantly. I suggest to choose the most readable.

logprep/processor/processor_strategy.py
dtrai2: I just tested these versions and interestingly my original version is on my pc most of the time the fastest.

dtrai2:
So in order to get it quickly back into the main, I have just applied @ekneg54's suggestion regarding the faster implementation. So the separate strategy with a global and per-processor based configuration we can still do later when the actual need arises. And yes the default value is that it is turned off now.

dtrai2 (approved):

ppcad:
> So in order to get it quickly back into the main, I have just applied @ekneg54's suggestion regarding the faster implementation. So the separate strategy with a global and per-processor based configuration we can still do later when the actual need arises. And yes the default value is that it is turned off now.

Thank you!

ekneg54:
@Malutthias still want to review it, we have the needed approvals

Issue #356, opened 2023-03-24
Logprep doesn't work with legacy SSL certificates


ppcad:
confluent-kafka 2.0.2 uses OpenSSL 3.0.x, which is incompatible with some older certificates.
Thus no connections to Kafka are possible if using such certificates.
It should be possible to configure Logprep to allow legacy certificates.

**Steps to reproduce**
Run Logprep with a kafka connector using SSL.

Logprep wil log warnings that it can not verify the SSL certificate and it will notread/write any events.

**Environment**

Logprep version: c549a3c946d11762f25ac6f9bfbb01665efe2b77
Python version: 3.9

ekneg54:
as suggested here: https://github.com/confluentinc/confluent-kafka-python/issues/1521 the best way to solve this is to update the certificates to use stronger ciphers as weak ciphers are disabled by default. 

as a workaround we could make the librdkafka option `ssl.provider` configurable. But if we do this, it could be a better solution to decouple the kafka configuration completely to give all configuration options in users hands.

ekneg54:
I added the possibility to give all kafka config options in #357 
With this you have the chance to find a configuration that fits to your usecase without code changes.

ppcad:
Thank you! I will test this out as soon as possible.

Issue #163, opened 2022-08-19
Application maximum poll interval (300000ms) exceeded by XXms


trimborn:
Since a few days i encounter the problem, that the logprep application dos produce the following error at night:
"Application maximum poll interval (300000ms) exceeded by 13ms"

Currently my workaround is to restart the logprep process entirely until the issue appears again since the maximum poll interval is also not configurable as of now via the config file.

**Expected behaviour**
The expected behaviour would be to reconnect to kafka

**Current behaviour**
After the error logprep does nothing

**Steps to reproduce**

Not entirely sure, it looks as if there are not enough logs to work on, the gap between active connections is to big.

**Environment**

Logprep version: main version
Python version: 3.6

**Possible solution**
The error gets raised by the rdkafka library. The following issue is similar to mine:
https://github.com/confluentinc/confluent-kafka-python/issues/552
The issue is that the rdkafka library does not reconnect to kafka after raining the error as stated in the issue above.
A solution would be to run the "poll" function once every maximum poll interval since according to the issue above this reconnects the kafka connector again to the cluster

alternatively to make the maximum poll time configurable

PR #357, opened 2023-03-24
add kafka advanced config option

ekneg54:


ekneg54:
closes #163 
closes #356

ppcad (changes requested):
Thanks, this works, but there is a problem when adding a value that requires a dict and not a string, like `offset_reset_policy`.

Furthermore, setting the advanced kafka configuration does completely overwrite the existing configuration, so that all configuration must be done there and the validator does still require the existence of the other configuration parameters, which means that the configuration is being doubled.

Could this be changed, so that the validator does not require both configs or that only existing parameters of the `kafka_config` are being used to replace the old config?

ekneg54:
and we had some struggles with the Changelog, wich were not updated possibly in cause of merge conflicts

ppcad (approved):
Thank you for the chages!

PR #359, opened 2023-03-27
Improve registry performance

dtrai2:
This pull request introduces a more dynamic importing of the logprep components.

Previously every logprep start imported literally all components, even those which weren't used due to the configuration. This lead to an unnecessary overhead for importing classes and their corresponding used classes/methods (e.g elasticsearch, opensearch, or hyperscan,...) even though they were unimportant for the configured run. 

This is not a huge performance improvement, but it should still improve the startup a little bit.

### DIFF
--- a/logprep/configuration.py
+++ b/logprep/configuration.py
@@ -1,11 +1,11 @@
 """module for component configuration """
 from typing import TYPE_CHECKING, Any, Mapping
 
-from logprep.registry import Registry
 from logprep.factory_error import (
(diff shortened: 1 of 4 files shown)


ekneg54 (commented):
I have a bad feeling with this. From developer perspective it is much more to write now. From a security perspective I'm not fine with importing from arbitrary strings and from performance perspective as you say it has not a big impact.

Is there another perspective I'm not aware of for this change?

PR #360, opened 2023-03-29
fix renaming duplication error

ekneg54:
main is failing because of that. I forgot to rebase my branch before it was merged


ekneg54:
will be done with #357

PR #362, opened 2023-04-01
avoid string splitting during processing

ekneg54:
this is a PR to improve logprep performance. the main idea behind this is, that splitting of string is an expensive operation and we could cache this easy. @ppcad could you please double check if this is really an improvement in performance. my local setup says it does.


ppcad (approved):
Thank you for the improvement

PR #370, opened 2023-04-16
improve connector performance

ekneg54:
use `msgspec.json` instead of builtin `json` module. This should lead to a significant performance boost in input and output connectors.

### DIFF
--- a/logprep/abc/component.py
+++ b/logprep/abc/component.py
@@ -3,6 +3,7 @@ from abc import ABC
 from logging import Logger
 from typing import Callable
 
+import msgspec
 from attr import define, field, validators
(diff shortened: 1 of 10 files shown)


ppcad (changes requested):
The tests pass, but a FatalOutputError is raised in Opensearch if Logprep is run. It is unable to serialize the events. I assume this also happens for Elasticsearch.

ekneg54:
thank you... I will investigate it

ekneg54:
I reverted the change for the open- and elasticsearch error document building.

ppcad (approved):
Thanks for the changes, it works now!

PR #371, opened 2023-04-19
Remove all geoip enricher fields if they are None

ppcad:
Remove all geoip fields that are `None` from the geoip results, so that the results for different events always have the same data types. This was missing for the geoip enricher coordinates.

### DIFF
--- a/logprep/processor/geoip_enricher/processor.py
+++ b/logprep/processor/geoip_enricher/processor.py
@@ -26,7 +26,7 @@ from geoip2 import database
 from geoip2.errors import AddressNotFoundError
 
 from logprep.abc.processor import Processor
-from logprep.processor.base.exceptions import FieldExistsWarning
+from logprep.processor.base.exceptions import FieldExistsWarning, ProcessingWarning
 from logprep.processor.geoip_enricher.rule import GEOIP_DATA_STUBS, GeoipEnricherRule
 from logprep.util.getter import GetterFactory
 from logprep.util.helper import add_field_to, get_dotted_field_value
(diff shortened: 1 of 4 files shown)


logprep/processor/geoip_enricher/processor.py
```diff
@@ -88,6 +89,20 @@ def _try_getting_geoip_data(self, ip_string):
...
+            if ip_data.location.longitude and ip_data.location.latitude:
+                geoip_data.update(
+                    {
+                        "geometry.type": "Point",
+                        "geometry.coordinates": [
+                            ip_data.location.longitude,
+                            ip_data.location.latitude,
+                        ],
+                    }
+                )
```
ekneg54, line 102:
```suggestion
                geoip_data |=  {
                        "geometry.type": "Point",
                        "geometry.coordinates": [
                            ip_data.location.longitude,
                            ip_data.location.latitude,
                        ],
                    }
                
```

logprep/processor/geoip_enricher/processor.py
```diff
@@ -70,13 +70,14 @@ def _try_getting_geoip_data(self, ip_string):
         try:
             ip_addr = str(ip_address(ip_string))
             ip_data = self._city_db.city(ip_addr)
+
+            if ip_string is None:
+                return {}
```
ekneg54, line 75: I think it is too late here... Consider to do this check in the `apply_rules` method

ekneg54: And I think if this string is None, we should raise a `ProcessingWarning` because the user thought there should be a value, but there is no value or the filter is not right. However, the user should know this.

ekneg54 (changes requested):
Here are some remarks from me, because @dtrai won`t come to it in the next weeks.

* Please add tests for your new implemented behavior, if you think that it should survive the next refactoring ;-)
* Consider to break out the further processing of None fields as early as possible. 
* Please add the changed behavior to the `CHANGELOG.md` file

ppcad:
Thanks for the feed-back. I changed it and realized that the geoip enricher did not need as many changes as I first thought. It was just missing the removal of geoip coordinates fields with `None` values.

PR #373, opened 2023-04-20
simplify quickstart

ekneg54:
* Rewrite quickstart setup:
  * Remove logstash, replace elasticsearch by opensearch and use logprep opensearch connector to stick to reference architecture
  * Use kafka without zookeeper and switch to bitnami container images


quickstart/docker-compose.yml
```diff
@@ -4,106 +4,86 @@ version: "3.9"
...
     networks:
-      elastic:
-        ipv4_address: 172.21.0.5
+      - opensearch-net
     environment:
-      KAFKA_ADVERTISED_HOST_NAME: kafka
-      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://172.21.0.5:9092
-      KAFKA_CREATE_TOPICS: "consumer:1:1,producer:1:1,producer_error:1:1,pseudonyms:1:1,sre:1:1"
-      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      - KAFKA_ENABLE_KRAFT=yes
```
herrfeder, line 56: this setting seems to require another Environment Variable `KAFKA_CFG_NODE_ID`, when I'm trying to run this docker-compose deployment with a fresh bitnami/kafka:latest I'm getting the following error and Kafka will exit:

image

herrfeder (changes requested):
Opensearch :heavy_check_mark: looks good, as you removed opensearch security completely for this test deployment, which is legitimate for this purpose, there is nothing that is missing that would otherwise do

Kafka :x: As I pointed out in the comment, for Kafka there seems to be missing a Environment Variable. I don't know if I did something wrong or maybe this requirement was baked into the image recently but it won't run this way on my machine.

ekneg54:
indeed they come with a newer version the last days. I fixed the NODE ID issue and pinned the kafka version to ensure, this never happens again.

quickstart/docker-compose.yml
ekneg54: fixed

quickstart/docker-compose.yml
```diff
@@ -4,106 +4,87 @@ version: "3.9"
...
+      - opensearch-net
     profiles:
       - logprep
     depends_on:
       - kafka
+    volumes:
+      - ./exampledata/:/home/logprep/
+    entrypoint:
+      - logprep
+      - /home/logprep/config/pipeline.yml
```
herrfeder, line 83: as the path inside the container doesn't start at quickstart anymore the config cannot resolve correctly

image

Would be the easiest to change the mount and the entrypoint that it will map and resolve correctly.

herrfeder (changes requested):
the logprep config doesn't resolve correctly because of the mounting path in logprep container

quickstart/docker-compose.yml
herrfeder: with Logprep not startet in the container from correct location it works.

herrfeder (approved):

herrfeder:
please see my suggestion for the readme change in my commit in this branch.

ekneg54:
thx for your commit. I removed zookeeper from readme too as the point for this simplification was to remove zookeeper from the configuration

PR #379, opened 2023-04-27
mark not used fields with init=False, repr=False, eq=False

ekneg54:
* mark inherited but not used fields in rules with `init=False, repr=False, eq=False` to signal, that they do not need to be rendered in external UI tools.


ekneg54:
funny... the container build pipelines are failing because of the branch name :-D
plz ignore it

sashs (approved):

PR #381, opened 2023-05-05
downgrade and pin confluent-kafka to 1.8.2

ekneg54:


ekneg54:
as seen here the donwgrade is not easy, because there are no packages for version 1.8.2 for python 3.11 so it has to be compiled during install which leads to install an external librdkafka which is not hold in all repositories. (we have no apt repositories in our build environment)

PR #382, opened 2023-05-08
Fix packaging of logprep to include grok default pattern

dtrai2:
The MANIFEST.in didn't include the grok pattern, such that logprep, if installed through pypi, didn't include the grok patterns.

I did test this through the [pypi test server](https://test.pypi.org/project/logprep/6.1.0.dev0/#files). It can be verified by running the following command in a new venv:

```bash
pip install --no-deps -i https://test.pypi.org/simple/ logprep==6.1.0.dev1
```
Afterwards you can check the installed site-packages of the venv. There you should find the logprep install with the grok patterns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 
 ### Bugfix
+* Fix missing default grok patterns in packaged logprep version
 
 
 ## v6.1.0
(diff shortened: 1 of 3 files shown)


ekneg54 (approved):
Looks good

PR #386, opened 2023-05-10
Fix confluent kafka 2 store offset

dtrai2:


dtrai2:
@ppcad can you verify if this current solution is working for you? In my local setup where I am reproducing and fixing the bug it appears to work. Would be great if you could check it. Especially also in regards to any possible data loss. On my side it seems to be good.

ppcad:
I'm testing if the error is gone. Or should I wait, since this is a draft?

Did you check if there is any data loss? If so, how?
As I see it this should work, since we set the offsets via the last message of each partition and those messages contain the partition offset and the partition ID (https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#message).

dtrai2:
> I'm testing if the error is gone. Or should I wait, since this is a draft?
> 
> Did you check if there is any data loss? If so, how? As I see it this should work, since we set the offsets via the last message of each partition and those messages contain the partition offset and the partition number (https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#message).

No please go ahead and check already. I forgot to marke it as "ready for review". 
Well I just had a small local kafka with fixed number of messages that had a counter in them and I used the logprep console output printing them out. I tried a few different runs with different consumer names and it appeared to me that there weren't any messages lost. But I didn't write them anywhere and checked in depth.

ppcad:
> > I'm testing if the error is gone. Or should I wait, since this is a draft?
> > Did you check if there is any data loss? If so, how? As I see it this should work, since we set the offsets via the last message of each partition and those messages contain the partition offset and the partition number (https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#message).
> 
> No please go ahead and check already. I forgot to marke it as "ready for review". Well I just had a small local kafka with fixed number of messages that had a counter in them and I used the logprep console output printing them out. I tried a few different runs with different consumer names and it appeared to me that there weren't any messages lost. But I didn't write them anywhere and checked in depth.

Ok, thank you.

ppcad (approved):

PR #387, opened 2023-05-10
Grokker: fix handling of missing source fields

dtrai2:


logprep/processor/grokker/processor.py
```diff
@@ -57,6 +57,10 @@ def _apply_rules(self, event: dict, rule: GrokkerRule):
         matches = []
         for dotted_field, grok in rule.actions.items():
             field_value = get_dotted_field_value(event, dotted_field)
+            if field_value is None:
+                error = BaseException(f"{self.name}: missing source_field: '{dotted_field}'")
```
ekneg54, line 61: Is base exception here reals needed?
I would prefer Processing Warning here to be consice to the handler

ekneg54 (changes requested):

logprep/processor/grokker/processor.py
dtrai2: Hehe I stole it from another processor, I think it doesn't really matter because inside the `handle_warning_error` it is wrapped in a `ProcessorWarning` anyways.

ekneg54 (approved):

PR #388, opened 2023-05-11
Extend fieldmanager rule to support writing into multiple targets in one rule

dtrai2:


logprep/processor/field_manager/rule.py
```diff
@@ -102,8 +102,21 @@ class Config(Rule.Config):
...
         )
         """The fields from where to get the values which should be processed"""
-        target_field: str = field(validator=validators.instance_of(str))
-        """The field where to write the processed values to"""
+        target_field: str = field(validator=validators.instance_of(str), default="")
+        """The field where to write the processed values to. Can be used to move/copy single values,
+        merge multiple values to one list or extend a list. To move multiple fields in one rule to 
+        multiple output fields see :code:`target_fields`. The target config settings are are
+        mutually exclusive."""
+        target_fields: list = field(
```
ekneg54, line 110: you have to disable this field in subclasses where it is not used by set the right field attributes (init=false, repr=false, eq=false) to prevent this field is used in underlying rules. (if used, a typeerror should be raised by attrs)

ekneg54: how about give this processor a field called `mapping` with a mapping from source_field to target_field. the grokker has this mapping field and the dissector has it too. it wouldn`t be completly new and is closer to the usecase then hoping the order in source_fields is the order in target_fields.

but as mentioned above. you have to deactivate this `mapping` field in subclasses where it is not used.

ekneg54 (changes requested):

logprep/processor/field_manager/rule.py
dtrai2: I did decide to go for the `mapping` field and removed the `target_fields` again. That way the distinction toe `target_field` is more clear also. I also did disabled the `mapping` field in all "child"-rules that do not need this new field.

sashs (approved):

ekneg54 (approved):

PR #389, opened 2023-05-11
Allow same target names in grokker mapping list

dtrai2:
Previously it was not possible to use a list of grok pattern that try to write into the same output fields. This was because the list of grok pattern was merged to one large regex-pattern with the `|` operator. This pull request changes this behavior by considering each list element as a separate regex pattern. That way it is possible to write a grokker rule with multiple grok-pattern that all try to write to the same output field.

### DIFF
--- a/logprep/util/grok/grok.py
+++ b/logprep/util/grok/grok.py
@@ -96,7 +96,11 @@ class Grok:
             matches = list(map(self._map_types, matches))
         # deduplicate matches
         matches = [dict(tup) for tup in {tuple(match.items()) for match in matches}]
-        return {self.field_mapper[field_hash]: value for match in matches for field_hash, value in match.items() }
+        return {
+            self.field_mapper[field_hash]: value
+            for match in matches
+            for field_hash, value in match.items()
+        }
 
     def _map_types(self, matches):
         for key, match in matches.items():
--- a/logprep/util/grok/grok.py
+++ b/logprep/util/grok/grok.py
@@ -86,12 +86,12 @@ class Grok:
         """
 
         if self.fullmatch:
-            match_obj = list(map(lambda x: x.fullmatch(text), self.regex_obj))
+            match_obj = [regex_pattern.fullmatch(text) for regex_pattern in self.regex_obj]
         else:
-            match_obj = list(map(lambda x: x.search(text), self.regex_obj))
+            match_obj = [regex_pattern.search(text) for regex_pattern in self.regex_obj]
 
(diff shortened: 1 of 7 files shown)


logprep/util/grok/grok.py
```diff
@@ -79,26 +79,35 @@ def __attrs_post_init__(self):
...
-        corresponding values.If not matched, return None.
+        corresponding values. If not matched, return None.
         custom patterns can be passed in by
         custom_patterns(pattern name, pattern regular expression pair)
         or custom_patterns_dir.
         """
 
         if self.fullmatch:
-            match_obj = self.regex_obj.fullmatch(text)
+            match_obj = list(map(lambda x: x.fullmatch(text), self.regex_obj))
```
ekneg54, line 89: I think a list comprehension should be faster here because of the additional converting to a list.

And it will be more readable in my opinion

logprep/util/grok/grok.py
```diff
@@ -79,26 +79,35 @@ def __attrs_post_init__(self):
...
         else:
-            match_obj = self.regex_obj.search(text)
+            match_obj = list(map(lambda x: x.search(text), self.regex_obj))
```
ekneg54, line 91: As above

logprep/util/grok/grok.py
```diff
@@ -79,26 +79,35 @@ def __attrs_post_init__(self):
...
 
-        if match_obj is None:
-            return None
-        matches = match_obj.groupdict()
+        match_obj = list(filter(lambda x: x is not None, match_obj))
```
ekneg54, line 93: As above

logprep/util/grok/grok.py
```diff
@@ -79,26 +79,35 @@ def __attrs_post_init__(self):
...
+        matches = list(map(lambda x: x.groupdict(), match_obj))
```
ekneg54, line 94: As above

ekneg54 (commented):
No complete review, but some comments

logprep/util/grok/grok.py
dtrai2: I revised it

logprep/util/grok/grok.py
dtrai2: I revised it

logprep/util/grok/grok.py
dtrai2: I revised it

logprep/util/grok/grok.py
dtrai2: I revised it

sashs (approved):

PR #402, opened 2023-06-12
rework logprep temp files

### LINKED ISSUE #401: stores temporary files not in /tmp
currently logprep downloads specific lists and external needed databases to ensure they are present when processing starts.
this leads to OSErrors in containerized environment if the rootfs in the container is mounted RO (which is best practice).

solution:

logprep should store these files in the systems configured temporary directory.

affected processors:

* domain_label_extractor
* domain_resolver
* geoip_enricher

additionally the filenames should not rotate after pipelines restart. this leads to downloading the file on every restart instead of using the already downloaded file.

dtrai2:
closes #401

### DIFF
--- a/logprep/processor/domain_label_extractor/processor.py
+++ b/logprep/processor/domain_label_extractor/processor.py
@@ -34,8 +34,8 @@ Processor Configuration
 .. automodule:: logprep.processor.domain_label_extractor.rule
 """
 import ipaddress
+import tempfile
 from functools import cached_property
-from multiprocessing import current_process
 from pathlib import Path
 from typing import Optional
 
@@ -87,7 +87,8 @@ class DomainLabelExtractor(Processor):
             downloaded_tld_lists_paths = []
             self._logger.debug("start tldlists download...")
             for index, tld_list in enumerate(self._config.tld_lists):
-                list_path = Path(f"{current_process().name}-{self.name}-tldlist-{index}.dat")
+                temp_dir = Path(tempfile.gettempdir())
+                list_path = temp_dir / Path(f"{self.name}-tldlist-{index}.dat")
                 list_path.touch()
                 list_path.write_bytes(GetterFactory.from_string(tld_list).get_raw())
                 downloaded_tld_lists_paths.append(f"file://{str(list_path.absolute())}")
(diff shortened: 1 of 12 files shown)


logprep/processor/domain_label_extractor/processor.py
```diff
@@ -87,7 +87,8 @@ def setup(self):
             downloaded_tld_lists_paths = []
             self._logger.debug("start tldlists download...")
             for index, tld_list in enumerate(self._config.tld_lists):
-                list_path = Path(f"{current_process().name}-{self.name}-tldlist-{index}.dat")
+                temp_dir = Path(tempfile.gettempdir())
+                list_path = temp_dir / Path(f"{self.name}-tldlist-{index}.dat")
                 list_path.touch()
                 list_path.write_bytes(GetterFactory.from_string(tld_list).get_raw())
```
ekneg54, line 93: I think we could get issues with multiprocessing on this solution. Not for touching the file and not for reading it, but in case two processed try to write to this file as they do in line 92.

ekneg54: same in the other touched files.

How about creating the file as before with the process name but delete it on restart?

ekneg54 (changes requested):
I think we will have issues with multiprocessing... :-/

logprep/processor/domain_label_extractor/processor.py
dtrai2: mh yeah I can try to delete the file or the whole temp directory on shutdown of logprep, but that doesn't help with your suggestion from the issue: "this leads to downloading the file on every restart **instead of using the already downloaded file.**"
If the issues can only happen on writing to the file, can't we solve this then with a `lock`? I think the processors only create this file in the startup process and are not writing to it during the logprep run. So creating these files initially with a lock should be possible if multiple processes can read from it at the same time. Besides, I think even the reading shouldn't be an issue as the `TLDExtract` library should load the lists content into memory.

ekneg54: yes.. a lock should work for this. 
And we also should give it a try for the geoip_enricher database. we then should lock every database access.. .

dtrai2: I added a `FileLock` to these write operations.

PR #411, opened 2023-06-20
Overhaul AMIDES output

clumsy9:
- Confidence value of the misuse detection step is stored into the `amides.confidence`-field of each event processed by AMIDES, even if the event was not classified as malicious. This has great influence in the process of reducing False Positives (FPs).
- In case of  a positive misuse detection result, rule attributions  are inserted into the `amides.attributions`-field. Attributions are now formatted as array of several (sub-)documents, holding `rule` and `confidence` for each attributed rule.


logprep/processor/amides/detection.py
```diff
@@ -88,14 +88,14 @@ def detect(self, sample: str) -> int:
...
+        predict: Tuple[boolean, float].
+            Detection result as boolean value and the corresponding confidence.
         """
         confidence_value = super().detect(sample)
 
         if confidence_value >= self._decision_threshold:
-            return 1
-        return 0
+            return True, round(confidence_value, 3)
+        return False, round(confidence_value, 3)
```
dtrai2, line 98: Could you also please adjust the method return type in line 80?

logprep/processor/amides/detection.py
```diff
@@ -128,7 +128,7 @@ def _setup(self, models: dict):
             except DetectionModelError as err:
                 raise RuleAttributorError from err
 
-    def attribute(self, sample: str) -> Dict[str, float]:
+    def attribute(self, sample: str) -> list[str, float]:
```
dtrai2, line 131:
```suggestion
    def attribute(self, sample: str) -> list[dict]:
```

logprep/processor/amides/detection.py
```diff
@@ -145,18 +145,18 @@ def attribute(self, sample: str) -> Dict[str, float]:
...
 
         for name, rule_model in self._rule_models.items():
             rule_confidence_value = rule_model.detect(cmdline)
-            conf_values.append((name, round(rule_confidence_value, 3)))
+            conf_values.append({"rule": name, "confidence": round(rule_confidence_value, 3)})
 
         return conf_values
 
     def _get_rules_with_highest_confidence_values(
         self, conf_values: List[Tuple[str, float]]
```
dtrai2, line 158:
```suggestion
        self, conf_values: List[dict]
```

logprep/processor/amides/processor.py
```diff
@@ -218,15 +218,16 @@ def _apply_rules(self, event: dict, rule: AmidesRule):
...
+        result = {}
 
-        return self._calculate_rule_attributions(cmdline)
+        malicious, result["confidence"] = self._perform_misuse_detection(cmdline)
+        if malicious:
+            result["attributions"] = self._calculate_rule_attributions(cmdline)
+
+        return result
 
     def _perform_misuse_detection(self, cmdline: str) -> int:
```
dtrai2, line 232:
```suggestion
    def _perform_misuse_detection(self, cmdline: str) -> Tuple[bool, float]:
```

The typing of the return of the method `_calculate_rule_attributions` below (line 241) is also not correct. It should be `List[dict]`.

tests/unit/processor/amides/test_amides.py
```diff
@@ -38,9 +38,10 @@ def test_process_event_malicious_process_command_line(self):
         self.object.process(document)
         assert self.object.metrics.number_of_processed_events == 1
 
-        rule_attributions = document.get("rule_attributions", None)
-        assert rule_attributions
-        assert len(rule_attributions) == 10
+        result = document.get("amides")
+        assert result
+        assert result["confidence"] >= 0.32
```
dtrai2, line 43: Just to understand it: Where does the `0.32` come from? Can't we fix it to the actual value `0.757`, maybe when setting the random seed? Or is it just to test a lower bounds and the actual confidence is not important.

dtrai2 (changes requested):

dtrai2:
I hope it's okay that I reviewed it already. ;) One more thing I just noticed: could you please also add a small `CHANGELOG.md` entry.

logprep/processor/amides/detection.py
clumsy9: Of course :)

logprep/processor/amides/detection.py
clumsy9: Changed the type hint as requested.

clumsy9:
Thx for the review! Code was not completely ready for being merged, but should be ready now :) I also added the changes into the `CHANGELOG.md`.

tests/unit/processor/amides/test_amides.py
dtrai2: ah okay, got it. Thanks. Maybe you can replace it then with the variable `self.CONFIG.get("decision_threshold")`, that way we have one magic number less and it's a bit easier to understand.

CHANGELOG.md
```diff
@@ -6,7 +6,9 @@
 
 * Replace rule_filter with lucene_filter in predetector output. The old internal logprep rule 
 representation is not present anymore in the predetector output, the name `rule_filter` will stay
-in place of the `lucene_filter` name. 
+in place of the `lucene_filter` name.
+* 'amides' now stores confidence values of processed events in the `amides.confidence` field.
+In case of positive detection results, rule attributions are now inserted in the `amides.attributions` field.
```
dtrai2, line 11:
```suggestion
* `amides` processor now stores confidence values of processed events in the `amides.confidence`
field. In case of positive detection results, rule attributions are now inserted in the
`amides.attributions` field.
```

dtrai2 (changes requested):

dtrai2:
Thanks, one last thing and then we can merge it :+1:

dtrai2 (approved):

PR #416, opened 2023-06-30
Refactor rule tree

ppcad:


logprep/framework/rule_tree/node.py
```diff
@@ -9,7 +9,13 @@
 class Node:
     """Tree node for rule tree model."""
 
-    def __init__(self, expression: FilterExpression):
+    __slots__ = ("_expression", "_children", "matching_rules")
+
+    _expression: Union[FilterExpression, str]
+    _children: list
```
ekneg54, line 15: if we can achieve to make the FilterExpression class hashable, we could use a dict here with the filterexpression as key. this ensures uniqueness and accessing or lookup an element in a dict has O(1) complexity. This could be an improvement for get_matching_rules in future refactorings

logprep/filter/expression/filter_expression.py
```diff
@@ -120,20 +102,20 @@ class Not(FilterExpression):
     """Filter expression that negates a match."""
 
     def __init__(self, expression: FilterExpression):
-        self.expression = expression
+        self.child = expression
 
     def __repr__(self) -> str:
-        return f"NOT ({str(self.expression)})"
+        return f"NOT ({str(self.child)})"
```
ekneg54, line 108: it is somehow suspicious to call the `str` method for generation of the `__repr__` dunder method.
```suggestion
        return f"NOT ({repr(self.child)})"
```

logprep/filter/expression/filter_expression.py
ekneg54, line 1: Please take a look at the FilterExpression class and its subclasses. Please consider to sharpen the interface a little bit more. In the child classes I the attributes like sometimes `children` and sometimes `child`. We should declare these attributes somewhere to make sure the reader knows what to expect from them. For me it is not clear if they are creates only in the __init__ method .

My stomach says that it could be a good idea to use only the plural and make it `children: list[FilterExpression]`

logprep/framework/rule_tree/rule_segmenter.py
ekneg54, line 1: I have no clue what you are doing here. I guess you are writing your own parser to get from a string to a list of expressions. 

It is dangerous. If we merge this you are the only guy who is able to change something on this side of logprep. 
I would suggest to take a look on existing parsing libraries. We have pyparsing in our dependencies. Do you think it would be possible to declare a Grammar and use pyparsing for this step?

ekneg54: Another approach could be a workshop where you explain this part. We could use a webex for that.

ekneg54 (changes requested):
Thank you for your work. This was a lot. But I can`t merge it, because I do not understand main parts of it. please read through the comments and write me a mail. We need to come together for some pair programming sessions to get this important part of logprep right and to ensure, that we not have to touch it again in the next steps.

tests/unit/framework/rule_tree/test_demorgan_resolver.py
```diff
@@ -0,0 +1,94 @@
...
+    Or,
+    Not,
+    CompoundFilterExpression,
+)
+from logprep.framework.rule_tree.demorgan_resolver import (
+    DeMorganResolver,
+    DeMorganResolverException,
+)
+
+from tests.unit.framework.rule_tree.shared_constants import sfe_1, sfe_2, sfe_3, sfe_4
```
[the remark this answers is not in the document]
ppcad: I have changed the names and placed the variables in each module where they are used.

tests/unit/framework/rule_tree/test_demorgan_resolver.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@pytest.fixture(name="demorgan_resolver")
+def fixture_demorgan_resolver():
+    return DeMorganResolver()
+
+
+class TestDeMorganResolver:
+    @pytest.mark.parametrize(
+        "expression, resolve_state",
```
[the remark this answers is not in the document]
ppcad: I've changed it.

tests/unit/framework/rule_tree/test_demorgan_resolver.py
```diff
@@ -0,0 +1,94 @@
...
+            (Or(And(sfe_1, sfe_2)), False),
+            (Not(And(sfe_1, sfe_2)), True),
+            (Or(Not(And(sfe_1, sfe_2)), sfe_3), True),
+        ],
+    )
+    def test_has_unresolved_not_expression(self, expression, resolve_state, demorgan_resolver):
+        assert demorgan_resolver._has_unresolved_expression(expression) == resolve_state
+
+    @pytest.mark.parametrize(
+        "expression, resolved_expr",
```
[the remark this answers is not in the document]
ppcad: I've changed it.

tests/unit/framework/rule_tree/test_demorgan_resolver.py
```diff
@@ -0,0 +1,94 @@
...
+            (Not(Or(And(sfe_1, sfe_2), sfe_3)), And(Or(Not(sfe_1), Not(sfe_2)), Not(sfe_3))),
+            (Not(And(Or(sfe_1, sfe_2), sfe_3)), Or(And(Not(sfe_1), Not(sfe_2)), Not(sfe_3))),
+            (And(Not(And(sfe_1, sfe_2)), sfe_3), And(Or(Not(sfe_1), Not(sfe_2)), sfe_3)),
+            (
+                And(Not(Or(sfe_1, sfe_2)), Not(And(sfe_3, sfe_4))),
+                And(And(Not(sfe_1), Not(sfe_2)), Or(Not(sfe_3), Not(sfe_4))),
+            ),
+        ],
+    )
+    def test_resolve(self, expression, resolved_expr, demorgan_resolver):
```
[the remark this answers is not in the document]
ppcad: I've changed it.

tests/unit/framework/rule_tree/test_demorgan_resolver.py
```diff
@@ -0,0 +1,94 @@
...
+        assert demorgan_resolver.resolve(expression) == resolved_expr
+
+    @pytest.mark.parametrize(
+        "expression, resolved_expr, error",
```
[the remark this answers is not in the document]
ppcad: I've changed it.

tests/unit/framework/rule_tree/test_demorgan_resolver.py
```diff
@@ -0,0 +1,94 @@
...
+                Not(sfe_1),
+                (
+                    DeMorganResolverException,
+                    r'Could not resolve expression ".*", '
+                    + r'since its child is neither of the type "AND" nor "OR".',
+                ),
+            ),
+        ],
+    )
+    def test_resolve_not_expression(self, expression, resolved_expr, error, demorgan_resolver):
```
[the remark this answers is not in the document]
ppcad: I've changed it.

logprep/framework/rule_tree/demorgan_resolver.py
```diff
@@ -0,0 +1,110 @@
...
+
+        if isinstance(expression, Not):
+            return self._resolve_not_expression(expression)
+        if isinstance(expression, CompoundFilterExpression):
+            return self._resolve_compound_expression(expression)
+
+        raise DeMorganResolverException(f"Could not resolve expression {expression}")
+
+    @staticmethod
+    def _has_unresolved_expression(expression: FilterExpression) -> bool:
```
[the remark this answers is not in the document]
ppcad: You are right. I removed it and it seems to work just fine without it. It is also more readable and about 10% faster in my tests.

logprep/filter/expression/filter_expression.py
ppcad: I've changed it.

logprep/filter/expression/filter_expression.py
ppcad: The NOT expression can never have more than one child. Making it children would imply that there could be more than one.
Children is used in compound filter expressions like AND and OR.
The other filter expressions, like the string filter expression, can never have children and the reader should not expect that they have.

So I'm not sure if this should be changed.

logprep/framework/rule_tree/rule_segmenter.py
ppcad: The segmenter doesn't get a string, but a (compound) FilterExpression as input.
This input expression must have been already resolved via demorgan.
The expression is then changed into the disjunctive normal form (DNF).
However, the output is not another FilterExpression, but a list of lists with rule segments (FilterExpressions that are not compound), which represent the FilterExpression in DNF.
The outer list representing OR and the inner lists representing AND.

This representation then allows us to easily sort the rule segments and build a tree out of them.

Example:
Let's say we have the following CompoundFilterExpression X  with the StringFilterExpressions A, B and C:

`X = (A and (B or C)) or D`
Call `segment_into_dnf(X)` => `[[A, B], [A, C], [D]]`
Which is equivalent to the FilterExpression `(A and B) or (A and C) or (D)` (the DNF of X).

I guess in theory we could rewrite it so that it simply changes the FilterExpression into another FilterExpression in DNF, but this is not really necessary and it would require a rewrite of the following steps and make it more complicated.

logprep/framework/rule_tree/node.py
```diff
@@ -9,7 +9,13 @@
 class Node:
```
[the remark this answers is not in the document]
ppcad: We could do that, but it would require every FilterExpression to have a children attribute.
FilterExpression like the StringFilterExpression can never have children.

logprep/framework/rule_tree/node.py
```diff
@@ -9,7 +9,13 @@
...
     """Tree node for rule tree model."""
 
-    def __init__(self, expression: FilterExpression):
+    __slots__ = ("_expression", "_children", "matching_rules")
+
+    _expression: Union[FilterExpression, str]
```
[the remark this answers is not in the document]
ppcad: I've changed it.

logprep/filter/expression/filter_expression.py
ekneg54: but if we take the plural and asume that every filterexpression could have 1 or more childs, then we are very clear in our interface, because then every filterexpression has an attribute where you can find a list with filterexpressions in it. And they can all be handled the same. And there is only one point where it is declared. and there we could give a realy good explaination about this attribute. 

the other solution could be, to declare the attribute where it is created in the init. and to add documentation for it in every subclass where it is declared.

logprep/framework/rule_tree/node.py
ppcad: Do I undertand it correctly that you would like to use a dict for the "matching_rules" attribute later on, but for now just make the FIlterExpressions hashable?

logprep/filter/expression/filter_expression.py
ppcad: If we choose the 1. solution, then it might be possible to get rid of the "Node" class.
I will try to do that.

ppcad: I added children to FilterExpression. However, I decided to not remove Node for now, since children in Node have a different meaning than in FilterExpression.

logprep/framework/rule_tree/node.py
ekneg54: we could do both in future itereations

ekneg54 (approved):
thank you for the changes.
plz merge if ready

ppcad:
I had to remove `__slots__` from `children` in `FilterExpression`, since this caused some rules to not match. I added a unit test for such a rule.
I don't know why `__slots__` causes this.
The changes are in commit [8190857f55b39c50872d921f3e5ea8821b266947](https://github.com/fkie-cad/Logprep/pull/416/commits/8190857f55b39c50872d921f3e5ea8821b266947).

ekneg54:
I'm fine with it 👍

PR #417, opened 2023-06-30
fix dissect pattern that end with a dissect

### LINKED ISSUE #410: Dissector - cannot extract last section of a field
When using the dissector, I am not able to extract the last section of a field.

In this example, I want to extract only the name of the SAP system, i.e. extract everything after "sap":

```
Field: sapicm
Pattern: sap%{sap_system}

Error: Logprep dissector ValueError mapping must match regex + + + + + / / d int float str bool + + + + + + / / d int float str bool sap system doesn t"
```

dtrai2:
closes #410


tests/unit/processor/dissector/test_dissector_rule.py
```diff
@@ -74,8 +74,8 @@ def test_create_from_dict_returns_dissector_rule(self):
                     "filter": "message",
                     "dissector": {"mapping": {"field": "%{ts}"}},
```
ekneg54, line 75: If this is a valid rule now, shouldn't we test if this rule would lead to the corresponding expected result?

ekneg54 (changes requested):
Please add a TestCase for the new valid dissect pattern

tests/unit/processor/dissector/test_dissector_rule.py
dtrai2: I added a test for that :+1:

ekneg54 (approved):

PR #428, opened 2023-07-19
fix crash on invalid json input in kafka input connector

### LINKED ISSUE #424: Crash on invalid JSON input in Kafka input connector
Logprep crashes if the Kafka input connector receives data that can't be parsed as JSON.

**Expected behavior**
Logprep writes the message into the error index if the Kafka input connector receives data that can't be parsed as JSON.

**Steps to reproduce**
Stat Logprep with the Kafka input connector and insert a message that can't be parsed as JSON.

**Environment**

Logprep version: 1b9a1172771e5a52b0a64ee5f2f0dbed5e9baf9c
Python version: 3.11

dtrai2:
closes #424 
furthermore, `ProcessingCriticalError` aren't wrapped again inside `ProcessingCriticalError` and `events` are only cleared after they were written to the error output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 ### Improvements
 ### Bugfix
 
+* Fix restarting of logprep every time the kafka input connector receives events that aren't valid
+json documents. Now the documents will be written to the error output.
+
 ## v6.7.0
 ### Improvements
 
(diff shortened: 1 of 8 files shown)


ppcad (approved):
Thank you!

PR #431, opened 2023-07-21
revise processing counter

### LINKED ISSUE #429: Printed events processed interval does not correspond to real processing time
Logprep can print the number of processed events for a configured interval.
However, the printed interval is smaller than the actual processing time if processing one event takes longer than that interval.
This can happen when Logprep is starting if it has to load a lot of rules and the print interval is smaller than the time it takes to load them.

**Expected behavior**
The printed interval for processed events corresponds to the real processing time.

**Current behavior**
The printed interval for processed can be smaller the real processing time.

**Steps to reproduce**
Set a small "print_processed_period" and load a lot of rules.

**Environment**

Logprep version: latest
Python version: 3.11

**Possible solution**
This might be solved in multiple ways:

1. Print the real time it took instead of the configured interval time. This means the printed interval might increase if it took longer than configured.
2. Ensure the processed count gets always printed at the interval, even if events are still being processed.
3. Ensure everything is loaded before the counting starts.

dtrai2:
closes #429

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Fix restarting of logprep every time the kafka input connector receives events that aren't valid
 json documents. Now the documents will be written to the error output.
+* Fix ProcessCounter to actually print counts periodically and not only once events are processed
 
 ## v6.7.0
 ### Improvements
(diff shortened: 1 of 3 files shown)


ppcad (approved):
Works very well. Thank you!

PR #436, opened 2023-08-22
Add creation timestamp to pre-detections

ppcad:


logprep/processor/pre_detector/processor.py
```diff
@@ -128,6 +129,8 @@ def _apply_rules(self, event, rule):
         if "@timestamp" in event:
             for detection in self._extra_data:
                 detection["@timestamp"] = event["@timestamp"]
+        for detection in self._extra_data:
+            detection["creation_timestamp"] = TimeParser.now().isoformat()
```
dtrai2, line 133: How about we turn this around? That way we only have to loop over the detections once?
```suggestion
        for detection in self._extra_data:
            detection["creation_timestamp"] = TimeParser.now().isoformat()
            if "@timestamp" in event:
                detection["@timestamp"] = event["@timestamp"]
```

dtrai2 (changes requested):
I just checked the changes. I only have two suggestions, one regarding the for-loop and could you also please add a CHANGELOG.md entry? Thank you!

dtrai2:
I am not really available for the next two weeks. @ekneg54 could you set the final approve after my changes are implemented?

logprep/processor/pre_detector/processor.py
ppcad: Thank you for reviewing it. I have made the changes.

ekneg54 (approved):

ekneg54 (approved):

PR #439, opened 2023-09-07
remove logaggregator

ekneg54:
In our work we noticed a big memory leak during debug mode. this should resolve it. 

@ppcad please confirm, that you do not need the log aggregation, because we do not need it either.


ppcad:
Thank you for finding this! I need to check if aggregation is still needed.

ppcad:
Log aggregation is still needed, so please don't remove it.
Fixing the leak will be the preferred solution if that is possible.

ekneg54:
The other way to fix it is to only log generic lines to avoid that the dict is getting bigger and bigger. 
but what kind of "logger" would this be? 😄

logprep/run_logprep.py
```diff
@@ -25,6 +24,31 @@
...
+name_to_level = {
+    "CRITICAL": CRITICAL,
+    "FATAL": FATAL,
+    "ERROR": ERROR,
+    "WARN": WARNING,
+    "WARNING": WARNING,
+    "INFO": INFO,
+    "DEBUG": DEBUG,
+    "NOTSET": NOTSET,
+}
```
dtrai2, line 50: you don't need this, as you can throw in the logel-level string from `log_config.get("level", ...)` directly into the `basicConfig(level=log_level, ...)` call.
```suggestion
```

logprep/run_logprep.py
```diff
@@ -137,10 +161,16 @@ def get_versions_string(args) -> str:
...
 
 
-def _setup_logger(args, config):
+def _setup_logger(args, config: Configuration):
     try:
-        AggregatingLogger.setup(config, logger_disabled=args.disable_logging)
-        logger = AggregatingLogger.create("Logprep")
+        log_config = config.get("logger", {})
+        log_level_name = log_config.get("level", "INFO")
+        log_level = name_to_level.get(log_level_name)
```
dtrai2, line 168: as above
```suggestion
```

logprep/run_logprep.py
```diff
@@ -25,6 +24,31 @@
...
 
+from logging import (
+    getLogger,
+    CRITICAL,
+    FATAL,
+    ERROR,
+    WARNING,
+    INFO,
+    DEBUG,
+    NOTSET,
```
dtrai2, line 35: see below
```suggestion
```

logprep/run_logprep.py
```diff
@@ -137,10 +161,16 @@ def get_versions_string(args) -> str:
     return version_string
 
 
-def _setup_logger(args, config):
+def _setup_logger(args, config: Configuration):
     try:
-        AggregatingLogger.setup(config, logger_disabled=args.disable_logging)
-        logger = AggregatingLogger.create("Logprep")
+        log_config = config.get("logger", {})
+        log_level_name = log_config.get("level", "INFO")
```
dtrai2, line 167:
```suggestion
        log_level = log_config.get("level", "INFO")
```

logprep/run_logprep.py
```diff
@@ -137,10 +161,16 @@ def get_versions_string(args) -> str:
...
+        log_level = name_to_level.get(log_level_name)
+        basicConfig(
+            level=log_level, format="%(asctime)-15s %(name)-5s %(levelname)-8s: %(message)s"
+        )
+        logger = logging.getLogger("Logprep")
+        logger.info(f"Log level set to '{log_level_name}'")
```
dtrai2, line 173:
```suggestion
        logger.info(f"Log level set to '{log_level}'")
```

dtrai2 (changes requested):

dtrai2:
ah, can you also please add a CHANGELOG.md entry.

dtrai2 (approved):

PR #440, opened 2023-09-07
time measurement checks for caller class name now

### LINKED ISSUE #434: Deleter doesn't delete logs
When using the deleter, nothing is deleted and an empty file is dragged through the whole pipeline resulting in only containing the processing times in the meta-data.

**Expected behavior**

Immediately deletes the log when filter of deleter matches in preprocessor step.

**Current behavior**

Creates an empty dictionary thats entirely processed and at the end, cause of a non existing index gets written into default instead.

**Steps to reproduce**

1. Running Logprep
2. Enabling the preprocessor **deleter**
3. Look at default index

**Environment**

Logprep version: 23.07.03
Python version: 3.11
Enviroment: Kubernetes with containerd

dtrai2:
the test also checks now for an actual deleter implementation

closes #434

### DIFF
--- a/logprep/util/time_measurement.py
+++ b/logprep/util/time_measurement.py
@@ -45,7 +45,7 @@ class TimeMeasurement:
                 return func(*args, **kwargs)
 
             def add_processing_times_to_event(event, processing_time, caller, name):  # nosemgrep
-                if caller.__class__.__name__ == "Deleter" and not event:
+                if not event:
                     return
                 if not event.get("processing_times"):
                     event["processing_times"] = {}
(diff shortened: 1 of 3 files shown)


ppcad:
Thank you for changing this, but the other problem was, that the pipeline adds time measurements, even after it has been removed by the deleter processor. It is called after all events have been processed.

Maybe check for `name = "pipeline"` in `time_measurement.py` and only add the times if the event is not empty.

ekneg54:
How about to remove the feature to write metrics to events completely in favor of the Prometheus exporter?

ppcad:
I still need it at the moment.

dtrai2:
okay, well then I remove checking who tries to write the times into the event and just assume that if the event is empty no measurement time is needed.

ppcad (approved):
Thank you, it works!

PR #444, opened 2023-09-21
rewrite kafka connector

ekneg54:
this PR rewrites the confluent kafka input and output processors with focus on the input side.
please take a look at the CHANGELOG.md for changes.


ekneg54:
rebased from main.

@ppcad now it`s ready to rumble 👍

ekneg54:
it is possible, that the `statistics.interval.ms` default was too low. It was set to 1s. I increased the default to 30 seconds.

ppcad:
> I added a validation for mandatory keys in kafka_config object. This prevents the Exceptions during start and logs the missing keys.

Thanks for the changes. This solves the problem.

ppcad:
> it is possible, that the `statistics.interval.ms` default was too low. It was set to 1s. I increased the default to 30 seconds.

This does not seem to be the reason for the performance loss.

It seems that the new version is faster than the old one with 1 process, but slower once I increase the amount up to the partition count.

ppcad (approved):

PR #445, opened 2023-09-28
Fix CVE 2023 37920

ekneg54:
this fixes 2 CVEs by upgrading the dependencies.
as seen in changes `hyperscan`, `confluent_kafka` and `opensearch-py` are upgraded too. 

please test this change.

* had to fix an issue with tldextract in cause of a new feature: https://github.com/john-kurkowski/tldextract/releases
* had to fix an issue with semgrep in ci pipeline in cause of new semgrep version


ppcad (approved):
Thank you

PR #447, opened 2023-10-02
add support python 3.12 and drop support python 3.9

ekneg54:
waits for:

- [x] https://github.com/confluentinc/confluent-kafka-python/issues/1629
- [x] https://github.com/aio-libs/aiohttp/issues/7646
- [x] https://github.com/aio-libs/aiohttp/issues/7675
- [x] https://github.com/darvid/python-hyperscan/issues/98

This PR adds logprep support for python 3.12 and with this drops support for python 3.9.
Along with this I had to remove semgrep from the project as it seems with the new version they only support the cloud based version of their tool. And as I know, we have no cloud account for the semgrep cloud.
We should search for another SAST tool to check our code in the future.


versioneer.py
```diff
@@ -1,3 +1,4 @@
+# pylint:disable-all
```
dtrai2, line 1: :smile:

dtrai2 (changes requested):
just found a typo, but besides that it loos good to me!

dtrai2 (approved):

ppcad (approved):

PR #448, opened 2023-10-03
add env variable enrichment preprocessor

ekneg54:
this PR adds a preprocessor to enrich incomming events by the systems running logprep env variables. This is useful to enrich the events with kubernetes pod names to show which pod handled the event.


ppcad (approved):
Thank you

PR #453, opened 2023-10-07
add inline rule config feature to pipline config

ekneg54:


logprep/abc/processor.py
```diff
@@ -41,22 +41,24 @@ class Config(Component.Config):
...
             validator=[
                 validators.instance_of(list),
-                validators.deep_iterable(member_validator=validators.instance_of(str)),
+                validators.deep_iterable(member_validator=validators.instance_of((str, dict))),
             ]
         )
         """List of rule locations to load rules from.
         In addition to paths to file directories it is possible to retrieve rules from a URI.
         For valid URI formats see :ref:`getters`.
+        As last option it is possible to define rules inline.
```
dtrai2, line 50:
```suggestion
        As last option it is possible to define entire rules with all their configuration parameters as list elements.
```

logprep/abc/processor.py
```diff
@@ -41,22 +41,24 @@ class Config(Component.Config):
...
```
dtrai2, line 61:
```suggestion
        As last option it is possible to define entire rules with all their configuration parameters as list elements.
```

logprep/abc/processor.py
```diff
@@ -230,8 +232,23 @@ def test_rules(self) -> dict:
...
+        Returns
+        -------
+        list
+            a list of files and rule definitions
+        """
         resolved_paths = []
         for rule_path in rule_paths:
+            if isinstance(rule_path, dict):
+                resolved_paths.append(rule_path)
+                continue
```
dtrai2, line 251: I have an issue here with the variable naming. The variable `rule_path` does not only contain `paths` anymore, as the list can contain entire rule dictionaries as well now. So it's a mixed data structure. I would have to ideas: (a) we just rename it to something more generic like maybe `rule_descriptions` (as a `file_path`, `uri` or `yaml` could all be seen as a description of a rule) or (b) we actually extend this method, such that the destination for all `file_paths` and `uris` is resolved to the actual rule dictionaries. That way we would only have one list with rule dictionaries. The latter choice would make this method here more complicated, but would simplify the `logprep/processor/base/rule.py::create_rules_from_target` method.

logprep/processor/base/rule.py
```diff
@@ -255,21 +255,23 @@ def lucene_filter(self):
     # pylint: enable=C0111
 
     @classmethod
-    def create_rules_from_target(cls, path: str) -> list:
+    def create_rules_from_target(cls, rule_target: str) -> list:
```
dtrai2, line 258: Depending on how you decide to implement my previous remark you could either rename this `rule_target` to something similar to `rule_description` or you just have here a list of rule dicts anways such that this whole method could be simplified.

logprep/util/schema_and_rule_checker.py
```diff
@@ -138,14 +139,17 @@ def _validate_rules_in_path(
...
         number_of_checked_rules = 0
-        for root, _, files in walk(path_rules):
-            for file in files:
-                number_of_checked_rules += 1
-                rule_path = join(root, file)
-
-                multi_rule = self.check_rule_creation_errors(rule_class, rule_path)
-                self._validate_schema(multi_rule, path_schema, rule_path)
-            self._print_schema_check_results(path_schema)
+        if isinstance(path_rules, dict):
```
dtrai2, line 142: Same problem here with the `path_rules`, though this might be indeed actually best solved by just renaming the variable.

logprep/util/schema_and_rule_checker.py
```diff
@@ -198,7 +202,9 @@ def check_rule_creation_errors(self, rule_class: Rule, rule_path: str) -> Option
         """
         rule = None
         try:
-            if rule_path.endswith(".json") or rule_path.endswith(".yml"):
+            if isinstance(rule_path, dict):
```
dtrai2, line 205: Same here.

tests/acceptance/test_full_configuration.py
```diff
@@ -225,3 +225,57 @@ def test_logprep_exposes_prometheus_metrics(tmp_path):
...
+    default_index: processed
+    error_index: errors
+    message_backlog_size: 10000
+    timeout: 10000
+    flush_timeout: 10
+    max_retries: 3
+    user: admin
+    secret: admin
+
+    """
```
dtrai2, line 281: I think the actual test is missing here :wink:

dtrai2 (changes requested):
I have found an inconsistency regarding a variable name and a probably missing test.

logprep/processor/base/rule.py
ekneg54: we will challenge that later

tests/acceptance/test_full_configuration.py
ekneg54: yes you are right. I tested the factory and the functionality is tested in the acceptance tests by changing the quickstart `pipeline.yml`

dtrai2 (approved):

PR #455, opened 2023-10-07
remove MultiprocessLogHandler

ekneg54:


logprep/framework/pipeline_manager.py
```diff
@@ -101,18 +102,11 @@ def restart_failed_pipeline(self):
...
-            logger.handle(self._log_handler.get(timeout))
-            while True:
-                logger.handle(self._log_handler.get(0.0))
-        except Empty:
-            pass
-
     def stop(self):
         """Stop processing any pipelines by reducing the pipeline count to zero."""
         self._decrease_to_count(0)
+        while len(self._pipelines) > 0:
```
ppcad, line 108: Will this ever be  >0? `_decrease_to_count(0)`  has basically the same condition in its while loop.
Also, if this wouldn't be always true, then it would rapidly spam the log message.

tests/acceptance/util.py
```diff
@@ -165,7 +165,6 @@ def get_patched_runner(config_path, logger):
         The patched logprep runner
     """
     runner = Runner(bypass_check_to_obtain_non_singleton_instance=True)
-    runner.set_logger(logger)
```
ppcad, line 168: I think logger can be removed from the arguments of the method, since it isn't used anymore.

ppcad (changes requested):

tests/acceptance/util.py
ekneg54: sure... removed it ... thank you

ekneg54:
done it ... plz have a look again. 

Thank you.

ppcad (approved):
Thank you, looks good to me!

PR #462, opened 2023-10-19
fix CVE-2023-45803

ekneg54:


ekneg54:
cve warning disappears in depandabot

PR #464, opened 2023-10-27
Improve rule loading times in rule tree

ppcad:
* Fix cnf to dnf conversion and improve its performance
* Optimize rule tree sorter performance
* Add rule tree changes to readme and changelog


ekneg54:
as we reimplement metrics now,  I suggest to wait with this to get vision about what is actual slow on loading rules and wich rules.

Am I right, that improve rule loading times is only done because restarting pipelines seems to be slow?
There are multiple things that could getting slow. Loading lists? loading rules? And so there are different ways to deal with them.  But before dealing with them we should get vision.

We implement some additional metrics for restarting pipelines. Having that, you should be able to identify what is actual slow in your configuration.

A timeout feels like a workaround and does not solves the actual problem with the risk that some rules will be missing in processing.

ppcad:
> as we reimplement metrics now, I suggest to wait with this to get vision about what is actual slow on loading rules and wich rules.
> 
> Am I right, that improve rule loading times is only done because restarting pipelines seems to be slow? There are multiple things that could getting slow. Loading lists? loading rules? And so there are different ways to deal with them. But before dealing with them we should get vision.
> 
> We implement some additional metrics for restarting pipelines. Having that, you should be able to identify what is actual slow in your configuration.
> 
> A timeout feels like a workaround and does not solves the actual problem with the risk that some rules will be missing in processing.

This change does include an optimization of rule sorting and dnf conversion. The algorithms did some redundant looping.
The dnf conversion can still explode in load time, even if the implementation would be perfectly optimized.
The conversion has simply such a high complexity.
So the timeout is needed to filter out those cases.

ppcad:
I removed the timeout.
Please have a look @ekneg54.

CHANGELOG.md
```diff
@@ -32,9 +32,12 @@
 * reimplemented prometheus metrics exporter to provide gauges, histograms and counter metrics
 * removed shared counter, because it is redundant to the metrics
 * get exception stack trace by setting environment variable `DEBUG`
+* improve loading times for the rule tree by optimizing the rule segmentation and sorting
```
ekneg54, line 35: version 8.0.0 is already released. please push your impovement up to the next section

CHANGELOG.md
```diff
@@ -32,9 +32,12 @@
...
 
 ### Bugfix
 
+* Fix the rule tree parsing some rules incorrectly, potentially resulting in more matches.
```
ekneg54, line 39: version 8.0.0 is already released. please push your impovement up to the next section

ekneg54 (changes requested):
minor change in Changelog please

ekneg54 (approved):

PR #467, opened 2023-11-01
refactor pre_detector

ekneg54:


tests/unit/processor/pre_detector/test_pre_detector.py
```diff
@@ -338,3 +338,12 @@ def _assert_equality_of_results(
         )
 
         assert sorted_detection_results == sorted_expected_detection_results
+
+    def test_adds_timestamp_to_extra_data_if_provided_by_event(self):
+        assert self.object.metrics.number_of_processed_events == 0
```
dtrai2, line 343: I think this line is there due to copy-pasting, but we probably don't need it ;) 
```suggestion
```

dtrai2 (changes requested):

dtrai2 (approved):

PR #471, opened 2023-11-15
Improve error message on empty rule filter

dtrai2:
When the rule filter is empty logprep logs the critical error: `... expected string or bytes-like object, got 'NoneType'` which can be confusing. The error `no filter defined` is more explicit. 

Removes also trailing whitespaces in CHANGELOG.md


logprep/processor/base/rule.py
```diff
@@ -320,7 +320,7 @@ def matches(self, document: dict) -> bool:
     @classmethod
     def _create_filter_expression(cls, rule: dict) -> FilterExpression:
         special_fields = cls._get_special_fields_for_rule_matching(rule)
-        if "filter" not in rule:
+        if "filter" not in rule or rule["filter"] is None:
```
ekneg54, line 323:
```suggestion
        if rule.get("filter") is None:
```
How about this? 

get returns None if filter is not in rule and if filter is in rule but None

ekneg54 (changes requested):
One minor change suggestion to simplify things

logprep/processor/base/rule.py
dtrai2: smart :D

ekneg54 (approved):

PR #476, opened 2023-11-21
replace re by re2

ekneg54:


ekneg54:
does not work, because re2 does not support backreferences as mentioned in their pypi site.

Issue #435, opened 2023-08-07
Predetector doesn't check for existing field when using expressions set in filter


Malutthias:
In the predetector the rules currently must have explicitly stated to check for existence of fields when previously this was already done - e.g. in the rule tree structure.

**Expected behavior**

Should immediatly for existence of field (filter_xy) when using filter like "AND NOT filter_xy: abc"

**Current behavior**

Only checks for field existence if its directly integrated like "And filter_xy: * AND NOT filter_xy: abc"

**Steps to reproduce**

1. Run logprep
2. Enable predector rule with e.g. "... AND NOT filter_xy: abc... "

**Environment**

Logprep version: 23.07.03
Python version: 3.11
Enviroment: Kubernetes with containerd

PR #480, opened 2023-11-24
always expect exists filter

ppcad:
See issue #435

### DIFF
--- a/logprep/framework/rule_tree/rule_parser.py
+++ b/logprep/framework/rule_tree/rule_parser.py
@@ -129,17 +129,17 @@ class RuleParser:
         """
         for parsed_rule in parsed_rules:
             temp_parsed_rule = parsed_rule.copy()
-            skipped_counter = 0
+            added_segments = 0
             for segment_idx, segment in enumerate(temp_parsed_rule):
                 if isinstance(segment, (Exists, Always)):
-                    skipped_counter += 1
                     continue
-
                 exists_filter = RuleParser._get_exists_filter(segment)
-                if exists_filter is None or exists_filter in parsed_rule:
-                    skipped_counter += 1
-                else:
-                    parsed_rule.insert(segment_idx * 2 - skipped_counter, exists_filter)
+                if exists_filter is None:
+                    continue
+                if exists_filter in parsed_rule:
+                    continue
+                parsed_rule.insert(segment_idx + added_segments, exists_filter)
+                added_segments += 1
 
     @staticmethod
     def _get_exists_filter(segment: FilterExpression) -> Optional[Exists]:
@@ -157,9 +157,7 @@ class RuleParser:
 
         """
         if isinstance(segment, Not):
-            not_child = segment.children[0]
-            if isinstance(not_child, KeyValueBasedFilterExpression):
-                return Exists(not_child.key)
-        elif isinstance(segment, KeyValueBasedFilterExpression):
+            segment = segment.children[0]
+        if isinstance(segment, KeyValueBasedFilterExpression):
             return Exists(segment.key)
         return None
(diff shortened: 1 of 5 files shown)


logprep/framework/rule_tree/rule_parser.py
```diff
@@ -126,15 +130,36 @@ def _add_exists_filter(parsed_rules: list):
         for parsed_rule in parsed_rules:
             temp_parsed_rule = parsed_rule.copy()
             skipped_counter = 0
-
-            for segment_index, segment in enumerate(temp_parsed_rule):
```
dtrai2, line 130: Could you maybe flatten the if-statements in this loop? That way it is a bit easier to understand. A suggestion cold look like this:

```python
  for segment_idx, segment in enumerate(temp_parsed_rule):
      if isinstance(segment, (Exists, Always)):
          skipped_counter += 1
          continue
      exists_filter = RuleParser._get_exists_filter(segment)
      if exists_filter is None:
          skipped_counter += 1
          continue
      if exists_filter in filter_expression:
          skipped_counter += 1
          continue
      filter_expression.insert(segment_idx * 2 - skipped_counter, exists_filter)
```

logprep/framework/rule_tree/rule_parser.py
```diff
@@ -126,15 +130,36 @@ def _add_exists_filter(parsed_rules: list):
...
                     continue
 
-                exists_filter = Exists(segment.key)
-                if exists_filter in parsed_rule:
+                exists_filter = RuleParser._get_exists_filter(segment)
+                if exists_filter is None or exists_filter in parsed_rule:
                     skipped_counter += 1
-                    continue
+                else:
+                    parsed_rule.insert(segment_idx * 2 - skipped_counter, exists_filter)
```
dtrai2, line 142: Can you explain where the 2 is coming from? I was trying to understand this and my first idea was that `segment_idx - 1` should work as well, but it doesn't. Maybe you could give this magic number a meaningful name?

logprep/framework/rule_tree/rule_parser.py
```diff
@@ -126,15 +130,36 @@ def _add_exists_filter(parsed_rules: list):
...
+            Returns an Exists FilterExpression from a KeyValueBased FilterExpression or None.
+
+        """
+        if isinstance(segment, Not):
+            not_child = segment.children[0]
+            if isinstance(not_child, KeyValueBasedFilterExpression):
+                return Exists(not_child.key)
+        elif isinstance(segment, KeyValueBasedFilterExpression):
+            return Exists(segment.key)
+        return None
```
dtrai2, line 165: How about we use this implementation? It's a bit easier to read and understand by preventing nested if-statements.
```suggestion
        if isinstance(segment, Not):
            segment = segment.children[0]
        if isinstance(segment, KeyValueBasedFilterExpression):
            return Exists(segment.key)
        return None
```

dtrai2 (changes requested):

logprep/framework/rule_tree/rule_parser.py
ppcad: I've changed it.

logprep/framework/rule_tree/rule_parser.py
ppcad: Thanks for the suggestion. I've changed it.

logprep/framework/rule_tree/rule_parser.py
ppcad: It was a way to account for the new elements added to the list.

`segment_idx * 2 - skipped_counter` 
<=> `segment_idx - (segment_idx  - skipped_counter)`
<=> `segment_idx - added_exists_filter_count`

It was indeed confusing, so I changed it.

dtrai2 (approved):
Ah thanks for the changes, now the inserting is a bit easier to understand :+1:

PR #481, opened 2023-11-28
add kafka exporter to quickstart

dtrai2:


quickstart/docker-compose.yml
```diff
@@ -62,6 +62,13 @@ services:
       - /var/run/docker.sock:/var/run/docker.sock
     command: sh -c "((sleep 15 && echo 'kafka up' && kafka-topics.sh --create --if-not-exists --bootstrap-server 127.0.0.1:9092 --replication-factor 1 --partitions 3 --topic consumer)&) && /opt/bitnami/scripts/kafka/run.sh"
 
+  kafka-exporter:
+    image: bitnami/kafka-exporter
+    command: [ "--kafka.server=127.0.0.1:9092" ]
+    network_mode: host
+    expose:
+      - 9308
```
ekneg54, line 70:
```suggestion
    healthcheck:
      test:
        [
          "CMD-SHELL",
          "kafka-topics.sh --bootstrap-server 127.0.0.1:9092 --topic consumer --describe"
        ]

  kafka-exporter:
    image: bitnami/kafka-exporter
    container_name: kafkaexporter
    command: [ "--kafka.server=127.0.0.1:9092" ]
    network_mode: host
    expose:
      - 9308
    depends_on:
      kafka:
        condition: service_healthy
```

because exporter complains that kafka is not available

ekneg54 (changes requested):
I cant see any metrics for kafka.
without starting logprep the grafana explore panel complains that no metrics were found

dtrai2:
> I cant see any metrics for kafka. without starting logprep the grafana explore panel complains that no metrics were found

I just checked the dashboard again and I don't know regarding the explore panel, but I figured out why the dashboard did only show "no data" (if no logprep is running). The variables of the dashboard used metrics related to consumer groups. Apparently those are not available yet without an actual consumer. I did fix that now, meaning the variables are correct directly from the start on. Problem is though that there are two visualizations that require the consumer group information meaning those will still show "no data" until logprep actually started.

PR #486, opened 2023-11-29
Fix tryout kafka commit on revoke

dtrai2:


ekneg54:
we should refactor all output connectors to a common interface where the backlog is written.
give the abc.output.Output cconnector an abstract method `write_backlog` and ensure that all Connectors use this to do their backlog flushing actions (if they have a backlog).

ekneg54 (approved):
as disscussed today we merge this to get a new release and unblock users of elasticsearch and opensearch outputs

PR #498, opened 2023-12-12
fix opensearch output connector

### LINKED ISSUE #497: Opensearch output connector broken in 9.0.1
currentely the opensearch output connector is broken. no logs were delivered because of using `parallel_bulk` api in `opensearch-py`.
affected versions: v9.0.1 upwards.

ekneg54:
will close #497

* implements the actual logic for `parallel_bulk` which only returns a generator, that has to be consumed
* add `MSGPECSerializer` to use the much faster serialisation by `msgspec` lib
* make all threading parameter of `parallel_bulk` configurable but with sane defaults
* you have to tune the `parallel_bulk` settings to get an increase in performance, but these setting depend a lot on the actual environment

* we tried with 
  - `message_backlog_size`: 160.000
  - `chunk_size`: 10.000
  - `thread_count`: 16
  - `queue_size`: 16
* this leads to a much better performance, but also to a much higher memory consumption (trade performance for memory)

### DIFF
--- a/quickstart/exampledata/config/prometheus/prometheus.yml
+++ b/quickstart/exampledata/config/prometheus/prometheus.yml
@@ -29,7 +29,7 @@ scrape_configs:
       - targets: ["localhost:9090"]
   - job_name: "logprep"
     static_configs:
-      - targets: ["localhost:8000"]
+      - targets: ["localhost:8000", "localhost:8001"]
   - job_name: "kafka"
     metrics_path: "/metrics"
     static_configs:
--- a/tests/unit/connector/test_opensearch_output.py
+++ b/tests/unit/connector/test_opensearch_output.py
@@ -32,7 +32,7 @@ class NotJsonSerializableMock:
 
 in_ci = os.environ.get("GITHUB_ACTIONS") == "true"
 
-# helpers.bulk = mock.MagicMock()
+helpers.parallel_bulk = mock.MagicMock()
 
 
 class TestOpenSearchOutput(BaseOutputTestCase):
@@ -48,7 +48,7 @@ class TestOpenSearchOutput(BaseOutputTestCase):
     def test_describe_returns_output(self):
         assert (
             self.object.describe()
-            == "OpensearchOutput (Test Instance Name) - Opensearch Output: ['host:123']"
+            == "OpensearchOutput (Test Instance Name) - Opensearch Output: ['localhost:9200']"
         )
 
     def test_store_sends_to_default_index(self):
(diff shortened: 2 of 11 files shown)


quickstart/exampledata/config/prometheus/prometheus.yml
```diff
@@ -29,7 +29,7 @@ scrape_configs:
       - targets: ["localhost:9090"]
   - job_name: "logprep"
     static_configs:
-      - targets: ["localhost:8000"]
+      - targets: ["localhost:8000", "localhost:8001"]
```
dtrai2, line 32: Why add here another port? I didn't see that you added a second instance to the docker-compose file?

tests/unit/connector/test_opensearch_output.py
```diff
@@ -360,3 +366,27 @@ def test_message_backlog_is_cleared_after_it_was_written(self):
         self.object._config.message_backlog_size = 1
         self.object.store({"event": "test_event"})
         assert len(self.object._message_backlog) == 0
+
+    @pytest.mark.skip(reason="This test is only for local debugging")
```
dtrai2, line 370: Can't we turn this into a similar acceptance test like you did for Kafka? So that the tests starts and opensearch instance against which this test can actually run then.

tests/unit/connector/test_opensearch_output.py
```diff
@@ -360,3 +366,27 @@ def test_message_backlog_is_cleared_after_it_was_written(self):
...
+            index="defaultindex", body={"query": {"match": {"foo": uuid_str}}}
+        )
+        len_before = len(result["hits"]["hits"])
+        output._message_backlog = [{"foo": uuid_str, "_index": "defaultindex"}]
+        output._write_backlog()
+        time.sleep(1)
+        result = output._search_context.search(
+            index="defaultindex", body={"query": {"match": {"foo": uuid_str}}}
+        )
+        assert len(result["hits"]["hits"]) > len_before
```
dtrai2, line 392: If you'd turn it into a proper test we could also query for a specific document with a timestamp.

dtrai2 (changes requested):

quickstart/exampledata/config/prometheus/prometheus.yml
ekneg54: this is if you want to start a second logprep on your machine.

tests/unit/connector/test_opensearch_output.py
ekneg54: the overhead is pretty high for this... opensearch starts slow.... I would prefer to not do this

tests/unit/connector/test_opensearch_output.py
ekneg54: ?

dtrai2: The idea would be that you create an event with a certain ID (randomized) and then you query for exactly this ID. That way you know that the write call did actually work. 
But anyway, it's not important if we don't turn it into a proper test.

dtrai2 (approved):

PR #499, opened 2023-12-15
Adapt s3 connector for kafka fix

ppcad:


logprep/connector/s3/output.py
```diff
@@ -195,52 +198,57 @@ def _write_to_s3_resource(self, document: dict, prefix: str):
...
-
-    def _write_document_batches(self, message_backlog):
-        self._logger.info(f"Writing {self._current_backlog_count + 1} documents to s3")
-        for prefix_mb, document_batch in message_backlog.items():
+        if self._get_backlog_size(self._message_backlog) >= self._config.message_backlog_size:
+            self._write_backlog()
+
+    @staticmethod
+    def _get_backlog_size(message_backlog: dict):
+        return sum(len(values) for values in message_backlog.values())
```
ekneg54, line 211:
```suggestion
    @property
    def _backlog_size():
        return sum(map(len, self._message_backlog.values()))
```
how about somethin like this, to avoid looping? And to make access to this easier?
not testet, but should do the trick...

logprep/connector/s3/output.py
```diff
@@ -195,52 +198,57 @@ def _write_to_s3_resource(self, document: dict, prefix: str):
...
+        """Write to s3 if it is not already writing."""
+        if not self._message_backlog:
+            return
+
+        self._bulk()
+
+    def _bulk(self):
+        self._logger.info(
+            "Writing %s documents to s3", self._get_backlog_size(self._message_backlog)
+        )
```
ekneg54, line 223: is this really necessary? Or schould this be a DEBUG info?

ekneg54 (changes requested):
Hey @ppcad, 

happy new year. Thank you for the PR and here is my review.
Two minor suggestions to improve it. Please have look.

logprep/connector/s3/output.py
ppcad: It is helpful to see the exact timestamps when something was written and which warnings or errors happened in relation to that.
Our debug output is very verbose at the moment, so that it's hard to use it for things like this.

ekneg54 (approved):
Thank you for your contribution

PR #503, opened 2024-01-05
add do nothing option to dummy output

ekneg54:
this adds a do nothing option do dummy output to ensure dummy does not fill memory


djkhl (approved):
looks good!

PR #505, opened 2024-01-12
Ensure no data is lost

ppcad:
* remove `call_input_callback` parameter from s3 output connector
* do not call `batch_finished_callback` for extra data in s3, elasticsearch and opensearch output to prevent updating offsets for not fully processed events
* do not call `batch_finished_callback` on `revoke_callback` in confluent kafka input to prevent possible data loss for not fully processed events
* allow multiple default output connectors per input and ensure that the offset is only set for events that all outputs have written

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## next release
 ### Breaking
 
+* remove `call_input_callback` parameter from s3 output connector
+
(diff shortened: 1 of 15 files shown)


ppcad:
This pull request won't be merged, since support for multiple output connectors will be dropped in the future.

PR #510, opened 2024-01-23
remove logger from Factory

ekneg54:
this removes the logger parameter from `Factory.create` method, because it is not needed.


ekneg54:
will retry later

PR #513, opened 2024-01-24
Revise CLI

dtrai2:


ekneg54:
following failures:

image

`python ./logprep/run_logprep.py test unit  ./quickstart/exampledata/config/pipeline.yml` leads to:

image

dtrai2:
I fixed the first problem you mentioned, but the `print` call still just raises an `NotImplementedError` and the second issue is not an issue of the cli but of the AutoRuleTester. The Quickstart configuration doesn't have tests for rules and then it comes to this problem. If you call `logprep test unit <CONFIG>` with a CONFIG that has tests, then it works fine.

ekneg54 (approved):

PR #514, opened 2024-01-26
Fix handling of `CriticalInputError` exceptions

clumsy9:
This PR fixes an error in the `_handle_pipeline_error()` function that leads to `FatalOutputError` when `CriticalInputError` exceptions are handled.


ekneg54 (approved):
Thank you for your contribution. I updated the Changelog to reflect this bugfix in our next release

PR #515, opened 2024-01-29
Revert not exists changes from #480, but keep refactorings in filter_expression

ppcad:


CHANGELOG.md
```diff
@@ -80,7 +80,6 @@
   - add lru caching for peudonymizatin of urls
 * improve loading times for the rule tree by optimizing the rule segmentation and sorting
 * add support for python 3.12 and remove support for python 3.9
-* always check the existence of a field for negated key-value based lucene filter expressions
```
dtrai2, line 83: Can you please leave this line inside the changelog and also add a new one with this change to the upcoming release? Otherwise it is not transparent and also wrong, as it wouldn't reflect the actual 9.0.0 release on pypi.

dtrai2 (changes requested):

CHANGELOG.md
ppcad: You are right, that was a mistake. I've changed it now.

PR #518, opened 2024-02-02
Set no offsets on store custom

ppcad:
Make `store_custom` in s3, opensearch and elasticsearch connector not call `batch_finished_callback` to prevent data loss that could be caused by partially processed events.

### DIFF
--- a/logprep/connector/elasticsearch/output.py
+++ b/logprep/connector/elasticsearch/output.py
@@ -212,10 +212,16 @@ class ElasticsearchOutput(Output):
 
         self._add_dates(document)
         self.metrics.number_of_processed_events += 1
-        self._write_to_search_context(document)
+        self._message_backlog.append(document)
+        self._write_to_search_context()
 
     def store_custom(self, document: dict, target: str):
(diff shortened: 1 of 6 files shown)


logprep/connector/elasticsearch/output.py
```diff
@@ -231,8 +237,7 @@ def store_custom(self, document: dict, target: str):
         """
         document["_index"] = target
         self._add_dates(document)
-        self.metrics.number_of_processed_events += 1
```
ekneg54, line 234: this is needed because it was built in to check via the ratio from input to output events if the output using processors work. from the perspective of the output connector a SRE or pseudonym is an event.

ekneg54 (changes requested):
thank you for your PR. please have a look on my comment

logprep/connector/elasticsearch/output.py
ppcad: Thanks, I have re-added it.

ekneg54 (approved):
Tank you for your changes

PR #526, opened 2024-02-11
pyproject.toml dependency and build managment

ekneg54:


dtrai2 (changes requested):
Looks quite good already. Guess we just have to try it out and see how well it will works. I just found some more mentions of the `requirements.txt` or `requirements_dev.txt` though. Just do a project wide search for those names and you should find them too. Some are for example in the files:
- `doc/source/getting_started.rst`
- `.devcontainer/Dockerfile` (as an optional comment though)
- `.github/workflows/ci.yml`
- `.github/workflows/main.yml`
- `.readthedocs.yaml`
- `README.md`
- ...

ekneg54:
@ppcad @dtrai2 
please have a look again. the pipeline is still failing for black. But I recommend to add the additional 74 format dependent fixes after review :-D

pyproject.toml
```diff
@@ -1,3 +1,122 @@
...
+packages = ["logprep"]
+
+[tool.setuptools.dynamic]
+version = { attr = "logprep.__version__" }
+
+[project]
+name = "logprep"
+description = "Logprep allows to collect, process and forward log messages from various data sources."
+requires-python = ">=3.10"
+# version = "0.1.0"
```
dtrai2, line 15:
```suggestion
```

dtrai2 (changes requested):
I just found one small line that you can still also remove. Could you also add an entry in the changelog? The question is are we considereing this as a breaking change? The software itself didn't change, but the install procedure of logprep did, and if someone depends on the `requirements.txt` they have a problem now.

dtrai2 (approved):

PR #544, opened 2024-03-15
Update FDA to version 0.21.0 in quickstart example

clumsy9:
This PR udpates the quickstart example to use FDA version 0.20.0.


ekneg54 (approved):
thank you for your contribution. I updated the version to 0.21.0 which was released yesterday.

PR #549, opened 2024-03-22
exchange version string generation from versioneer to setuptools-scm

### LINKED ISSUE #533: Wrong logprep version in feature branch container images
Placeholder

Container images that are created for feature branches contain wrong version info like: `logprep-0+20...untaged...`
Also true for the pex-build

Intended behaviour should be the correct version like `logprep-10.0.3`

### LINKED ISSUE #546: logprep version string broken for releases 10.0.4
image

to reproduce:
```
pip install logprep==10.0.4

logprep --version
```

ekneg54:
# TODO:

- [x] use setuptools_scm for version generation
- [x] fix tests that depend on `get_version` by reimplementing it
- [x] exclude unneeded files from package
- [x] fix version for containerbuilds

closes #533
closes #549


ppcad:
When running the pex for this pull-request I get version 0.1.dev1+g3650834.
Is that the desired behavior?

ppcad (commented):
When running the pex for this pull-request I get version 0.1.dev1+g3650834.
Is that the desired behavior?

ekneg54:
It is not the desired behavior for sure :)
But pex build will be deleted in near future as I've heard ;)
Do you think it is worth it to fix this?

PR #556, opened 2024-03-27
fix merge of version string for multiple configs

ekneg54:
This PR fixes the version generation for multiple configs.

the resulting configuration version is now concatenated in the same order as the configs were given.


ekneg54:
@djkhl have you installed logprep dependencies with `pip install -e .[dev]`?

I tried the same with success, but for me the logprep version differs from yours:

image

djkhl (approved):
It works perfectly now thank you!

PR #568, opened 2024-04-18
fix broken gracefull shutdown

ekneg54:
testing this is hard. too hard for me and so I decided to not add a tests for this.

test the fix explorative with:

```bash
logprep  run ./quickstart/exampledata/config/http_pipeline.yml
```

then press `CRTL + C`

you should get this:

```
^C2024-04-19 15:41:18,980 Logprep Runner INFO    : Shutting down
2024-04-19 15:41:18,980 Logprep Runner INFO    : Initiated shutdown
2024-04-19 15:41:18,980 Logprep Pipeline 2 DEBUG   : Stopping pipeline (MainProcess)
2024-04-19 15:41:18,980 Logprep Pipeline 2 DEBUG   : Stopping pipeline (MainProcess)
2024-04-19 15:41:20,659 Logprep Pipeline 1 DEBUG   : Stopping pipeline (MainProcess)
2024-04-19 15:41:20,659 Logprep Pipeline 1 DEBUG   : Stopping pipeline (MainProcess)
2024-04-19 15:41:25,673 Logprep HTTPServer INFO    : Shutting down
2024-04-19 15:41:25,778 Logprep HTTPServer ERROR   : Cancel 0 running task(s), timeout graceful shutdown exceeded
2024-04-19 15:41:25,779 Logprep HTTPServer INFO    : Finished server process [229968]
2024-04-19 15:41:25,788 Prometheus Exporter INFO    : Cleaned up /tmp/logprep1
2024-04-19 15:41:25,791 Logprep Runner INFO    : Shutdown complete
```

### DIFF
--- a/logprep/connector/http/input.py
+++ b/logprep/connector/http/input.py
@@ -370,6 +370,8 @@ class HttpConnector(Input):
         self.port = self._config.uvicorn_config["port"]
         self.host = self._config.uvicorn_config["host"]
         self.target = f"http://{self.host}:{self.port}"
+        if self.messages is None:  # this is only for testing. Is set by pipeline_manager
+            self.messages = mp.Queue(maxsize=self._config.message_backlog_size)
 
     def setup(self):
         """setup starts the actual functionality of this connector.
@@ -382,6 +384,7 @@ class HttpConnector(Input):
             raise FatalInputError(
                 self, "Necessary instance attribute `pipeline_index` could not be found."
             )
+
         self._logger.debug(
             f"HttpInput Connector started on target {self.target} and "
             f"queue {id(self.messages)} "
--- a/logprep/connector/http/input.py
+++ b/logprep/connector/http/input.py
@@ -364,7 +364,7 @@ class HttpConnector(Input):
         internal_uvicorn_config = {
             "lifespan": "off",
             "loop": "asyncio",
-            "timeout_graceful_shutdown": 0,
+            "timeout_graceful_shutdown": 5,
         }
         self._config.uvicorn_config.update(internal_uvicorn_config)
         self.port = self._config.uvicorn_config["port"]
@@ -385,7 +385,7 @@ class HttpConnector(Input):
         self._logger.debug(
             f"HttpInput Connector started on target {self.target} and "
             f"queue {id(self.messages)} "
-            f"with queue_size: {self.messages._maxsize}"
+            f"with queue_size: {self.messages._maxsize}"  # pylint: disable=protected-access
         )
         # Start HTTP Input only when in first process
         if self.pipeline_index != 1:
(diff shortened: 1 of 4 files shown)


ekneg54:
the strange thing is that it depends on how logprep is startet.

by starting it with this the problem occurs:

```
logprep run ./quickstart/exampledata/config/pipeline.yml    
```

by starting it the old way the old gracefull shutdown behaviour is triggered:

```
$env:PYTHONPATH="."; python ./logprep/run_logprep.py run ./quickstart/exampledata/config/pipeline.yml
```

ekneg54:
the problem was, that the startup script installed via pyproject.toml calls the `cli()` function directly. So the signal handlers were never registered and the graceful shutdown behaviour was lost.

logprep/connector/http/input.py
```diff
@@ -364,12 +364,14 @@ def __init__(self, name: str, configuration: "HttpConnector.Config", logger: Log
...
             "loop": "asyncio",
-            "timeout_graceful_shutdown": 0,
+            "timeout_graceful_shutdown": 5,
         }
         self._config.uvicorn_config.update(internal_uvicorn_config)
         self.port = self._config.uvicorn_config["port"]
         self.host = self._config.uvicorn_config["host"]
         self.target = f"http://{self.host}:{self.port}"
+        if self.messages is None:  # this is only for testing. Is set by pipeline_manager
+            self.messages = mp.Queue(maxsize=self._config.message_backlog_size)
```
dtrai2, line 374: If we only need this in testing, can't we also set the attribute in the corresponding test class then? Besides, I just executed the test without these lines and all the test run just fine.

logprep/run_logprep.py
```diff
@@ -60,6 +58,9 @@ def cli() -> None:
     Logprep allows to collect, process and forward log messages from various data sources.
     Log messages are being read and written by so-called connectors.
     """
+    if "pytest" not in sys.modules:  # needed for not blocking tests
+        signal.signal(signal.SIGTERM, signal_handler)
+        signal.signal(signal.SIGINT, signal_handler)
```
dtrai2, line 63: How about we add a test that checks if these signals actual work as promised? Like running logprep with `process = subprocess.Popen()`, `process.send_signal()` and then checking if the process actually terminated as expected?

ekneg54: I tried this, but it did not work. 
I guess the reason is, that the logprep process name if started with subprocess is not "MainProcess". So the triggers do not work.

logprep/connector/http/input.py
ekneg54: in my setup the tests hang if run without these lines. So the ci pipline tests deadlocks also.

dtrai2: ah okay strange, but okay. Thanks for clearing that up for me.

logprep/run_logprep.py
dtrai2: too bad. but good to know that you tried it already, thanks! :+1:

dtrai2 (approved):

PR #571, opened 2024-04-23
Fix grokker rules that have fields with common prefixes

### LINKED ISSUE #564: Grokker won't allow for extension of already existing field
If a grokker rule contains two similiar fields and the second one is the same only extended by "_something", then the field wont get written.

Example: ......... ( [A-Za-z: ]*\%)?(%{CISCO_REASON:cisco.**facility**}......%{CISCO_REASON:cisco.**facility_mnoe**}...................

dtrai2:
closes #564

### DIFF
--- a/logprep/processor/grokker/rule.py
+++ b/logprep/processor/grokker/rule.py
@@ -68,10 +68,14 @@ def _dotted_field_to_logstash_converter(mapping: dict) -> dict:
     def _transform(pattern):  # nosemgrep
         fields = re.findall(FIELD_PATTERN, pattern)
         for dotted_field, _ in fields:
-            splitted_field = dotted_field.split(".")
-            if len(splitted_field) > 1:
-                replacement = "".join(f"[{element}]" for element in splitted_field)
-                pattern = re.sub(re.escape(dotted_field), replacement, pattern)
+            split_field = dotted_field.split(".")
+            if len(split_field) > 1:
+                replacement = "".join(f"[{element}]" for element in split_field)
+                # ensure full field is replaced by scanning for ':' at the front and '}' or ':'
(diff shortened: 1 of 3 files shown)


logprep/processor/grokker/rule.py
```diff
@@ -68,10 +68,14 @@ def _dotted_field_to_logstash_converter(mapping: dict) -> dict:
     def _transform(pattern):  # nosemgrep
         fields = re.findall(FIELD_PATTERN, pattern)
         for dotted_field, _ in fields:
-            splitted_field = dotted_field.split(".")
-            if len(splitted_field) > 1:
-                replacement = "".join(f"[{element}]" for element in splitted_field)
-                pattern = re.sub(re.escape(dotted_field), replacement, pattern)
+            split_field = dotted_field.split(".")
+            if len(split_field) > 1:
```
ekneg54, line 72:
```suggestion
            if "." in dotted_field:
```
* if a dot is in dotted field, the length will be greater than 1
* much cheaper operation than two method calls

ekneg54 (changes requested):
thank you for your contribution.
please consider to have a look at one minor change

logprep/processor/grokker/rule.py
dtrai2: Well we still need the `dotted_field.split(".")` in the next line. So what I could offer is something like following. But it doesn't really reduce the method calls except the change of `len(...) > 1` to `"." in dotted_field`.
```python
if "." in dotted_field:
    replacement = "".join(f"[{element}]" for element in dotted_field.split("."))
```

ekneg54: still better 👍

ekneg54 (approved):

PR #577, opened 2024-04-25
Handle Opensearch Stacktraces

dtrai2:


dtrai2:
> I think there is an except missing somewhere?

I did investigate a bit and we do catch this exception already:
```bash
2024-05-02 08:47:28,608 Logprep Pipeline 1 ERROR   : FatalOutputError in OpensearchOutput (opensearch) - Opensearch Output: ['127.0.0.1:9200']: ConnectionError(<urllib3.connection.HTTPConnection object at 0x720f9feb8c90>: Failed to establish a new connection: [Errno 111] Connection refused) caused by: NewConnectionError(<urllib3.connection.HTTPConnection object at 0x720f9feb8c90>: Failed to establish a new connection: [Errno 111] Connection refused)
```
It is thrown inside the `logprep/connector/elasticsearch/output.py` > `ElasticsearchOutput.setup()` when calling `self._search_context.info()`. Two lines below that there is also the `except` block that catches the `Connection refused` error. The problem is though that the opensearch library still logs the stacktrace manually, with a `logger.warning` call. The only option I found so far would be to raise the loglevel of the opensearch logger to `ERROR` with following line inside the `setup()` method:

```python
logging.getLogger("opensearch").setLevel(logging.ERROR)
```

PR #578, opened 2024-04-26
add http input connector metrics

ekneg54:
* this adds 2 new metrics to `http_input` connector

* `message_backlog_size` implemented as gauge which represents the size of the shared queue
* `number_of_http_requests` implemented as counter which represents obviously the number of incoming requests


djkhl (approved):
thank you for implementing the helpful additional metric! 
everything looks very good!

PR #582, opened 2024-05-03
integrate UCL into quickstart setup

dtrai2:


djkhl (approved):
everything looks good and works as expected :)

PR #590, opened 2024-05-15
add queuelistener for logging

ekneg54:


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 ### Improvements
 
 * remove logger from Components and Factory signatures
+* add `LogprepMPQueueListener` to outsource logging to a separate process
+* add a single `Queuehandler` handler to root logger to ensure all logs were handled by `LogprepMPQueueListener`
```
dtrai2, line 14:
```suggestion
* add a single `Queuehandler` to root logger to ensure all logs were handled by `LogprepMPQueueListener`
```

logprep/runner.py
```diff
@@ -151,6 +151,8 @@ def start(self):
         self._logger.info("Shutting down")
         self._manager.stop()
         self._logger.info("Shutdown complete")
+        if self._manager.loghandler is not None:
+            self._manager.loghandler.stop()
```
dtrai2, line 155:
```suggestion
        if self._manager.loghandler is not None:
            self._manager.loghandler.stop()
        self._logger.info("Shutdown complete")
```
I hope we can still use the regular logger after the loghandler was stopped.

tests/unit/util/test_logging.py
```diff
@@ -8,12 +12,42 @@
...
+        logger = logging.getLogger("console")
+        assert isinstance(logger.handlers[0].formatter, LogprepFormatter)
 
-    def test_default_log_config(self):
-        logging.config.dictConfig(DEFAULT_LOG_CONFIG)
-        logger = logging.getLogger("test")
-        assert isinstance(logger.root.handlers[0].formatter, LogprepFormatter)
+    def test_root_logger_has_only_quehandler(self):
+        logger = logging.getLogger("root")
+        assert isinstance(logger.handlers[0], logging.handlers.QueueHandler)
```
dtrai2, line 28: according to the test title you also need one more assertions
```suggestion
        assert len(logger.handlers) == 1, "only one handler should be set"
        assert isinstance(logger.handlers[0], logging.handlers.QueueHandler)
```

dtrai2 (changes requested):

logprep/runner.py
ekneg54: no ... that's the reason, this won't work. After stopping the handler, no logs will be handled bei the console handler

dtrai2 (approved):

PR #593, opened 2024-05-16
make extend_target_list preserve ordering

### LINKED ISSUE #592: Field Manager - source fields are not appended to end of list
**Expected behavior**
As a user of the Field Manager, I expect source fields to be added to the end of a list, if I make use of the `extend_target_list` feature. I expect the list items of the target list to remain in the same order, and the source fields to be appended to the end of the list in the order given in the `source_fields` array (also in case of a newly created target list).

**Current behavior**
At the moment, the output list seems to be sorted alphabetically, which is not the desired behavior since it messes up the original list order.

**Steps to reproduce**
The behavior above can be reproduced as follows, without any use of FDA.

Create the following event:
```
{"existing_list": ["hello", "world"]}
```

Create the following pipeline:

```
pipeline:
- 0_generic_adder:
    generic_rules:
    - filter: '*'
      generic_adder:
        id: 9effca73-7f76-427f-9c8e-08d871d69fce
        add:
          foo: bar
          test: value
        overwrite_target: false
    specific_rules: []
    type: generic_adder
- 1_field_manager:
    generic_rules:
    - filter: (foo) OR (test)
      field_manager:
        id: 5cfa7a26-94af-49de-bc82-460c42e9dc56
        source_fields:
        - foo
        - test
        target_field: existing_list
        delete_source_fields: false
        overwrite_target: false
        extend_target_list: true
    specific_rules: []
    type: field_manager
```

The pipeline creates two fields, `"foo":"bar"` and `"test":"value"` and appends them to the list `existing_list:["hello", "world"]`

Run Logprep by using the CLI command `logprep run /path_to_pipeline/pipeline.yml`

Output document:
```
{'existing_list': ['bar', 'hello', 'value', 'world'],
 'foo': 'bar',
 'test': 'value'}
```

Expected output document:
```
{'existing_list': ['hello', 'world', 'bar', 'value'],
 'foo': 'bar',
 'test': 'value'}
```
The expected list contains the original values in the correct order `hello` and `world`, appended by `bar` and `value`.

**Environment**

Logprep version: 11.3.0, installed from pypi
Python version: 3.10.12

**Possible solution**
Maintain original order of target list and append source fields at the end of list, in the order given in the `source_fields` array.

ekneg54:
fixes #592
closes #592


tests/unit/processor/field_manager/test_field_manager.py
```diff
@@ -187,7 +187,7 @@
             "field3": ["value5", "value6"],
             "new_field": ["i exist"],
         },
-        {"new_field": ["i exist", "value1", "value2", "value3", "value4", "value5", "value6"]},
+        {"new_field": ["i exist", "value1", "value2", "value3", "value5", "value6", "value4"]},
```
ekneg54, line 190: the fields should be appended as the ordering in source_fields.

```suggestion
        {"new_field": ["i exist", "value1", "value2", "value3", "value4", "value5", "value6"]},
```

tests/unit/processor/field_manager/test_field_manager.py
```diff
@@ -209,7 +209,7 @@
             "field3": ["value5", "value6", "value4"],
             "new_field": ["i exist"],
         },
-        {"new_field": ["i exist", "value1", "value2", "value3", "value4", "value5", "value6"]},
+        {"new_field": ["i exist", "value1", "value2", "value3", "value5", "value6", "value4"]},
```
ekneg54, line 212: same but different. duplicates could be resolved in first comes first serves manner:

```suggestion
        {"new_field": ["i exist", "value1", "value2", "value3", "value5", "value4", "value6",]},
```

tests/unit/processor/field_manager/test_field_manager.py
```diff
@@ -232,7 +232,7 @@
             "field3": ["value5", "value6", "value4"],
             "new_field": ["i exist"],
         },
-        {"new_field": ["value1", "value2", "value3", "value4", "value5", "value6"]},
+        {"new_field": ["value1", "value2", "value3", "value5", "value6", "value4"]},
```
ekneg54, line 235: same

```suggestion
        {"new_field": ["value1", "value2", "value3", "value5", "value4", "value6"]},
```

djkhl (approved):

PR #601, opened 2024-06-08
add calling setup during configuration verify

### LINKED ISSUE #597: Pseudonymizer no validation of existing files on logprep start
Configuring a pseudonymizer with files that do not exist results in stacktraces.

**Expected behavior**
On logprep start there should be once an error message that logprep couldn't start because it wasn't configured properly.

**Current behavior**
Starts but repeatedly prints stacktraces / error log messages

**Steps to reproduce**
Start logprep with the following configuration: 

```yaml
input:
    dummy:
        type: dummy_input
        documents: [{"foo":"bar"}]
output:
    dummy:
        type: dummy_output
pipeline:
    - pseudonymizername:
        type: pseudonymizer
        specific_rules:
            - filter: foo
              pseudonymizer:
                  mapping:
                      foo: RE_WHOLE_FIELD_CAP
        generic_rules: []
        outputs:
            - dummy: pseudonyms_topic
        pubkey_analyst: /path/to/analyst_pubkey.pem
        pubkey_depseudo: /path/to/depseudo_pubkey.pem
        hash_salt: secret_salt
        regex_mapping: /path/to/regex_mapping.json
        max_cached_pseudonyms: 1000000
```

Do different start ups with (non)existing paths for the fields `regex_mapping`, `pubkey_analyst`/`pubkey_depseudo` and watch the different error behavior. 

**Environment**

Logprep version: 856ceaf8
Python version: 3.11.9

**Possible solution**
Validate file paths on logprep start, before the pipeline object is created.

ekneg54:
it turns out, that the bug should be handled for all missing files during verification of Configuration.

closes #597 

TODO:

- [x] write tests for populating cached properties during setup time

### DIFF
--- a/logprep/util/configuration.py
+++ b/logprep/util/configuration.py
@@ -816,9 +816,12 @@ class Configuration:
         for processor_config in self.pipeline:
             try:
                 processor = Factory.create(deepcopy(processor_config))
+                processor.setup()
                 self._verify_rules(processor)
             except (FactoryError, TypeError, ValueError, InvalidRuleDefinitionError) as error:
                 errors.append(error)
+            except FileNotFoundError as error:
+                errors.append(InvalidConfigurationError(f"File not found: {error.filename}"))
             try:
                 self._verify_processor_outputs(processor_config)
             except Exception as error:  # pylint: disable=broad-except
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 * add `LogprepMPQueueListener` to outsource logging to a separate process
 * add a single `Queuehandler` to root logger to ensure all logs were handled by `LogprepMPQueueListener`
 * refactor `http_generator` to use a logprep http output connector
+* ensure all `cached_properties` are populated during setup time
 
 ### Bugfix
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 ### Bugfix
 
 * make `--username` and `--password` parameters optional in http generator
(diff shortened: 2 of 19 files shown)


logprep/util/configuration.py
```diff
@@ -816,9 +816,12 @@ def _verify(self):
         for processor_config in self.pipeline:
             try:
                 processor = Factory.create(deepcopy(processor_config))
+                processor.setup()
```
dtrai2, line 819: That's a good start, but I don't think calling `setup` will suffice. The reason being that we often gather resources through `properties`. In the pseudonymizer for example the public keys are loaded not in the `setup` method but via the `_encrypter` property which again is called for the first time inside the `_pseudonymize` method. So with the current implementation I still see logging like this:
```
024-06-10 10:19:53 174478 Pipeline1  ERROR   : ProcessingCriticalError: ProcessingCriticalError: [Errno 2] No such file or directory: '/path/to/analyst_pubkey.pem' -> event was send to error output and further processing stopped, rule.id='ae29c373ba466990d81428cb2d8ff6253a35cb88dc37e9b4ef5e9669333f1806', rule.description='', event={'foo': 'bar'}
```

dtrai2 (changes requested):
The current solution is sadly not enough yet, see other comment for an explanation. Besides that could you also please add a changelog entry.

logprep/util/configuration.py
ekneg54: fixed by calling all cached_properties during setup time

dtrai2 (approved):

PR #602, opened 2024-06-10
revise restart behavior

ekneg54:
TODO:

- [x] fix not finishing tests
- [x] add documentation about exit codes
- [x] add documentation about restart behavior
- [x] make `restart_count` a logprep config option
- [x] add old behavior on negative `restart_count`

### DIFF
--- a/logprep/framework/pipeline_manager.py
+++ b/logprep/framework/pipeline_manager.py
@@ -127,6 +127,7 @@ class PipelineManager:
         ]
 
         if not failed_pipelines:
+            self.restart_count = 0
             return
 
         for index, failed_pipeline in failed_pipelines:
--- a/logprep/framework/pipeline_manager.py
+++ b/logprep/framework/pipeline_manager.py
@@ -142,6 +142,8 @@ class PipelineManager:
                 pipeline_index,
                 exit_code,
             )
+        if self._configuration.restart_count < 0:
+            return
         self.restart_count += 1
         time.sleep(self.restart_timeout_ms / 1000)
         self.restart_timeout_ms = self.restart_timeout_ms * 2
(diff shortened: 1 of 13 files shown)


logprep/framework/pipeline_manager.py
```diff
@@ -137,6 +143,9 @@ def restart_failed_pipeline(self):
                 pipeline_index,
                 exit_code,
             )
+        self.restart_count += 1
+        time.sleep(self.restart_timeout_ms / 1000)
+        self.restart_timeout_ms = self.restart_timeout_ms * self.restart_count
```
dtrai2, line 148:
```suggestion
        self.restart_timeout_ms = self.restart_timeout_ms * 2
```

logprep/util/defaults.py
```diff
@@ -1,5 +1,18 @@
...
+from enum import Enum
+
+
+class EXITCODES(Enum):
+    """Exit codes for logprep."""
+
+    SUCCESS = 0
+    ERROR = 1
+    CONFIGURATION_ERROR = 2
+    PIPELINE_ERROR = 3
```
dtrai2, line 12: That's something we could add to the documentation. Every Enum-Entry could get a doc string explaining when these errors could happen and then you could integrate it to the shpinx doc somewhere.

logprep/util/defaults.py
```diff
@@ -1,5 +1,18 @@
...
+
+
+DEFAULT_RESTART_COUNT = 5
```
dtrai2, line 15: should we make this configurable? I think it would be nice if the end user could configure how forgiving logprep should be? Like restart_count of 0 would mean logprep stops after every error, and restart_count = 100 could mean something like never give up. And that "never give up" could even actually be implemented with something like "-1" maybe.

dtrai2 (changes requested):

logprep/util/defaults.py
ekneg54: done :D

dtrai2 (approved):

PR #604, opened 2024-06-11
fix quickstart compose setup for WSL

ekneg54:
* docker networking depends on environment
* you cant't predict that `172.17.0.1` is always the network docker uses
* configured fixed ip subnets for all networks and fixed the host entry for prometheus.


dtrai2 (approved):
looks good to me. We can merge this, but I'm afraid we have to open a bug for the kafka grafana dashboard. The kafka metrics are available in grafana, but it seems like that the metrics the dashboard expects do not exist anymore.

PR #605, opened 2024-06-11
add retry mechanism to write_backlog

dtrai2:


logprep/connector/opensearch/output.py
```diff
@@ -87,16 +89,21 @@ class Config(ElasticsearchOutput.Config):
...
             default=500, validator=[validators.instance_of(int), validators.gt(1)]
         )
         """Chunk size to use for bulk requests."""
+        bulk_retries: int = field(
+            default=5, validator=[validators.instance_of(int), validators.gt(1)]
+        )
+        """Number of retries to use for bulk requests, if OpenSearch returns a 429
+        rejected_execution_exception. Between each retry there will be a small delay, starting
+        with one second. With each consecutive retry 500 to 1000 ms will be added to the delay,
+        chosen randomly. :code:`bulk_retries` is optional and has a default value of 5."""
```
ekneg54, line 106: you have the `max_retries` parameter on this Component (have a look in elasticsearch_output). Why not use this parameter?

logprep/connector/opensearch/output.py
```diff
@@ -124,7 +131,7 @@ def describe(self) -> str:
     def _bulk(self, client, actions, *args, **kwargs):
         try:
             if self._config.parallel_bulk:
-                self._parallel_bulk(client, actions, *args, **kwargs)
+                self._parallel_bulk_with_retries(client, actions, *args, **kwargs)
                 return
```
ekneg54, line 135: how about extending the signature of this `_bulk` method with `max_retries` like in the eleasticsearch implementation?

logprep/connector/opensearch/output.py
```diff
@@ -136,7 +143,24 @@ def _bulk(self, client, actions, *args, **kwargs):
...
+            except search.ConnectionError as error:
+                raise error
+            except search.exceptions.TransportError as error:
+                if self._message_exceeds_max_size_error(error):
+                    raise error
+                time.sleep(bulk_delays)
+                bulk_delays += random.randint(500, 1000) / 1000
+        raise FatalOutputError(
+            self, "Opensearch too many requests, all parallel bulk retries failed"
+        )
```
ekneg54, line 161: instead of writing this new method, how about extending the called method and add the `max_retries` parameter to the signature and use it?

ekneg54 (changes requested):
thanks for the contribution. Please have a look for my comments on your implementation

logprep/connector/opensearch/output.py
dtrai2: well I implemented it as a second method to reduce nesting. I reverted it now, please have a look at it and let me know which version you like more: (a) all in one method or (b) split up into two methods.

logprep/connector/opensearch/output.py
```diff
@@ -137,15 +137,29 @@ def _bulk(self, client, actions, *args, **kwargs):
...
```
ekneg54, line 165: It is one of these rare cases where a `for else loop` would be nice

```suggestion
         for _ in range(self._config.max_retries + 1):
            try:
                for success, item in helpers.parallel_bulk(
                    client,
                    actions=actions,
                    chunk_size=self._config.chunk_size,
                    queue_size=self._config.queue_size,
                    raise_on_error=True,
                    raise_on_exception=True,
                ):
                    if not success:
                        result = item[list(item.keys())[0]]
                        if "error" in result:
                            raise result.get("error")
                break
            except search.ConnectionError as error:
                raise error
            except search.exceptions.TransportError as error:
                if self._message_exceeds_max_size_error(error):
                    raise error
                time.sleep(bulk_delays)
                bulk_delays += random.randint(500, 1000) / 1000
        else:
            raise FatalOutputError(
                self, "Opensearch too many requests, all parallel bulk retries failed"
            )
```

I like the single method implementation more

ekneg54 (changes requested):
one last thing 🥇

ekneg54 (approved):

PR #616, opened 2024-06-20
field_manager does not writes to a non list target field if extend_target_list

ekneg54:


djkhl (approved):
Everything looks good and works as it should. And thank you for adding the additional case very good idea. :)

PR #622, opened 2024-06-27
replace quickstart setup by deployment examples for helm and compose

ekneg54:


doc/source/examples/compose.rst
```diff
@@ -77,12 +77,12 @@ Run with getting config from http server with mTLS authentication
 
     .. code-block:: bash
 
-      export LOGPREP_CREDENTIALS_FILE="quickstart/exampledata/config/credentials.yml"
+      export LOGPREP_CREDENTIALS_FILE="examples/exampledata/config/credentials.yml"
       logprep run https://localhost:8082/config/pipeline.yml
```
djkhl, line 81: when i run this example i get the following error:
image
is this an error that only shows up on my setup or can you reproduce it?

charts/logprep/values.yaml
```diff
@@ -103,6 +103,9 @@ logger:
 #
 # See https://logprep.readthedocs.io/en/latest/user_manual/configuration/input.html
 # for available configuration options.
+# Note:
+# For the `http_input` endpoints you have to add the endpoint `/health: plaintext` to ensure
+# readiness probes are working.
```
djkhl, line 108: do we need a new chart version because of this change?

charts/logprep/templates/deployment.yaml
```diff
@@ -14,6 +14,7 @@ spec:
       {{- include "logprep.selectorLabels" . | nindent 6 }}
   template:
     metadata:
+      name: {{ include "logprep.fullname" . }}
```
djkhl, line 17: do we need a new chart version because of these changes?

doc/source/examples/compose.rst
```diff
@@ -133,12 +133,12 @@ Utilizing FDA and UCL
...
+0. Run the example compose setup with the :code:`oauth2` profile:
    :code:`docker compose --profile oauth2 up -d`.
 1. Sign into the keycloak admin panel and create a logprep user in the :code:`logprep` realm.
    Make sure that the user is part of the :code:`logprep-admin` group and has a password. If you
    choose a password other than :code:`logprep` you have to update the credentials file
-   :code:`quickstart/exampledata/config/credentials.yml`, such that the password of
+   :code:`examples/exampledata/config/credentials.yml`, such that the password of
    :code:`http://localhost:3001` and :code:`http://localhost:3002` reflects your choice.
 2. You have to login to the FDA with the previously created user and create a release, as well
    as your first logclass. It is also necessary to add an example event to this logclass in order
```
djkhl, line 144: should we maybe change the layout of this example to be consistent with the other examples?

djkhl (changes requested):

doc/source/examples/compose.rst
ekneg54: yes... this seems to be a bug in our configuration for mtl. 
we should address this in a bug ticket to make it transparent and solve it later. It was broken before this PR.

ekneg54: added this bug: https://github.com/fkie-cad/Logprep/issues/636

djkhl (approved):

PR #635, opened 2024-07-18
add support for extraVolumes and extraMounts

ekneg54:
populate volumes and mounts from value keys
adds logprep-cache-dir in memory volume and mounts it to /home/logprep/.cache


herrfeder (approved):
looks good and even good documentation strings in the values.yaml file

PR #641, opened 2024-07-30
generalize chart version

ekneg54:
this generalizes the chart version in the example to always install the latest version


herrfeder (approved):
works with this settings.

PR #658, opened 2024-08-27
Fix throtteling queue increases progressively

ekneg54:


logprep/framework/pipeline_manager.py
```diff
@@ -40,12 +43,16 @@ def __init__(self, ctx, maxsize):
 
     def throttle(self, batch_size=1):
         """Throttle put by sleeping."""
-        time.sleep((self.wait_time**self.consumed_percent) / batch_size)
+        if self.consumed_percent > 90:
+            sleep_time = max(
+                self.wait_time, int(self.wait_time * self.consumed_percent / batch_size)
+            )
```
djkhl, line 49: i tested this and i had a feeling that the timeout does not really increase or just really really really slightly increases once the initial timeout has set in.  Did you check this yourself and is this the desired behavior?

djkhl (changes requested):

logprep/framework/pipeline_manager.py
ekneg54: Indeed it is. Look... There are tests for it

djkhl (approved):

PR #661, opened 2024-08-31
make configuration immutable

### LINKED ISSUE #660: Configuration is not immutable and can't be applied multiple times
Currently it is not possible to use one instance of a `logprep.util.configuration.Configuration` instance twice if it includes a `pseudonymizer` or `labeler` and perhaps some other processors.

These processors change their configuration by compiling regex strings to regex expressions in rules. this leads to a `TypeError` or `ValueError` if you try to instantiate the same processor class with this instance of the configuration again.

**Expected behavior**

the behavior should be idempotent. a configuration should be applicable multiple times to a new component class and the result should always be the same.  this could be achieved by initially make the configuration attrs instance frozen

**Current behavior**
image

**Steps to reproduce**

try to instanciate a `logprep.framework.pipeline.Pipeline` class twice with `./examples/exampledata/config/pipeline.yml`

**Environment**

Logprep version: 13.1.2
Python version: 3.11

**Possible solution**

* first make all configuration frozen
* second find another way to implement the failing tasks

**Possible implementation**

none

ekneg54:
closes #660

### DIFF
--- a/tests/unit/util/test_configuration.py
+++ b/tests/unit/util/test_configuration.py
@@ -10,7 +10,6 @@ from unittest import mock
 import pytest
 import responses
 from attrs import asdict
-from attrs.exceptions import FrozenInstanceError
 from requests.exceptions import HTTPError
 from ruamel.yaml.scanner import ScannerError
 
--- a/tests/unit/util/test_configuration.py
+++ b/tests/unit/util/test_configuration.py
@@ -10,6 +10,7 @@ from unittest import mock
 import pytest
 import responses
 from attrs import asdict
+from attrs.exceptions import FrozenInstanceError
 from requests.exceptions import HTTPError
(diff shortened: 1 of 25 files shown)


tests/unit/util/test_configuration.py
```diff
@@ -10,6 +10,7 @@
 import pytest
 import responses
 from attrs import asdict
+from attrs.exceptions import FrozenInstanceError
```
djkhl, line 13: this is an unused import. Do you want to add more tests here to test for this error?

djkhl (changes requested):
To my understanding everything looks good and works as expected. The only thing i saw is one unsued import. I don't know if you wanted to add more tests to check for this error.

djkhl (approved):

PR #663, opened 2024-09-10
Handle UnicodeDecodeError in kafka input

ppcad:
Prevent pipeline crash if kafka input tries to decode non-unicode character.


ekneg54 (approved):
thank you for this

PR #665, opened 2024-09-10
Remove autorulecorpustester

ekneg54:


CHANGELOG.md
```diff
@@ -2,8 +2,14 @@
...
 ## next release
 ### Breaking
+
+* remove AutoRuleCorpusTester
+
 ### Features
 ### Improvements
+
+* remove AutoRuleCorpusTester
+
```
djkhl, line 12: Are both entries neccessary?

djkhl (changes requested):

CHANGELOG.md
ekneg54: Sure... Because the one Signals the kind of the change and the other signals if the change is a breaking change

djkhl (approved):

PR #666, opened 2024-09-13
add support for python 3.13 and drop support for python 3.10

ekneg54:


ekneg54:
There is a good chance that python 3.13 is the faster option for your configuration. Measurements showed that it is the better python for a config with the http_input connector.

ekneg54:
had to fix the acceptance tests so that they will be skipped in CI on python 3.13 in cause of a strange bug that only occurs in CI on python 3.13. We can investigate that later.

dtrai2 (approved):

PR #668, opened 2024-09-16
add abiltity to configure error output

ekneg54:
This changes the error handling implementation of logprep.
This PR aims at two main goals.

1. no error should permit further processing

This is handled by not raising FatalOutputError or FatalInputError but instead raise a CriticalInputError or a CriticalOutputError.
No error is handled in the pipeline process anymore all error causing events are written to error output.
To achieve this, I had to change the batch_finished_callback mechanic. As now every event gotten from input via get_next is committed to kafka utilizing the batch_finished_callback mechanic in the pipeline.py. no connections between intput and output connectors anymore.

3. no event should get lost

to make it simple. This PR has to ensure, that every event goes into output, error output or gets logged to console as last resort.

Every error raising event will be serialized together with its raising error to an error event and is put into a multiprocessing.Queue (ThrottlingQueue). In the main Thread these events were handled in a configured error output connector which indeed can be any output connector implemented in logprep.

To achieve theses goals I had to reimplement the opensearch output connector to simplify things a lot.

Please have a look on my changes and lets discuss. Feel free to give feedback and to ask your questions. 
It is a very big PR. Sorry for that but the cut was a fundamental one.

### DIFF
--- a/logprep/framework/pipeline_manager.py
+++ b/logprep/framework/pipeline_manager.py
@@ -120,12 +120,24 @@ class ComponentQueueListener:
                 target(item)
             except Exception as error:  # pylint: disable=broad-except
                 logger.error("Error processing item: %s due to %s", item, error)
+        self._drain_queue(target)
+        component.shut_down()
+
+    def _drain_queue(self, target):
+        while not self.queue.empty():
+            item = self.queue.get()
+            if item is self.sentinel:
+                logger.debug("Got another sentinel")
+            try:
+                target(item)
+            except Exception as error:  # pylint: disable=broad-except
+                logger.error("Error processing item: %s due to %s", item, error)
+        self.queue.close()  # close queue after draining to prevent message loss
 
     def stop(self):
         """Stop the listener."""
         self.queue.put(self.sentinel)
         self._instance.join()
-        self.queue.close()
         logger.debug("Stopped listener.")
 
 
--- a/logprep/framework/pipeline_manager.py
+++ b/logprep/framework/pipeline_manager.py
@@ -204,7 +204,10 @@ class PipelineManager:
             multiprocessing.get_context(), self._configuration.error_backlog_size
         )
         self._error_listener = ComponentQueueListener(
-            self.error_queue, "store", self._configuration.error_output
+            self.error_queue,
+            "store",
+            self._configuration.error_output,
+            implementation=self._configuration.component_queue_listener_implementation,
         )
         self._error_listener.start()
         # wait for the error listener to be ready before starting the pipelines
--- a/logprep/framework/pipeline_manager.py
+++ b/logprep/framework/pipeline_manager.py
@@ -86,10 +86,26 @@ class OutputQueueListener:
         logger.debug("Starting listener with target: %s", self.target)
         self._process.start()
 
-    def get_component_instance(self):
-        component = Factory.create(self.config)
+    def get_output_instance(self):
+        """
+        Returns an instance of an output.
+
+        This method creates an instance of an output using the Factory and configures it with the setup method.
+        If an error occurs during setup, it logs an error message, puts a sentinel value into the queue, and raises
+        the error. The method waits until the queue is empty before returning the component instance. The '1' added to
+        the queue and the waiting for the queue being empty again is to ensure that the process are synchronized.
+
+        Raises:
+        -------
+        SystemExit: If an error occurs during the setup of the output component.
+
+        Returns:
+        -------
+        The configured output instance.
+        """
+        output = Factory.create(self.config)
         try:
-            component.setup()
+            output.setup()
             self.queue.put(1)
         except SystemExit as error:
             logger.error("Error output not reachable. Exiting...")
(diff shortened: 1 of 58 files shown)


ekneg54:
additional work:

- [ ] add error output to health checks
- [ ] add metric for error ouput queue
- [ ] add option in helm chart to configure error_output
- [ ] inspect behavior on high error count situations -> it looks that the error_output is dying and not restartet 
- [ ]

tests/unit/framework/test_pipeline_manager.py
```diff
@@ -314,3 +392,229 @@ def test_throttle_sleep_time_increases_with_qsize(self):
...
+        queue = ThrottlingQueue(multiprocessing.get_context(), 100)
+        queue.empty = mock.MagicMock(return_value=True)
+        listener = ComponentQueueListener(queue, target, output_config)
+        with mock.patch("logprep.connector.dummy.output.DummyOutput.setup") as mock_setup:
+            listener.queue.put("test")
+            listener.queue.put(listener.sentinel)
+            listener._listen()
+        mock_setup.assert_called()
+
+    def testget_component_instance_raises_if_setup_not_successful(self):
```
dtrai2, line 567:
```suggestion
    def test_get_component_instance_raises_if_setup_not_successful(self):
```

tests/unit/framework/test_pipeline_manager.py
```diff
@@ -314,3 +392,229 @@ def test_throttle_sleep_time_increases_with_qsize(self):
...
+        target = "store"
+        output_config = {"random_name": {"type": "dummy_output"}}
+        queue = ThrottlingQueue(multiprocessing.get_context(), 100)
+        listener = ComponentQueueListener(queue, target, output_config)
+        random_number = random.randint(10, 50)
+        listener.queue.put(listener.sentinel)
+        for _ in range(random_number):
+            listener.queue.put("test")
+        while not listener.queue.qsize() == random_number + 1:
+            pass
```
dtrai2, line 617: Could probably be removed, test putting something into the queue only after the `_listen`.

tests/unit/util/test_configuration.py
```diff
@@ -340,6 +340,7 @@ def test_verify_passes_for_valid_configuration(self):
             ),
             ("verifies input config", {"input": {"random_name": {"type": "UNKNOWN"}}}, 1),
             ("verifies output config", {"output": {"kafka_output": {"type": "UNKNOWN"}}}, 1),
+            ("verifies error_output config", {"output": {"kafka_output": {"type": "UNKNOWN"}}}, 1),
```
dtrai2, line 343:
```suggestion
```

logprep/processor/base/exceptions.py
```diff
@@ -54,18 +54,18 @@ class ProcessingError(LogprepException):
 
     def __init__(self, message: str, rule: "Rule"):
         rule.metrics.number_of_errors += 1
+        self.rule = rule
```
dtrai2, line 57: Consider removing this as it is not used anywhere

dtrai2 (changes requested):
Some more considerations:
- please check the rst configuration references for the documentation, currently the error_output is missing
- `tests/unit/charts/test_output_config.py` has a `error_index` inside the opensearch configuration, which should be removed
- consider if closing the component queue listener queue before draining (prevents adding new events from other processes)

logprep/framework/pipeline_manager.py
```diff
@@ -53,6 +57,93 @@ def put(self, obj, block=True, timeout=None, batch_size=1):
...
+                break
+            try:
+                target(item)
+            except Exception as error:  # pylint: disable=broad-except
+                logger.error("Error processing item: %s due to %s", item, error)
+        self._drain_queue(target)
+        component.shut_down()
+
+    def _drain_queue(self, target):
+        while not self.queue.empty():
```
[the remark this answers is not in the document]
dtrai2: closing the queue does indeed prevent also taking from the queue so that is not an option. A possibility would be to implement a custom queue, that inherits from the original one, and has it's own closing mechanism that allows reading from the queue, but not adding to it. I decided not to go for this solution here, as the current implementation should suffice for now.

tests/acceptance/test_error_output.py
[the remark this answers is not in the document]
dtrai2: Creating an acceptance test for the `ProcessingCriticalError` doesn't make much sense as these errors only appear if there is a bug in the code, and if we have a bug it is better to fix the bug instead of writing a test that depends on it. Yes the `GenericResolver` raises a `ProcessingCriticalError`, but that behavior is actually not correct as well, as the cases where it raises should be checked on logprep startup (checking if the configuration is valid).

Creating an acceptance test for the `OutputCriticalError` is not really doable, at least not without introducing a test that really depends on a running `kafka`/`òpensearch` output. Those are the only outputs that raises the `OutputCriticalError` and with those logprep will only start when they are healthy.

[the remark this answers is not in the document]
dtrai2: I double checked again, removing the output file of the jsonl output to trigger a `OutputCriticalError` won't work as well due to two reasons: 1) the current implementation uses `open` with `a+`, which actually just recreates the file and 2) any error that I can trigger somehow would cause an error other than `OutputCriticalError`, for example a FileNotFoundError. These are handled differently though than the `OutputCriticalError`, which would miss the point of the test in the first place.

PR #672, opened 2024-09-21
add replacer processor

### LINKED ISSUE #669: Suggestion for a new processor: Replacer
Based on Issue #[643](https://github.com/fkie-cad/Logprep/issues/643) we agreed to implement a new processor to conduct replacements in logs.

Following requirements should be met, every listed point here is open for further discussion:

**Pattern Matching**: The replacer processor should support various pattern matching techniques to identify and replace specific patterns or strings in the logs. This often involves regular expressions or predefined string patterns; or fuzzy matching in some cases.

**Customizable Replacement Rules**: It should allow users to define and customize replacement rules. This could include simple string replacements, complex regex-based substitutions, or conditional replacements based on specific log content like severity e.g. Possible replacements with hashes are open for discussion. In case of random replacements or non deterministic ones, if they are even considered; a history for possibly needed rollbacks should be kept.

**Rule Prioritization**: If wildcard rules (like test.*) and specific rules (like test.subfield) exist, there should be clear prioritization. Typically, specific rules take precedence over wildcard rules.

**Granularity in Replacement Operations**: The processor should allow users to apply replacements with different granularities, like;

- Field level: Apply to a single field like test.subfield.
- Subtree level: Apply to an entire field hierarchy like test.*.
- Global level: Apply across all fields if needed, for global transformations.

ekneg54:
closes #669 

TODO:

- [x] provide basic processor boilerplate
- [x] add basic rule design with tests
- [x] add example testcase
- [ ] provide simple implementation
- [ ] optimize implmentation
- [ ] add testcases
- [ ] add documentation
- [ ] add processor case example


ppcad:
@ekneg54 Could you please also have a look? I could not add you as reviewer, because you created this pull request.

kmeinerz (changes requested):
Please take a look at my suggestions. I mainly added some type annotations, but also there are some cases, which are not covered by tests, e.g. ```template``` is ```None```.

tests/unit/processor/replacer/test_replacer.py
```diff
@@ -0,0 +1,544 @@
+# pylint: disable=missing-docstring
```
kmeinerz, line 1:
```suggestion
# pylint: disable=missing-docstring
# pylint: disable=protected-access
```

kmeinerz (approved):
I suggest adding ```# pylint: disable=protected-access``` to test_replacer.py to handle the warning when accessing _handle_wildcard in the tests.

Everything else looks great to me—I'm happy to give my approval!

PR #677, opened 2024-09-28
Fix exporter restart 

ekneg54:
This fixes an error from previous health check implementation causing the exporter restarting on first start of logprep.

it was extracted from #668 to reduce amount of changes


ppcad (approved):
Thank you

PR #679, opened 2024-10-02
Ensure kafka output flushes queue on shutdown

ekneg54:


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 * add manual how to use local images with minikube example setup to documentation
 * move `Configuration` to top level of documentation
 * add `CONTRIBUTING` file
+* sets the default for `flush_timeout` and `send_timeout` in `kafka_output` connector to `0.1` seconds
```
ppcad, line 26: It's set to `0.0` seconds if I'm not mistaken. Is it supposed to be `0.0` in the changelog or `0.1` in the code?

ppcad (changes requested):

CHANGELOG.md
ekneg54: you are right
```suggestion
* sets the default for `flush_timeout` and `send_timeout` in `kafka_output` connector to `0` seconds
```

ppcad (approved):
Thank you!

PR #688, opened 2024-10-25
629 no support for the mysql connector python library in version 900

ppcad:
Update import for new mysql-connector-python version and unpin the version.
The mysql connector will be eventually removed, but this is required now, since trivy finds a vulnerability in the old version, which was pinned before this pull request.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,7 @@ dependencies = [
   "jsonref",
   "luqum",
   "more-itertools==8.10.0",
-  "mysql-connector-python",
+  "mysql-connector-python>=9.1.0",  #  CVE-2024-21272
   "numpy>=1.26.0",
   "opensearch-py",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -67,7 +67,7 @@ dependencies = [
   "jsonref",
   "luqum",
   "more-itertools==8.10.0",
-  "mysql-connector-python<9",
+  "mysql-connector-python",
```
dtrai2, line 70:
```suggestion
  "mysql-connector-python>=9.1.0",  #  CVE-2024-21272
```

dtrai2 (changes requested):
Okay, that was an easy fix indeed! Can you please fix the pinning to accommodate the CVE though, and also add a changelog entry? :)

dtrai2 (approved):
Thank you! :)

PR #698, opened 2024-11-07
Dev unit tests speed up

fabian-moessner:


tests/unit/connector/test_http_input.py
```diff
@@ -57,7 +60,9 @@ def setup_method(self):
         super().setup_method()
         self.object.pipeline_index = 1
         self.object.setup()
+        self.app = self.object.http_server.server.config.app
         self.target = self.object.target
+        self.falconClient = testing.TestClient(self.app)
```
ekneg54, line 65: Please call it simply client. So we have not refactor it in case of switching to another framework. Additionally is CamelCase naming not pythonic.

ekneg54 (approved):

PR #710, opened 2024-11-21
Add missing SSL config to kafka admin clients

ppcad:
The SSL config was missing in the kafka admin client, which led to errors and failed health checks if SSL was used.


ekneg54 (changes requested):
Thank you for this. Could you add tests that ensure this code won't get lost in future refactorings?

ekneg54 (approved):

PR #715, opened 2024-11-29
Fix log arrival timezone

ppcad:


logprep/abc/input.py
```diff
@@ -312,9 +318,9 @@ def _add_env_enrichment_to_event(self, event: dict):
...
     def _add_arrival_time_information_to_event(self, event: dict):
         new_field = {
-            self._config.preprocessing.get(
-                "log_arrival_time_target_field"
-            ): TimeParser.now().isoformat()
+            self._config.preprocessing.get("log_arrival_time_target_field"): TimeParser.now(
+                self._log_arrival_timestamp_timezone
+            ).isoformat()
         }
         add_fields_to(event, new_field)
```
dtrai2, line 325: maybe we can restructure this a bit, just so that it is easier to read:
```suggestion
        target = self._config.preprocessing.get("log_arrival_time_target_field")
        time = TimeParser.now(self._log_arrival_timestamp_timezone).isoformat() 
        add_fields_to(event, {target: time})
```

dtrai2 (changes requested):
Looks good, I have just a minor reformatting suggestion (see comment). Could you also please add a changelog entry though?

dtrai2 (approved):

PR #716, opened 2024-12-02
Pre-compile patterns, add cache and support ignoring case in generic resolver

ppcad:


logprep/processor/generic_resolver/processor.py
```diff
@@ -36,8 +41,77 @@
 class GenericResolver(FieldManager):
     """Resolve values in documents by referencing a mapping list."""
 
+    @define(kw_only=True)
+    class Config(Processor.Config):
+        """ListComparison config"""
```
dtrai2, line 46:
```suggestion
        """GenericResolver config"""
```

logprep/processor/generic_resolver/processor.py
```diff
@@ -36,8 +41,77 @@
...
+
+        max_cache_entries: Optional[int] = field(
+            validator=validators.optional(validators.instance_of(int)), default=0
+        )
+        """(Optional) Size of cache for results when resolving form a list."""
+        cache_metrics_interval: Optional[int] = field(
+            validator=validators.optional(validators.instance_of(int)), default=1
+        )
+        """(Optional) Cache metrics won't be updated immediately.
+        Instead it's skipped a number of times determined by cache_metrics_interval (default: 1)."""
```
dtrai2, line 56:
```suggestion
        """(Optional) Cache metrics won't be updated immediately.
        Instead updating is skipped for a number of events before it's next update. 
        :code:`cache_metrics_interval` sets the number of events between updates (default: 1)."""
```

logprep/processor/generic_resolver/processor.py
```diff
@@ -36,8 +41,77 @@
...
+        """Tracks statistics about the generic resolver"""
+
+        new_results: GaugeMetric = field(
+            factory=lambda: GaugeMetric(
+                description="Number of new values",
+                name="generic_resolver_new_results",
+            )
+        )
+        """Number of new values"""
+
```
dtrai2, line 69:
```suggestion
        """Number of newly resolved values"""
```

logprep/processor/generic_resolver/processor.py
```diff
@@ -36,8 +41,77 @@
...
+        """(Optional) Cache metrics won't be updated immediately.
+        Instead it's skipped a number of times determined by cache_metrics_interval (default: 1)."""
+
+    @define(kw_only=True)
+    class Metrics(FieldManager.Metrics):
+        """Tracks statistics about the generic resolver"""
+
+        new_results: GaugeMetric = field(
+            factory=lambda: GaugeMetric(
+                description="Number of new values",
```
dtrai2, line 64:
```suggestion
                description="Number of newly resolved values",
```

logprep/processor/generic_resolver/processor.py
```diff
@@ -36,8 +41,77 @@
...
+            factory=lambda: GaugeMetric(
+                description="Relative cache load.",
+                name="generic_resolver_cache_load",
+            )
+        )
+        """Relative cache load."""
+
+    __slots__ = ["_cache_metrics_count"]
+
+    _cache_metrics_count: int
```
dtrai2, line 94: Maybe we can rename this to something like `_cache_metrics_skip_count` to indicate that it counts the events between the metrics updates.

logprep/processor/generic_resolver/processor.py
```diff
@@ -36,8 +41,77 @@
...
     """Resolve values in documents by referencing a mapping list."""
 
+    @define(kw_only=True)
+    class Config(Processor.Config):
+        """ListComparison config"""
+
+        max_cache_entries: Optional[int] = field(
+            validator=validators.optional(validators.instance_of(int)), default=0
+        )
+        """(Optional) Size of cache for results when resolving form a list."""
```
dtrai2, line 51:
```suggestion
        """(Optional) Size of cache for results when resolving form a list.
        The cache can be disabled by setting it this option to :code:`0`"""
```

logprep/processor/generic_resolver/rule.py
```diff
@@ -98,16 +117,7 @@ class Config(FieldManagerRule.Config):
...
-            validator=[
-                validators.instance_of(dict),
-                validators.deep_mapping(
-                    key_validator=validators.instance_of(str),
-                    value_validator=validators.instance_of(str),
-                ),
-            ],
-            factory=dict,
-        )
+        resolve_list: dict = field(validator=[validators.instance_of(dict)], factory=dict)
```
dtrai2, line 120: Why did you simplify the validator?

logprep/processor/generic_resolver/rule.py
```diff
@@ -156,7 +171,25 @@ def resolve_list(self) -> dict:
...
+        ]
+
     @property
     def resolve_from_file(self) -> dict:
         """Returns the resolve file"""
         return self._config.resolve_from_file
+
+    @property
+    def ignore_case(self) -> bool:
+        """Returns if the matchin should be case-sensitive or not"""
```
dtrai2, line 189:
```suggestion
        """Returns if the matching should be case-sensitive or not"""
```

logprep/processor/generic_resolver/rule.py
```diff
@@ -156,7 +171,25 @@ def resolve_list(self) -> dict:
...
+        return self._config.ignore_case
+
+    @cached_property
+    def pattern(self) -> re.Pattern:
+        """Pattern used ot resolve from file"""
```
dtrai2, line 194:
```suggestion
        """Pattern used to resolve from file"""
```

tests/unit/processor/generic_resolver/test_generic_resolver.py
```diff
@@ -409,3 +467,163 @@ def test_resolve_generic_and_multiple_match_first_only(self):
...
+
+        assert self.object.metrics.new_results == 1
+        assert self.object.metrics.cached_results == 0
+        assert self.object.metrics.num_cache_entries == 1
+
+        self.object.process(event)
+
+        assert self.object.metrics.new_results == 2
+        assert self.object.metrics.cached_results == 1
+        assert self.object.metrics.num_cache_entries == 2
```
dtrai2, line 501: with my current understanding I would have expected that `num_cache_entries` is only 1 here, as you pass in the same event two times? also `new_results` I thought would be `1`, can you maybe explain the metrics to me again?

tests/unit/processor/generic_resolver/test_generic_resolver.py
```diff
@@ -409,3 +467,163 @@ def test_resolve_generic_and_multiple_match_first_only(self):
...
+        assert self.object.metrics.cached_results == 0
+        assert self.object.metrics.num_cache_entries == 0
+
+        self.object.process({"to_resolve": "bar"})
+
+        assert self.object.metrics.new_results == 0
+        assert self.object.metrics.cached_results == 0
+        assert self.object.metrics.num_cache_entries == 0
+
+    def test_resolve_from_cache_with_interval_2_sec(self):
```
dtrai2, line 585:
```suggestion
    def test_resolve_from_cache_with_update_interval_2(self):
```

dtrai2 (changes requested):

tests/unit/processor/generic_resolver/test_generic_resolver.py
ppcad: Those integers don't represent the actual values of the metrics, however they are calculated based on them.

The values have to be added to the metrics instead of setting them for the metrics to work:

```python
self.metrics.new_results += cache_info.misses
self.metrics.cached_results += cache_info.hits
self.metrics.num_cache_entries += cache_info.currsize
```

The integers seem to represent the sum of values that have been added.
For passing the same event twice it is:

```
new_results += 1_miss  # 0 += 1
new_results += 1_miss  # 1 += 1
=> 2
```

For passing two different events it is:

```
new_results += 1_miss    # 0 += 1
new_results += 2_misses  # 1 += 2
=> 3
```

logprep/processor/generic_resolver/rule.py
ppcad: I wanted the ability to add non-string values like dictionaries.
The only thing preventing that is the validator.

dtrai2: ah okay, I didn't find that anywhere that's why I was wondering. Do you want to add a small test for this then, just so we fix this behavior and nobody will change it again in the future.

tests/unit/processor/generic_resolver/test_generic_resolver.py
dtrai2: ahh wait, sorry now I understand. In the tests we mock the `GaugeMetric` again with a regular `int` that's why the value continues increasing. I missed that. Could you then maybe add a comment to the test describing that the difference between the metric values (before and after each `self.object.process` call) is actually the relevant information because of the "mock"?

logprep/processor/generic_resolver/rule.py
ppcad: Sure, I have added a test.

tests/unit/processor/generic_resolver/test_generic_resolver.py
ppcad: I have added docstrings to the relevant tests.

dtrai2 (approved):
thanks for the implemented suggestions! 👍

PR #719, opened 2024-12-05
Fix filter matching with bypass tree

ppcad:


logprep/filter/expression/filter_expression.py
```diff
@@ -87,7 +87,10 @@ def _get_value(key: List[str], document: dict) -> Any:
         for item in key:
             if item not in current:
                 raise KeyDoesNotExistError
-            current = current[item]
+            try:
+                current = current[item]
+            except TypeError as error:
+                raise KeyDoesNotExistError from error
```
ekneg54, line 93: I have a bad feeling with that. This only solves the exception not the reason. We should dig deeper to change the code in a more sustainable way to avoid future bugs like this. We should investigate, why this error happens instead of sticking plasters on bleeding wounds.

ekneg54 (changes requested):
Could you investigate a litte bit deeper?

logprep/filter/expression/filter_expression.py
ppcad: The error comes from parsing a dict and not stopping the the result was found.
Instead of checking for it being dict I used try/except, since it's faster and this part of the code is used a lot.
This method is used for matching. As far as I understand this is the root of the problem.
It does not happen with the rule tree, since it matches differently.

ekneg54: Yes it is passed often but we have no chance to refactor it to better performance if nobody knows what this code does. Please write with focus on readability not performance. we can use more hardware for more performance or get better ideas if the code is readable or rewrite something in another language. But the starting point for all of this is clear and readable code.

ppcad: I have changed it to use `isinstance`.

ekneg54: Thank you very much

ekneg54 (approved):

ppcad:
I also want to clarify why this problem doesn't occur with the tree.
The tree does first check with an `Exists` filter expression for existence before it applies a `KeyValueBasedFilterExpression`.
The `Exists` filter expression uses a different method for matching.

PR #734, opened 2024-12-19
Fix bug lucene compliance

fabian-moessner:


ekneg54 (changes requested):
please take a look at the jupyther nodebook and provide examples that met the following acceptance criterias:

* strict Englisch
* a user understand the magic on the first look. for this the user has to see the rule, the processor and the input and expected events directly in the notebook
* clean up all not needed examples

ekneg54 (approved):

PR #735, opened 2024-12-19
add scheduled workflow for weekly image build

dtrai2:
- adds a new workflow that is scheduled to run once a week
- create the latest released image again with a new tag using the suffix "weekly"
- that way new CVE fixes from the underlying bitnami baseimage are automatically integrated


dtrai2:
I close this attempt again as we need a more structured way with which we can ensure CVE free images, for example with updates to older releases. One way could be with release branches, according to: https://trunkbaseddevelopment.com/branch-for-release/

PR #740, opened 2025-01-13
fix multiarch container build

ekneg54:


djkhl (approved):
thanks for the quick fix!

PR #747, opened 2025-01-28
Remove colorama depedency

fabian-moessner:


ekneg54:
the task was to eliminate the colorama dependency. there is no need to reinvent the wheel. Consider to copy the code for the import and its dependencies to a colorama module in logprep. Then you should only fix the import. Please ensure to copy their license too.

fabian-moessner:
When running locally `black --verbose --config ./pyproject.toml logprep/` all files would be left unchanged. So I don't understand the results of quality  checks at the moment. The criticized files came from a rebase.

ekneg54 (approved):

ekneg54:
and fixed black. they released a new version. we should reinstall the dev dependencies.

PR #748, opened 2025-01-29
Fix push mirror

ppcad:


ppcad:
I'm not sure why some pipelines fail. I did only change the workflow for the push mirror.

ekneg54:
btw. I have no clue, why black is complaining

PR #750, opened 2025-01-30
Adding test for a valid config update after an invalid config

fabian-moessner:


tests/acceptance/test_config_refresh.py
```diff
@@ -73,3 +74,21 @@ def test_no_config_refresh_after_5_seconds(tmp_path, config):
...
+
+
+def test_recover_after_invalid_then_valid_config(tmp_path, config):
+    config_path = tmp_path / "generated_config.yml"
+    config_path.write_text(config.as_json())
+    proc = start_logprep(config_path)
+    wait_for_output(proc, "Config refresh interval is set to: 5 seconds", test_timeout=5)
+
+    config.config_refresh_interval = None
+    config.version = "invalid"
```
ekneg54, line 86: as described in https://github.com/fkie-cad/Logprep/blob/00688049ef280e41bd4ece256eddcf70d459b7bc/logprep/util/configuration.py#L434-L435 the version  attribute is of type "str". So setting it to the string "invalid" doesn't result in an invalid configuration.

tests/acceptance/test_config_refresh.py
```diff
@@ -73,3 +74,21 @@ def test_no_config_refresh_after_5_seconds(tmp_path, config):
...
+    config_path.write_text(config.as_json())
+    time.sleep(3)
+
+    config.config_refresh_interval = 4
+    config.version = "valid_again"
+    config_path.write_text(config.as_json())
+    wait_for_output(proc, "Successfully reloaded configuration", test_timeout=10)
+    wait_for_output(proc, "Config refresh interval is set to: 4 seconds", test_timeout=5)
```
ekneg54, line 94: as defined here https://github.com/fkie-cad/Logprep/blob/00688049ef280e41bd4ece256eddcf70d459b7bc/logprep/runner.py#L120-L129 the minimum for a config refresh interval should be 5 Seconds. If you are able to set it to 4 Seconds we might have a new bug.

tests/acceptance/test_config_refresh.py
```diff
@@ -73,3 +74,21 @@ def test_no_config_refresh_after_5_seconds(tmp_path, config):
...
+def test_recover_after_invalid_then_valid_config(tmp_path, config):
+    config_path = tmp_path / "generated_config.yml"
+    config_path.write_text(config.as_json())
+    proc = start_logprep(config_path)
+    wait_for_output(proc, "Config refresh interval is set to: 5 seconds", test_timeout=5)
+
+    config.config_refresh_interval = None
+    config.version = "invalid"
+    config_path.write_text(config.as_json())
+    time.sleep(3)
```
ekneg54, line 88: sleep actions in test leads to flaking tests because you never know if a test run on a slower system does not needs some seconds more. It is better to wait for some condition in the logs for example.

ekneg54 (changes requested):
This does not reproduces the failure. Please have a look deeper. 
unittests for the refresh interval can be found in `tests/unit/test_runner.py`. Consider starting there to go deeper into testing and bug reproduction.

fabian-moessner:
Was already tested in another test. Thus, withdrawing pull request.

PR #764, opened 2025-02-20
Fix thread and threadpool

kmeinerz:
Fixed the ThreadPoolExecutor in _generate_load() and made the generator in batcher thread safe.
Testing showed , that the desired number of threads were opened and all were closed later.


ekneg54:
merged your changes to mine. thx

PR #770, opened 2025-02-25
add ability to write the full event to one singular target field

djkhl:
This PR schould make it possible for logprep to add a whole event as a string to a designated target field.

### DIFF
--- a/tests/unit/connector/test_http_input.py
+++ b/tests/unit/connector/test_http_input.py
@@ -268,6 +268,21 @@ class TestHttpConnector(BaseInputTestCase):
         assert re.search(r"\d+\.\d+\.\d+\.\d+", message["custom"]["remote_addr"])
         assert isinstance(message["custom"]["user_agent"], str)
 
+    def test_event_original(self):
+        message = {"message": "my message"}
+        updated_config = {"event_origin_field": "event.original"}
+        connector_config = deepcopy(self.CONFIG)
+        connector_config.update(updated_config)
+        connector = Factory.create({"test connector": connector_config})
+        connector.pipeline_index = 1
+        connector.setup()
+        client = testing.TestClient(connector.app)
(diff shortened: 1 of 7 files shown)


ekneg54:
add a simple skeleton how to get this working. please have a look.

tests/unit/connector/test_http_input.py
ekneg54, line 1: please add testcases for interaction between your preprocessor an the stringificaiton in this processor.

What should happen if both are enabled? 
What if version_info is written?
What if logarival timestamp is written?
What if the metadata of this processor is written. Should it be written before or after your feature?
should it be in your stringified data or should add after your preprocessor

ekneg54 (changes requested):
Thank you for your contribution. 
Please have a look at my requested changes.

tests/unit/connector/test_http_input.py
djkhl: in my opinion every other added pre processor should come after the addition of the event. I think the event should be written in its "raw" form with as little added to it as possible.

ekneg54:
something changed on the runner and so acceptance tests fail because running processes where not killed. so ports are in usage and the next test is not able to spin up next logprep process.

has to wait for #774

ekneg54:
@djkhl could you please have a look with this code in our integration setup how the logs will look like?

djkhl:
With the suggestion from @ekneg54 ist works fine in our integration setup and every document is sent to opensearch. 
grafik
@gigallucci it seems that opensearch is able to parse the log in the given string format.

ekneg54:
if we use it after kafka, the event looks as follows:

image


perhaps the kafka output serialisation reduces the escaping.

djkhl:
image
event before preprocessor is used

djkhl:
with preprocessor and str format active 
image

PR #777, opened 2025-03-13
Add refresh token to client credentials

djkhl:


djkhl:
Wrong approach to fix the problem.

PR #790, opened 2025-03-28
Prevent restart timeout from increasing endlessly

ppcad:


ekneg54 (changes requested):
Thank you. But please implemented at minimum one test to prevent regression and add an entry in the changelog.

ekneg54 (approved):

PR #794, opened 2025-04-06
Log record returns error

kmeinerz:


kmeinerz:
Something went wrong with rebasing and it showed 43 changed instead of 3. The simplest solution was to just reopen a new branch and PR.

PR #798, opened 2025-04-09
Fix test config endless load

kmeinerz:
The test config had a logging queue, but no QueueListener.


ppcad:
Thank you, this fixes the issue!
I've noticed the auto rule tester itself has this issue as well.
I will make another issue for that, since there is one more thing that can prevent the auto rule tester from stopping.

ppcad (approved):

PR #808, opened 2025-04-24
Fix domain resolver error, add resolve status and add tests

ppcad:


logprep/processor/domain_resolver/processor.py
```diff
@@ -50,6 +51,13 @@
 logger = logging.getLogger("DomainResolver")
 
 
+class ResolveStatus(Enum):
```
ekneg54, line 54: Consider using an `IntEnum` here as you use `ResolveStatus.SUCCESS` in return of methods, which evaluates now to a EnumType. By using `IntEnum` it resolves to an primitive integer, which should use less memory I guess. 
Please provide a docstring for this class.

logprep/processor/domain_resolver/processor.py
```diff
@@ -158,35 +180,49 @@ def _apply_rules(self, event, rule):
...
+        self._add_resolve_infos_to_event(event, rule, resolved_ip)
+
+        if self._config.debug_cache:
+            self._store_debug_infos(event, requires_storing)
+
     def _add_resolve_infos_to_event(self, event, rule, resolved_ip):
         if resolved_ip:
             self._write_target_field(event, rule, resolved_ip)
 
     def _resolve_ip(self, domain, hash_string=None):
```
ekneg54, line 208:
```suggestion
    def _resolve_ip(self, domain):
```

`hash_string` is not used anymore. Is this right?

logprep/processor/domain_resolver/processor.py
```diff
@@ -158,35 +180,49 @@ def _apply_rules(self, event, rule):
...
             resolved_ip = result.get(timeout=self._config.timeout)
-            return resolved_ip
-        except (context.TimeoutError, OSError):
-            if hash_string:
-                self._domain_ip_map[hash_string] = None
+            return resolved_ip, ResolveStatus.SUCCESS
+        except ValueError:  # Makes no connection so does not need to be cached
+            self.metrics.invalid_domains += 1
+            return None, ResolveStatus.INVALID
+        except context.TimeoutError:
```
ekneg54, line 220:
```suggestion
        except TimeoutError:
```

Why do you use the TimeoutError from "context" ??

ekneg54 (changes requested):
Please add a changelog entry and have a look at my remarks.

logprep/processor/domain_resolver/processor.py
ppcad: It's a different error despite having the same name. It's thrown by the context manager and the regular TimoutError can't be caught in its place.

logprep/processor/domain_resolver/processor.py
ppcad: Yes, thank you.

ekneg54:
rebased from main

ekneg54 (approved):

PR #811, opened 2025-05-06
Fix kafka input

kmeinerz:


ekneg54 (approved):
had to fix the revoke_callback too

Issue #757, opened 2025-02-05
Logprep config validation runs endlessly if many logmessages are being sent


ppcad:
Logprep `test config` runs endlessly after being finished if many log messages have been sent.
One way to reproduce this is by having many Logprep rules that use the deprecated lucene syntax.
This will create log messages during loading the config.

This problem does likely extend to all Logprep features that load the config and want to close Logprep after they are finished.

This could be circumvented by printing less deprecation messages or disabling the logging while verifying the config, but this would be just a work-around.

PR #814, opened 2025-05-13
Fix failing github checks

kmeinerz:


kmeinerz:
I replaced the print output to proper logging in the get_configuration function.
I also tried to remove the queue listener, as it seems unnecessary for a single line of code, but this would reintroduce the bug https://github.com/fkie-cad/Logprep/issues/757 , as it does also handle the logging of the warning produced in `config = Configuration.from_sources(config_paths)`.
I do think that there is a better way to handle the logging, but I suggest to open a separate issue or PR for that.

ekneg54 (approved):
good job

PR #815, opened 2025-05-13
Fix datetime deprecation warnings

ppcad:


ekneg54 (changes requested):
Thank you for these changes. But there is still one Deprecation Warning for this topic from botocore by running the tests. 

botocore is only used by the S3 Output I guess. I heard you don`t use it anymore. Could we remove the "s3_output" module and tests to decrease dependency footprint on one hand and get rid of this last deprecation warning on the other hand?

I guess we get rid of the "boto3" and "protobuf" dependency, aren't we?

ppcad:
> Thank you for these changes. But there is still one Deprecation Warning for this topic from botocore by running the tests.
> 
> botocore is only used by the S3 Output I guess. I heard you don`t use it anymore. Could we remove the "s3_output" module and tests to decrease dependency footprint on one hand and get rid of this last deprecation warning on the other hand?
> 
> I guess we get rid of the "boto3" and "protobuf" dependency, aren't we?

I've only wanted to fix the ones caused by our own code. I should have specified that. However, we can get rid of botocore or silence the warnings for now by adding this to run_logprep.py: `warnings.filterwarnings("default", category=DeprecationWarning, module="botocore")` (or `ignore`) to prevent it from spamming the console.

Should we do another pull request for the removal of botocore?

ekneg54:
Let`s merge this and consider to remove s3_output with its dependency in the near future before architectural changes will take place, to avoid unnecessary tasks.

ekneg54 (approved):

ppcad:
That sounds good to me.

PR #817, opened 2025-05-16
Make custom yaml tag loader work with large files

ppcad:
Custom YAML anchor tags did not work with large files if the anchors were set near the end of a file, because it relied on a buffer that could change.

This fixes it by no longer relying on a buffer to track changes and load tag values.

Mainly `_extract_anchor_value` and `clear_anchors_if_buffer_changed ` (now `clear_anchors_if_stream_changed`) where changed.

`_extract_anchor_value` parses the node value instead of using the read buffer.

`clear_anchors_if_stream_changed` now checks if the current line number of the yaml reader is smaller than the last one, which would mean that a new stream/file has been opened. Furthermore, checks are now performed by an implicit tag that is processed for every node.

### DIFF
--- a/logprep/util/tag_yaml_loader.py
+++ b/logprep/util/tag_yaml_loader.py
@@ -75,7 +75,8 @@ Examples:
 """
 
 import os.path
-from typing import Set, Callable, Any
+import re
+from typing import Callable, Any
 
 from ruamel.yaml import (
     YAML,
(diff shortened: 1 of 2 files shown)


ppcad:
It no longer clears anchor tags for new files.
I did this because the implementation for it changed the formatting for our custom MyYAML loader in some helm chart unit tests when running all tests (succeeds if tests are run seperately). Mainly, it does not keep quotes and results in a different ordering for those tests.

The amount of anchor tags is still limited to ten.

PR #819, opened 2025-05-21
Fix timezone for getting current time via TimeParser

ppcad:
`TimeParser.now()` created a timestamp with the local timezone and then changed the timezone to UTC if no timezone was specified.
With this pull request a UTC timestamp will be created instead.

Example:
* Let's say the local timezone is `+02:00` and no timezone was provided to `TimeParser.now()`
* `TimeParser` would create a timestamp `2025-01-01T02:00:00.000` in local time, but without a timezone
* The timezone would be set to UTC without adapting the time values, resulting in `2025-01-01T02:00:00.000+00:00`
* However, `2025-01-01T00:00:00.000+00:00` would be correct

### DIFF
--- a/logprep/util/time.py
+++ b/logprep/util/time.py
@@ -56,12 +56,12 @@ class TimeParser:
         return time_object
 
     @classmethod
-    def now(cls, timezone: tzinfo = None) -> datetime:
+    def now(cls, timezone: tzinfo | None = UTC) -> datetime:
         """returns the current time
 
         Parameters
         ----------
-        timezone : tzinfo
+        timezone : tzinfo | None
             the timezone to use for the timestamp
 
         Returns
@@ -69,8 +69,8 @@ class TimeParser:
         datetime
             current date and time as datetime
         """
+        timezone = timezone if timezone else UTC
         time_object = datetime.now(timezone)
-        time_object = cls._set_utc_if_timezone_is_missing(time_object)
         return time_object
 
     @classmethod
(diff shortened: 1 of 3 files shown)


logprep/util/time.py
```diff
@@ -56,21 +56,21 @@ def from_timestamp(cls, timestamp: Union[int, float]) -> datetime:
...
 
     @classmethod
-    def now(cls, timezone: tzinfo = None) -> datetime:
+    def now(cls, timezone: tzinfo | None = UTC) -> datetime:
         """returns the current time
 
         Parameters
         ----------
-        timezone : tzinfo
+        timezone : tzinfo | None
```
kmeinerz, line 64:
```suggestion
        timezone : tzinfo
```

kmeinerz (changes requested):
Thank you for your contribution! I've reviewed the suggested changes.

After updating the default value of the timezone parameter, I don't see a case where the now() function would receive ```timezone=None```. The suggested changes align with this reasoning.

Please let me know if I’ve overlooked any potential use cases.

logprep/util/time.py
ppcad: In future some more timezones might come from configuration files or rules. I wanted to make sure that there is always a timezone even if some configuration values are accidentally `None` or `""`. Ideally the configuration value would be checked, but I just don't want want timezone problems to slip in again, since they can be hard to notice.

However, I can change it if you want. Then we just have to keep it in mind if we work with timeszones.

kmeinerz: Thank you for specifying your intention. Looking back at your initial changes with that context, I agree that handling the ```None``` case by defaulting to ```UTC``` is the right approach, which can prevent us some headache in in the long run.

kmeinerz (approved):

PR #821, opened 2025-05-26
Fix-generator-multithreading

kmeinerz:


kmeinerz:
Note on the implementation:
The following lines were added to satisfy mypy code quality requirements:
```
    @define(kw_only=True)
    class Metrics(Connector.Metrics):
        """Tracks statistics about this connector"""
```
In ```logprep.connector.http.output.py```, we define:
```
class Metrics(Output.Metrics):
```
However, Output previously did not explicitly define a Metrics class, which led to type checking issues.

ekneg54 (approved):
thanks for clearing this and I agree that testing this would not be worth the needed workarounds.

PR #824, opened 2025-06-10
Dev eventstate implementation

kaya-david:
* implement EventState class to manage the lifecycle of log events
* integrate a finite state machine to control valid state transitions
* support for both automatic and conditional (success-based) state transitions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@
 ### Improvements
 * implement EventState class to manage the lifecycle of log events
 * integrate a finite state machine to control valid state transitions
-* support for both automatic and conditional (success-based) state transitions
 
 * add ng packages as namespace in dirs 'unit' and 'logprep' as preparation for new architecture implementation
 * add abstract EventMetadata class and KafkaInputMetadata class
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,12 @@
 ### Breaking
 
 ### Features
+
 ### Improvements
+* implement EventState class to manage the lifecycle of log events
+* integrate a finite state machine to control valid state transitions
+* support for both automatic and conditional (success-based) state transitions
+
 ### Bugfix
 
 ## 17.0.0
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 ### Features
 
 ### Improvements
+* implement EventState class to manage the lifecycle of log events
+* integrate a finite state machine to control valid state transitions
+* support for both automatic and conditional (success-based) state transitions
```
kmeinerz, line 11: I would remove the last point, as it is too technical. The important part was covered in the previous point.

logprep/ng/event_state.py
```diff
@@ -0,0 +1,205 @@
...
+
+    _state_machine = None  # Will be initialized lazily
+    """Class-level state transition map, initialized once and shared across
+    all instances."""
+
+    def __init__(self) -> None:
+        """Initialize the event state with the default starting state."""
+        if EventState._state_machine is None:
+            EventState._state_machine = EventState._construct_state_machine()
+        self.current_state: EventStateType = EventStateType.RECEIVING
```
kmeinerz, line 95: As we are not keeping track of the state history, the variable could also be renamed to state,

kmeinerz (changes requested):
Please look at my comments

logprep/ng/event_state.py
```diff
@@ -0,0 +1,208 @@
...
+            chosen = self._resolve_by_success_flag(next_states, success)
+
+            if chosen:
+                self.current_state = chosen
+                return self.current_state
+
+        raise ValueError("Invalid state transition.")
+
+    @classmethod
+    def _resolve_by_success_flag(cls, options: list[str], success: bool) -> str | None:
```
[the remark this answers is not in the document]
kaya-david: The built-in next() function inside _resolve_by_success_flag returns None instead of raising a StopIteration exception, due to the provided default value. That’s why _resolve_by_success_flag can return None when no match is found.

To avoid confusion with Python’s built-in next() function, I renamed the method in the class to next_state().

PR #823, opened 2025-06-10
Dev eventmetadata class implementation

kaya-david:
*add abstract event metadata class
*update CHANGELOG.md
*implement concrete kafka metaclass
*add unittests for kafka metaclass

PR #825, opened 2025-06-12
Dev event class implementation

kaya-david:
* implement abstract Event class to encapsulate event data, processing state, warnings, and errors
* integrate dotted field handling methods directly into Event, enabling structured field access and manipulation
* support event identity and hashability based on data, allowing usage in sets and as dictionary keys

### DIFF
--- a/tests/unit/ng/test_event.py
+++ b/tests/unit/ng/test_event.py
@@ -18,14 +18,6 @@ class DummyEvent(Event):
     __slots__ = Event.__slots__
 
 
-class DummyRule:
-    def __init__(self):
-        self.id = "dummy_rule"
-        self.description = "Dummy rule description"
-        self.source_fields = []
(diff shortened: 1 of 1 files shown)


kaya-david:
ON HOLD

rebase to main required after merging open PRs: 
- https://github.com/fkie-cad/Logprep/pull/824
- https://github.com/fkie-cad/Logprep/pull/823

kmeinerz:
I've removed some parts from this PR and opted to import them instead. Specifically, the following functions are now imported:
    ```add_fields_to```, ```get_dotted_field_value```, ```pop_dotted_field_value```
Supporting functions that were only used by these have been completely removed.

Although the code in the new namespace shouldn't affect other parts of the repository, I believe we should continue importing shared functionality from the existing codebase, long as we're not modifying it. Otherwise, we risk duplicating functionality unnecessarily or will eventually need to reconcile similar functions later.

kmeinerz (approved):

PR #833, opened 2025-06-30
use new logevent class instead of plain dict

ekneg54:


logprep/ng/abc/processor.py
```diff
@@ -208,29 +152,29 @@ def _process_rule_tree_once(tree: RuleTree, event: dict):
...
         except ProcessingCriticalError as error:
-            if self.result is None:
+            if self._event is None:
                 raise error
-            self.result.errors.append(error)  # is needed to prevent wrapping it in itself
+            self._event.errors.append(error)  # is needed to prevent wrapping it in itself
             event.clear()
         except Exception as error:  # pylint: disable=broad-except
-            if self.result is None:
+            if self._event is None:
```
kaya-david, line 166:
```suggestion
            if self._event is None or not self._event.data:
```

Maybe check also, that events data is not empty?

kaya-david (commented):
Please check out my comments

logprep/ng/abc/processor.py
ekneg54: according to the existing implementation this is not needed or perhaps not the intended behavior

logprep/ng/abc/processor.py
```diff
@@ -105,29 +52,26 @@ class Config(Component.Config):
     __slots__ = [
         "_event",
         "_rule_tree",
-        "result",
         "_bypass_rule_tree",
     ]
 
     rule_class: ClassVar["Type[Rule] | None"] = None
```
kmeinerz, line 58:
```suggestion
    rule_class: ClassVar[Type["Rule"] | None] = None
```
I suggest writing the typing with only Rule as string. This way it removes the error unused import warning from the Type import.

kmeinerz (changes requested):
Thank you for your work! Please look at my comment.

I have one small style suggestion for the ```processor.py``` file. When adding type hints, it would be more consistent with the rest of our code to use the built-in ```dict``` and ```list``` generics instead of ```Dict``` and ```List``` from typing.

kmeinerz (approved):

PR #843, opened 2025-07-04
make predetector use new logevents

ekneg54:


logprep/ng/event/sre_event.py
```diff
@@ -1,7 +1,28 @@
...
 from logprep.ng.abc.event import Event
+from logprep.ng.event.event_state import EventState
 
 
 class SreEvent(Event):
     """Concrete Sre event class."""
+
+    __slots__ = ("outputs",)
+
+    outputs: tuple[dict[str, str]]
```
kaya-david, line 12: Would it perhaps make more sense to define output as a class variable in the abstract base class Event?

kaya-david (commented):
Thanks for the effort! Looks good to me overall. Please check my comments.

kaya-david (approved):

PR #844, opened 2025-07-04
Make datetime extractor use LogEvents

kmeinerz:


kaya-david:
I noticed that some tests are failing on my local machine. Could you please check whether they are running successfully on your side?

kaya-david (approved):
Seems good to me.

PR #848, opened 2025-07-09
Dev use logevent pseudonymizer

ekneg54:


logprep/ng/abc/event.py
```diff
@@ -209,6 +209,35 @@ def pop_dotted_field_value(self, dotted_field: str) -> Optional[Union[dict, list
         return pop_dotted_field_value(self.data, dotted_field)
 
 
+class ExtraDataEvent(Event):
```
kaya-david, line 212: For me this class makes it too complicated. Would suggest to move output to event base class for now.

Do you currently see any urgency in introducing another level of abstraction? SreEvent also requires output, so as suggested in another PR, it might make sense to me to define output as an attribute in the abstract Event class.

kaya-david (commented):
I suggest reviewing this again once we’ve resolved the open points in PR [#843](https://github.com/fkie-cad/Logprep/pull/843) regarding the output attribute. Depending on the outcome, the additional class I commented on in this PR might no longer be necessary.

If the structure of PR [#843](https://github.com/fkie-cad/Logprep/pull/843) changes, please rebase this branch onto main, and I’ll be happy to take another detailed look at the latest version.

PR #862, opened 2025-07-25
Fix crash on TimeParserException during preprocessing

ppcad:


kaya-david (approved):
Thank you for this fix! @ppcad

PR #863, opened 2025-07-29
Fix auto-rule tester getting stuck due to logging

ppcad:
* Fix stuck auto-rule tester
* Fix failing domain resolver tests


ekneg54 (approved):
the logqueue listener will be deleted in the current rewrite of logprep. So I think this change were resolved by simply letting time pass.

PR #864, opened 2025-07-29
add basic pipeline

ekneg54:


logprep/ng/pipeline.py
```diff
@@ -0,0 +1,67 @@
...
+        """Iterate over processed events."""
+        while True:
+            events = (event for event in self._input if event is not None and event.data)
+            batch = list(islice(events, 10))
+            if not batch:
+                break
+            yield from map(self._process_event, batch)
+
+    def __next__(self):
+        return self._process_event(next(self._input))
```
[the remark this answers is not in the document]
ekneg54: done

logprep/ng/pipeline.py
```diff
@@ -0,0 +1,76 @@
...
+            events = (event for event in self._input if event is not None and event.data)
+            batch = list(islice(events, self._process_count))
+            if not batch:
+                break
+            yield from map(self._process_event, batch)
+
+    def __next__(self):
+        """Get the next processed event."""
+        try:
+            while (next_event := next(self._input)) is None or not next_event.data:
```
kaya-david, line 59: please implement helper method:


```
def _is_valid_event(self, event: Event) -> bool:
    return event is not None and bool(event.data)
```
and adapt this line:
```suggestion
            while not self._is_valid_event(next_event := next(self._input)):
```

so you can easily re-use the helper function in:
```
# in __iter__()
events = (event for event in self._input if self._is_valid_event(event))
```

**Reason**:
Both __iter__() and __next__() are implemented and intentionally used separately, which is fine. However, the event filtering logic is duplicated in both methods. Consider extracting it into a private helper (e.g., `_is_valid_event`) to keep behavior consistent and avoid divergence in future changes.

logprep/ng/pipeline.py
```diff
@@ -0,0 +1,76 @@
...
+        self, input_connector: Iterator[Event], processors: list[Processor], process_count: int = 10
+    ) -> None:
+        self._input = input_connector
+        self._processors = processors
+        self._process_count = process_count
+
+    def __iter__(self) -> Generator[LogEvent, None, None]:
+        """Iterate over processed events."""
+        while True:
+            events = (event for event in self._input if event is not None and event.data)
```
kaya-david, line 50: Implement helper method mentioned in next comment and use:

```suggestion
            events = (event for event in self._input if self._is_valid_event(event))
```

logprep/ng/pipeline.py
```diff
@@ -0,0 +1,76 @@
...
+        >>> len(processed_events)
+        2
+        >>> processed_events[0].data["processed"]
+        True
+        >>> processed_events[1].data["message"]
+        'test2'
+    """
+
+    def __init__(
+        self, input_connector: Iterator[Event], processors: list[Processor], process_count: int = 10
```
kaya-david, line 41: Please use LogEvent as annotation to match with mehtod `_process_event` return type or vice versa.

```suggestion
        self, input_connector: Iterator[LogEvent], processors: list[Processor], process_count: int = 10
```

logprep/ng/pipeline.py
```diff
@@ -0,0 +1,76 @@
...
+        """Iterate over processed events."""
+        while True:
+            events = (event for event in self._input if event is not None and event.data)
+            batch = list(islice(events, self._process_count))
+            if not batch:
+                break
+            yield from map(self._process_event, batch)
+
+    def __next__(self):
+        """Get the next processed event."""
```
kaya-david, line 57: Perhaps use a docstring like this to explain why the Iterator protocol inherited by this class is intentionally broken:

```suggestion
        """
       Return the next processed event or None if no valid event is found.

       This method intentionally deviates from the standard Python iterator protocol.
       Normally, __next__() must raise StopIteration to signal that there are no more
       items. In this Pipeline, __next__() instead returns None when no valid event
       is available. This design allows callers to check for "no event" without
       handling StopIteration explicitly, but means the method is not strictly
       iterator-compliant by design.
       """
```

kaya-david (commented):
Nice job! Please check out my comments. :)

logprep/ng/pipeline.py
ekneg54: in this case I think it is not worth it do hide this expression behind a helper method, that is not descriptive enough to explain when a return value is a valid event. additionally you can't do this in this case, because the walrus operator won`t be evaluated and you have to add additional variable assignments. 

from performance perspective it is a bad idea too, because you wrap the expression in two function calls. one for getting the method and one for the method itself.

logprep/ng/pipeline.py
ekneg54: This makes an unnecessary decision for the future. It is possible to put all `Events` here. We should not decide this for our future self ;)

logprep/ng/pipeline.py
```diff
@@ -0,0 +1,85 @@
...
+        iterator-compliant by design.
+        """
+        try:
+            while (next_event := next(self._input)) is None or not next_event.data:
+                continue
+        except StopIteration:
+            return None
+        return self._process_event(next_event)
+
+    def _process_event(self, event: LogEvent) -> LogEvent:
```
kaya-david, line 74:
```suggestion
    def _process_event(self, event: Event) -> Event:
```

PR #867, opened 2025-08-03
fix crash on TimeParserException during preprocessing caused by inval…

kaya-david:
* Fix crash on TimeParserException during preprocessing caused by invalid timestamps

### DIFF
--- a/tests/unit/ng/connector/base.py
+++ b/tests/unit/ng/connector/base.py
@@ -890,6 +890,46 @@ class BaseInputTestCase(BaseConnectorTestCase):
 
             connector.shut_down()
 
+    @pytest.mark.parametrize(
+        ["timestamp", "expected_error_message"],
+        [
+            ("0000-00-00 00:00:00", "year 0 is out of range"),
+            ("invalid", "Invalid isoformat string: 'invalid'"),
+        ],
+    )
+    def test_pipeline_preprocessing_raises_criticalinputerror_on_timeparser_error(
+        self, timestamp, expected_error_message
+    ):
+        test_event = {"any": "content", "@timestamp": timestamp}
+
+        with self.patch_documents_property(document=test_event):
+            preprocessing_config = {
+                "preprocessing": {
+                    "log_arrival_time_target_field": "arrival_time",
+                    "log_arrival_timedelta": {
+                        "target_field": "log_arrival_timedelta",
+                        "reference_field": "@timestamp",
+                    },
+                }
+            }
+            connector_config = deepcopy(self.CONFIG)
+            connector_config.update(preprocessing_config)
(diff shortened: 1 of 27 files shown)


kaya-david:
Bug from old architecture: https://github.com/fkie-cad/Logprep/pull/862

Fix adopted into the new architecture

tests/unit/ng/connector/base.py
```diff
@@ -890,6 +890,46 @@ def test_pipeline_preprocessing_does_not_add_timestamp_delta_if_configured_but_l
...
             connector.shut_down()
 
+    @pytest.mark.parametrize(
+        ["timestamp", "expected_error_message"],
+        [
+            ("0000-00-00 00:00:00", "year 0 is out of range"),
+            ("invalid", "Invalid isoformat string: 'invalid'"),
+        ],
+    )
+    def test_pipeline_preprocessing_raises_criticalinputerror_on_timeparser_error(
```
ekneg54, line 900: the test suggests that a criticalinputerror should be raised but I see no pytest.raises. consider to find a better name for this test.

tests/unit/ng/connector/base.py
```diff
@@ -890,6 +890,46 @@ def test_pipeline_preprocessing_does_not_add_timestamp_delta_if_configured_but_l
...
+        self, timestamp, expected_error_message
+    ):
+        test_event = {"any": "content", "@timestamp": timestamp}
+
+        with self.patch_documents_property(document=test_event):
```
ekneg54, line 905: I do not like this context_manager, because it is located in a shared base class for a specialized subclass. consider to overwrite the failing tests in the subclass instead and to write the tests in the base class for all sub testclasses.

ekneg54 (changes requested):

tests/unit/ng/connector/base.py
kaya-david: What do you mean by "…for a specialized subclass"?
This mock is also used and required in many tests within the base class itself.
Could you please clarify your idea in more detail?

ekneg54: this patch is only for the json_input connector tests. as now, this is used for all input connectors. but they don't have a documents_property

kaya-david: Thank you, understood. That makes sense. I will move the patch into the specific subclass and overwrite the failing tests there instead.

kaya-david:
Don't let the large number of file changes discourage you, I had to fix several pylint issues by adding pylint: disable comments at the top of many test files.

tests/unit/ng/connector/base.py
kaya-david: Done.

PR #882, opened 2025-09-23
Refreshable http getters

ppcad:


logprep/util/configuration.py
```diff
@@ -629,10 +629,16 @@ class Configuration:
...
+    changed: bool = field(default=False, repr=False, eq=False, init=False)
+
+    _current_variable_values: list = field(default=[], repr=False, eq=False, init=False)
+
     _unserializable_fields = (
         "_getter",
         "_configs",
         "_config_failure",
+        "changed",
+        "_current_variable_values",
```
kmeinerz: See previous comment

logprep/runner.py
```diff
@@ -119,9 +119,10 @@ def _iterate(self) -> None:
             if self._exit_received:
                 break
             self._configuration.refresh()
-            if self._configuration.version != self._config_version:
+            if self._configuration.version != self._config_version or self._configuration.changed:
```
kmeinerz: I think it would probably be better to only check if the version has changed. Allowing also dynamic updates can lead to confusion, where same versions are actually different configs.

logprep/processor/list_comparison/rule.py
```diff
@@ -103,15 +104,25 @@ def init_list_comparison(self, list_search_base_path: Optional[str] = None):
...
+    def _init_list_comparison_from_http(self, list_search_base_path):
+        for list_path in self._config.list_file_paths:
+            list_search_base_path_resolved = Template(list_search_base_path).substitute(
+                {**os.environ, **{"LOGPREP_LIST": list_path}}
+            )
+            http_getter = GetterFactory.from_string(list_search_base_path_resolved)
+            self._update_compare_sets_via_http(http_getter, list_path)
+            http_getter.add_callback(self._update_compare_sets_via_http, http_getter, list_path)
+
+    def _update_compare_sets_via_http(self, http_getter, list_path):
```
kmeinerz: Missing type hints

logprep/util/getter.py
```diff
@@ -139,28 +167,178 @@ def __attrs_post_init__(self):
...
+                    raise error
+        else:
+            try:
+                self._update_cache()
+            except requests.exceptions.RequestException as error:
+                if self.cache is None:
+                    raise error
+                self._log_cache_warning(error)
+        if self.cache is None:
+            raise ValueError(f"Cache is empty for http getter with url '{self.url}'")
```
kmeinerz: Not covered by tests

kmeinerz (changes requested):
Thank you for your work, looks already really good.
Please take a look at my comments.

pyproject.toml
```diff
@@ -90,6 +90,7 @@ dependencies = [
   "falcon>=3.1.3",
   "uvloop",
   "httptools",
+  "starlette"
```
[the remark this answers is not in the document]
ppcad: It was not really necessary, I removed it.

logprep/runner.py
ppcad: I have removed the reload being triggered by config changes, but kept the detection of getter value changes to determine if a config has changed.

ppcad:
I added an option for refreshable getters to return default values. Can you please take a look? @kmeinerz

logprep/processor/generic_adder/rule.py
```diff
@@ -143,9 +143,23 @@ class Config(FieldManagerRule.Config):
...
         def __attrs_post_init__(self):
+            self._base_add = copy.deepcopy(self.add)
+
+            for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
+                getter = GetterFactory.from_string(add_file)
+                if isinstance(getter, RefreshableGetter):
+                    getter.add_callback(self._refresh_add)
+
             if self.add_from_file:
                 self._add_from_path()
```
mhoff:
```suggestion
            if self.add_from_file:
                for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
                    getter = GetterFactory.from_string(add_file)
                    if isinstance(getter, RefreshableGetter):
                        getter.add_callback(self._refresh_add)
                self._add_from_path()
```

Seems logical to me to group the file-related stuff when doing the conditional stuff either way.

logprep/ng/processor/generic_resolver/processor.py
```diff
@@ -149,17 +156,14 @@ def _apply_rules(self, event: dict, rule: GenericResolverRule) -> None:
         if conflicting_fields:
             raise FieldExistsWarning(rule, event, conflicting_fields)
 
-    def _find_content_of_first_matching_pattern(
-        self, rule: GenericResolverRule, source_field_value: str
-    ) -> str | None:
+    def _find_content_of_first_matching_pattern(self, rule, source_field_value):
```
kmeinerz: Is there a reason why you removed the type hints?

logprep/util/configuration.py
```diff
@@ -871,6 +875,29 @@ def reload(self) -> None:
...
+
+            if isinstance(value, dict):
+                for val in value.values():
+                    get_variable_values(val, variable_values)
+            elif isinstance(value, (list, tuple, set)):
+                for val in value:
+                    get_variable_values(val, variable_values)
+            elif isinstance(value, str):
+                if value.startswith(("http://", "https://", "file://")):
+                    variable_values.append(GetterFactory.from_string(value).get_json())
```
mhoff: I am struggling to grasp the implications of this change and see a potential risk here.

What the change does is to follow all references (http/https endpoints and files), collect their contents (`get_json`) and include this data in the check whether the configuration has changed. Semantically I can see where this is going -- configuration changes might be hidden behind files/endpoints and we want to make sure that the config is also reloaded when the referenced data changes.

Potential problems I see:
1. We are always using `get_json` which will probably fail (=Exception) for referenced yaml-files. Solution: use something more flexible like `get_dict` instead
2. We are traversing the whole config tree, effectively pulling in all referenced values (e.g. processor-specific rule configuration data)
2.1. Who is responsible for reloading processor-specific data? Does the processor register callbacks on its getters (as [this PR does it for the generic resolver](https://github.com/fkie-cad/Logprep/pull/882/files#diff-003f0669d3897777df80e086070b66beee9a7fad11a894b7c47c58cf7ae43cfbR196)) and update itself, or does the config reload everything and initiate pipeline reloads? In any case we shouldn't mix approaches. (I btw favor the variant that the processor takes the responsibility and reloads itself).
2.2. We are potentially iterating over a lot of references, taking quite some time in the manager process (I/O) and collecting a significant amount of data (caches + parsed data). We even need to preserve all the data for comparison purposes. This might block the manager process and even lead to OOMs.
2.3. The manager process suddenly collects data which otherwise would only be loaded in the pipeline processes, leading to duplicated data across processes and thus a higher overall memory usage.
3. In the end, [the pipelines are only being reloaded if the configuration version has changed](https://github.com/fkie-cad/Logprep/blob/main/logprep/runner.py#L122). So I wonder if loading the complete configuration is actually required and at least the pipelines-subtree of the configuration could be excluded.
3.1. I am not sure about how other things (like metrics on the runner- or manager-level) are being reloaded and maybe these things take place immediately and without a version change. However, this would be a weird mixture of paradigms again.

logprep/util/getter.py
```diff
@@ -139,40 +384,62 @@ def __attrs_post_init__(self):
...
+            (
+                "Credentials file not found. Please set the environment variable "
+                f"'{ENV_NAME_LOGPREP_CREDENTIALS_FILE}'"
+            )
+        ) from error
+
+
+def refresh_getters():
+    """Refreshes all refreshable getters"""
+    HttpGetter.refresh()
```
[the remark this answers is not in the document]
ppcad: This would not work, since each class has it's own `_shared` variable.
`_shared` is defined as an abstract property in the superclass.

logprep/util/configuration.py
ppcad: > I am struggling to grasp the implications of this change and see a potential risk here.
> 
> What the change does is to follow all references (http/https endpoints and files), collect their contents (`get_json`) and include this data in the check whether the configuration has changed. Semantically I can see where this is going -- configuration changes might be hidden behind files/endpoints and we want to make sure that the config is also reloaded when the referenced data changes.
> 
> Potential problems I see:
> 
>     1. We are always using `get_json` which will probably fail (=Exception) for referenced yaml-files. Solution: use something more flexible like `get_dict` instead
Yes, I will change it to `get_dict`.
> 
>     2. We are traversing the whole config tree, effectively pulling in all referenced values (e.g. processor-specific rule configuration data)
>        2.1. Who is responsible for reloading processor-specific data? Does the processor register callbacks on its getters (as [this PR does it for the generic resolver](https://github.com/fkie-cad/Logprep/pull/882/files#diff-003f0669d3897777df80e086070b66beee9a7fad11a894b7c47c58cf7ae43cfbR196)) and update itself, or does the config reload everything and initiate pipeline reloads? In any case we shouldn't mix approaches. (I btw favor the variant that the processor takes the responsibility and reloads itself).
A reload rebuilds all pipelines so it would reload all processors with their rules. This does only happen on config version changes when the configuration data has changed. So it is triggered intentionally. The only difference to before this pull request is that it also checks for the difference of data in paths and not the paths themselves. Shifting the responsibility for reloading processor specific stuff on changes of values resolved from paths solely to callbacks might be better.
>        2.2. We are potentially iterating over a lot of references, taking quite some time in the manager process (I/O) and collecting a significant amount of data (caches + parsed data). We even need to preserve all the data for comparison purposes. This might block the manager process and even lead to OOMs.
Yes, maybe it would be better to check for a config version change before parsing the whole config if we want to keep this functionality.
>        2.3. The manager process suddenly collects data which otherwise would only be loaded in the pipeline processes, leading to duplicated data across processes and thus a higher overall memory usage.
Alternatively, one could always reload if the config version changes even if the config content did not change. Then we don't have to keep track of variable values, but every config version change would cause a reload, which would probably happen anyways, since the version is usually changed when the content is also changed.
>     3. In the end, [the pipelines are only being reloaded if the configuration version has changed](https://github.com/fkie-cad/Logprep/blob/main/logprep/runner.py#L122). So I wonder if loading the complete configuration is actually required and at least the pipelines-subtree of the configuration could be excluded.
Yes, we could exclude the pipeline from resolving paths and rely only on callbacks for processor specific reloads triggered by changes from resolved paths.
>        3.1. I am not sure about how other things (like metrics on the runner- or manager-level) are being reloaded and maybe these things take place immediately and without a version change. However, this would be a weird mixture of paradigms again.

ppcad: Yes, that would be fine for me.

logprep/ng/processor/generic_resolver/processor.py
```diff
@@ -47,7 +47,15 @@ class Config(Processor.Config):
             validator=validators.optional(validators.instance_of(int)), default=0
         )
         """(Optional) Size of cache for results when resolving form a list.
```
mhoff:
```suggestion
        """(Optional) Size of cache for results when resolving from a list.
```

logprep/ng/processor/generic_resolver/processor.py
```diff
@@ -153,13 +164,12 @@ def _find_content_of_first_matching_pattern(
...
     ) -> str | None:
         if rule.resolve_from_file:
-            replacements = rule.resolve_from_file["additions"]
             matches = rule.pattern.match(source_field_value)
             if matches:
                 mapping = matches.group("mapping")
                 if rule.ignore_case:
                     mapping = mapping.upper()
-                content = replacements.get(mapping)
+                content = rule.additions.get(mapping)
```
mhoff: @ppcad Could you please elaborate a bit on this change?

My understanding:
`rule.additions` essentially represents the contents of the file in `rule.resolve_from_file['path']`. This seems different from `rule.resolve_from_file['additions']` (which also [seems to be a disallowed key for the dict](https://github.com/fkie-cad/Logprep/blob/2d70682dbad2e24c6e08eec8ac1bf6a1ef31058c/logprep/processor/generic_resolver/rule.py#L155)). I am struggling to see how this change is simply refactoring existing functionality. If there has been a bug before, I'd expect a short statement in the changelog, but feel free to elaborate. I feel like I am missing something.

logprep/abc/getter.py
```diff
@@ -93,6 +93,27 @@ def get_json(self) -> Union[Dict, List]:
...
 
+    def get_collection(self) -> Union[Dict, List]:
+        """gets and parses the raw content to yaml or json if yaml fails"""
+        try:
+            return self.get_yaml()
+        except YAMLError:
+            return self.get_json()
+
+    def get_dict(self) -> dict:
+        """gets dict and fails otherwise"""
```
mhoff:
```suggestion
        """Gets dict and fails otherwise"""
```

logprep/abc/getter.py
```diff
@@ -93,6 +93,27 @@ def get_json(self) -> Union[Dict, List]:
...
+
+    def get_list(self) -> list:
+        """gets list and fails otherwise"""
+        result = self.get_collection()
+        if not isinstance(result, list):
+            raise ValueError("Value is not a list")
+        return result
+
     def get_jsonl(self) -> List:
         """gets and parses the raw content to jsonl"""
```
mhoff:
```suggestion
        """Gets and parses the raw content to jsonl"""
```

logprep/util/getter.py
```diff
@@ -56,11 +69,11 @@ def from_string(cls, getter_string: str) -> "Getter":
         if protocol is None:
             protocol = "file"
         if protocol == "file":
-            return FileGetter(protocol=protocol, target=target)
+            return FileGetter(protocol=protocol, target=target)  # type: ignore
         if protocol == "http":
-            return HttpGetter(protocol=protocol, target=target)
+            return HttpGetter(protocol=protocol, target=target)  # type: ignore
```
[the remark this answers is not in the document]
mhoff: I'm wondering... why doesn't it suffice to have the (not-abstract) class variable on the level for `RefreshableGetter`? I get that subclasses might want to use different `SharedDataPerTarget` classes or sth., but I am not sure if we actually need that.

logprep/ng/processor/generic_resolver/processor.py
ppcad: As far as I can see `rule.additions` and `replacements` are the same. The only difference is that getting it can raise a configuration exception now.

In rules.py:
`self.resolve_from_file["additions"] = additions` became `self.additions = additions`.
Thus `self.resolve_from_file["additions"]` is the same as `self.additions`.

In processory.py:
`replacements = rule.resolve_from_file["additions"]` is therefore the same as `rule.additions`.

`resolve_from_file` is a configuration parameter that did not expect `additons`, but it was used to store `additions` after initialization. It made more sense to me to put `additons` in its own variable, but functionally nothing should have changed.

mhoff: So we essentially use the new config field to _write_ to it in post-init and then later on use the pre-computed value. That is definitely better than to add an untyped dict key ("additions") and store that in a user-supplied config dict. Thank you for the clarification -- that makes sense and I like this change.

On a more general note: In my mind, the config is primarily user-defined/supplied and should generally be read-only. I do understand though, that `init=False` basically says that the field is derived / set internally and that we (not limited to `init=False`) do indeed modify config fields in some processors/rules. At some point in the future I would like to draw a clear boundary between a readonly config and state which is derived / computed as part of the runtime.

Also, even though I like python properties, sometimes using a lot of them makes it harder to trace what we actually do. It took me a moment to realize that `self._file_path` is `self.resolve_from_file['path']` whereas that refers to `config.resolve_from_file`. Maybe we could improve on the property naming to make things more clear. I am still thinking about that, so no TODO for now.

logprep/util/configuration.py
```diff
@@ -839,15 +836,6 @@ def reload(self) -> None:
...
             new_config = Configuration.from_sources(self.config_paths)
-            if self._config_failure:
-                logger.info("Config refresh recovered from failing source")
-            self._config_failure = False
-            if new_config == self:
-                logger.info(
-                    "Configuration version didn't change. Continue running with current version."
-                )
-                self._set_config_refresh_interval(new_config.config_refresh_interval)
-                return
```
mhoff: We [keep this block in the new architecture](https://github.com/fkie-cad/Logprep/pull/882/files#diff-6c9f91348fe111796648b32ee3de8debd0be48ae56053dcfdd1f17d731730b91R842), but we delete it in the old one. Is there a reason for the discrepancy?

I wonder about the benefits of this conditional. Do we save CPU resources by shortcutting the case where the configs are equal? Do you have any insight on this?

ppcad: I believe it was added to save resources.
Furthermore, the manager did only reload if the version changed **and** `new_config == self` was true.
Now the manager reloads on all version changes so I removed it.

I can change it in the ng version too. I was barely touching the ng version.

tests/unit/util/test_getter.py
```diff
@@ -672,3 +689,754 @@ def test_get_raw_reuses_mtls_session_and_ca_cert_is_not_updated(self, tmp_path):
...
+            http_getter: HttpGetter = GetterFactory.from_string(f"http://{target}")
+            with pytest.raises(RefreshableGetterError, match="Test"):
+                http_getter.get_raw()
+
+    @mock.patch(
+        "logprep.util.getter.HttpGetter._update_cache", side_effect=RefreshableGetterError("Test")
+    )
+    def test_get_raw_without_interval_logs_warning_on_error_with_empty_cache(
+        self, _, tmp_path, caplog
+    ):
```
kmeinerz:
```suggestion
    def test_get_raw_without_interval_logs_warning_on_error_with_empty_cache(
        self, _, caplog
    ):
```

logprep/util/configuration.py
```diff
@@ -841,13 +838,10 @@ def reload(self) -> None:
             new_config = Configuration.from_sources(self.config_paths)
             if self._config_failure:
                 logger.info("Config refresh recovered from failing source")
-            self._config_failure = False
```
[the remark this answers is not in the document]
mhoff: Can we combine the two log messages?

```
            new_config = Configuration.from_sources(self.config_paths)
            if self._config_failure:
                logger.info("Config refresh recovered from failing source")
                logger.info("Configuration didn't change.")
                self._set_config_refresh_interval(new_config.config_refresh_interval)
                return
            self._config_failure = False
            if new_config.config_refresh_interval is None:
                new_config.config_refresh_interval = self.config_refresh_interval
```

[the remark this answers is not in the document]
ppcad: I have removed `logger.info("Configuration didn't change.")` from the non-ng configuration, since reloads don't depend on it anymore. Do you want me to re-add it to both versions?

kmeinerz (commented):
Thanks for incorporating the changes. The PR looks good to me.
But I'll leave give the approval to the other review to avoid premature merging.

logprep/util/getter.py
```diff
@@ -56,11 +69,11 @@ def from_string(cls, getter_string: str) -> "Getter":
...
         if protocol == "https":
-            return HttpGetter(protocol=protocol, target=target)
+            return HttpGetter(protocol=protocol, target=target)  # type: ignore
```
mhoff: Do we still need those ignores after fixing the problem with `_shared`?

ppcad: I was hoping we wouldn't , but mypy still complains, even now.

mhoff:
> > Hello @ppcad, looks good to me as well. I found 2 small topics which should be resolved quickly. It would also be great if you could squash some of the review commits as the list has become quite long
> 
> The commits are going to be squashed after merging this branch into main. Is that OK for you?

That would be perfectly fine. I just didn't see it for the other PRs, so I thought standard merge commits (and no squashing) would rather be common practice

PR #885, opened 2025-10-09
Make list comparison accept lists in documents and support matching networks

ppcad:


mhoff:
@ppcad Thank you for the PR!

I think it would be more suitable to have a dedicated processor for this kind of functionality, which might/should extend the `list_comparison` processor. This new processor would then only expect ip addresses or ip address ranges and could then specialize on this functionality.

Advantages:
- Fine-granular processors
- No or less try...except flows for happy path scenarios
- No complex branching logic if further formats are to be supported (e.g. regex, ...)

What do you think?

ppcad:
> @ppcad Thank you for the PR!
> 
> I think it would be more suitable to have a dedicated processor for this kind of functionality, which might/should extend the `list_comparison` processor. This new processor would then only expect ip addresses or ip address ranges and could then specialize on this functionality.
> 
> Advantages:
> 
>     * Fine-granular processors
> 
>     * No or less try...except flows for happy path scenarios
> 
>     * No complex branching logic if further formats are to be supported (e.g. regex, ...)
> 
> 
> What do you think?

Yes, it may be better extract this functionality into a dedicated processor. Do you think that the new processor should inherit from list comparison or should it be a separate processor?

logprep/ng/processor/list_comparison/processor.py
```diff
@@ -81,14 +81,20 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list[
         the result (key) and a list of filenames pertaining to said result (value).
         """
 
+        # get value that should be checked in the lists
```
mhoff: I personally would prefer to not have this kind of comment and instead name the variable `field_value_to_be_checked` (or sth.) instead. I know this is kind of explicit and due to this longer expressions can feel clunky. The advantage however is that typically this kind of documentation ages better and makes the code more readable because you don't have to keep the mapping between variable name and comment in your head

logprep/ng/processor/list_comparison/processor.py
```diff
@@ -81,14 +81,20 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list[
...
-        list_matches = [
-            compare_list
-            for compare_list in rule.compare_sets
-            if field_value in rule.compare_sets[compare_list]
-        ]
-
-        if not list_matches:
+        value_list = field_value if isinstance(field_value, list) else [field_value]
+
+        # iterate over lists and check if element is in any
```
mhoff: Same here with the comment. Rather extract a method with the name `check_for_value_in_lists` (or sth.)

logprep/ng/processor/list_comparison/processor.py
```diff
@@ -81,14 +81,20 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list[
         the result (key) and a list of filenames pertaining to said result (value).
```
mhoff: ^ is "res_key" defined somewhere?

logprep/ng/processor/list_comparison/processor.py
```diff
@@ -81,14 +81,20 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list[
...
+        value_list = field_value if isinstance(field_value, list) else [field_value]
+
+        # iterate over lists and check if element is in any
+        list_matches = []
+        for field_value in value_list:
+            for compare_list in rule.compare_sets:
+                if compare_list in list_matches:
+                    continue
+                if field_value in rule.compare_sets[compare_list]:
+                    list_matches.append(compare_list)
```
mhoff: Optional:
One way to simplify this could be to convert everything to sets and simply do intersections checks. However, intersections compute the whole intersection and don't terminate after the first overlap, [so it would probably be less performant](https://stackoverflow.com/questions/6821329/speed-differences-between-intersection-and-object-for-object-in-set-if-object).

logprep/processor/network_comparison/processor.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # iterate over IP string lists and check if element is in any
+        list_matches: list = []
+        for field_value in value_list:
+            self._compare_with_networks(list_matches, field_value, rule)
+
+        # if matching list was found return it, otherwise return all list names
+        if len(list_matches) == 0:
+            return list(rule.compare_sets.keys()), "not_in_list"
+        return list_matches, "in_list"
```
mhoff: See my comments for the list_comparison processor. Additionally it seems very useful to have a method for lines 53 - 55 and simply override the parent method. Lines 49/50 + (58 - 60) are the same between both implementations and can be deduduplicated

logprep/processor/network_comparison/processor.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @staticmethod
+    def _compare_with_networks(list_matches: list, field_value: str, rule: NetworkComparisonRule):
+        """Iterate over network lists, check if element is in any."""
+        try:
+            ip_address_object = ip_address(field_value)
+        except ValueError:
+            return
```
mhoff: I wonder if we should (at least have the possibililty) to produce warnings here. We intentionally want to compare with ip ranges and it might be a configuration problem if the source fields do not contain ip addresses as expected

logprep/ng/processor/list_comparison/processor.py
ppcad: Thank you for the suggestion, In my opinion performance is more important in this place, even if it's just a small difference.

logprep/ng/processor/list_comparison/processor.py
ppcad: I cleaned up the comments. I previously kept them as they are without noticing that they were already removed from the ng version. I agree that it's better to not have them.

logprep/processor/network_comparison/processor.py
mhoff: I think it's fair to assume for processor methods to have the respective event being processed in-scope. So I'd be fine with passing it down in this case.

mhoff (approved):
Thank you for the changes. The compose setup tests can be disregarded for now

PR #889, opened 2025-10-20
Doc new runner notebook

kaya-david:
* add notebook example for runner

* fix review issues:
-- avoid using explicit is None checks. Instead, use typing.cast(...) where appropriate (see test coverage for runner/sender to identify the related cases).
-- stopping the runner should be possible even when there are events in the event_backlog with the state "DELIVERED".

### DIFF
--- a/logprep/ng/runner.py
+++ b/logprep/ng/runner.py
@@ -44,7 +44,7 @@ class Runner:
 
         self.configuration = configuration
         self._running_config_version = configuration.version
-        self.__input_connector: Input | None = None
+        self._input_connector: Input | None = None
 
         # Initialized in `setup()`; updated by runner logic thereafter:
         self.should_exit: bool | None = None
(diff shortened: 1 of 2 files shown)


doc/source/development/notebooks/new_architecture_examples/13_runner.ipynb
```diff
@@ -0,0 +1,293 @@
...
+    "# waiting for kafka to be ready\n",
+    "sleep 30"
+   ],
+   "id": "30f803a069d4c7dc",
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)\n",
```
mhoff: Can we silence this? I didn't see this in the other outputs

PR #891, opened 2025-10-30
fix: replace bitnami images with functional alternatives

Pablu23:
Fix docker-compose and k8s setup by changing out pay to use bitnami images for free to use bitnamilegacy images

### DIFF
--- a/doc/source/examples/minikube.rst
+++ b/doc/source/examples/minikube.rst
@@ -4,8 +4,8 @@ Kubernetes Example Deployment
 For this example, we need a working kubernetes cluster. Here we will use minikube,
 but every other kubernetes environment should do the job.
 
-Setup Minikube (MacOS)
-----------------------
+Setup Minikube (MacOS EXPERIMENTAL)
+-----------------------------------
 
 This tutorial assumes that you have `podman`, `minikube`,` `kubectl` and `helm` already installed.
 


mhoff:
@Pablu23 The compose setup didn't work for me.

I had the following issues:

- logprep didn't start without PROMETHEUS_MULTIPROC_DIR
- logprep tried to launch while kafka wasn't up yet (mitigatable through restarting logprep manually, unclean though)
- kafka-exporter only produced "connection refused" when talking to kafka

Did this actually work on your system? And does it work now?

Pablu23:
> @Pablu23 The compose setup didn't work for me.
> 
> I had the following issues:
> 
>     * logprep didn't start without PROMETHEUS_MULTIPROC_DIR
> 
>     * logprep tried to launch while kafka wasn't up yet (mitigatable through restarting logprep manually, unclean though)
> 
>     * kafka-exporter only produced "connection refused" when talking to kafka
> 
> 
> Did this actually work on your system? And does it work now?

Hello @mhoff thanks for your reply, regarding your issues:
1. Yes that is intentional, you have to set it up manuelly
2. Afaik the Compose setup doesnt start logprep itself / is mostly used to setup Opensearch and Kafka so you can run logprep locally, meaning yes, if you start it before kafka is up it tries to launch anyways
3. That I do not know, maybe we should think about sliming down the compose setup, the only components commonly started are kafka and opensearch afaik.

I did test this, the compose setup runs, but I did not test the functionality of each individuell component as it was not the main goal of this PR, this PR was just supposed to fix the Docker error when trying to pull images that dont exist.

Hope this clears things up, if not please contact me again :)

mhoff (approved):
I have checked Helm and the Compose environment. Thanks!

mhoff (dismissed):

mhoff (approved):
As discussed. The minikube tutorial for MacOS is a bit flaky and not properly reproducible (hence "experimental"). However the other changes are fine and we could locally verify that everything works at least on one machine. Approved.

PR #895, opened 2025-11-06
feat: upgrade to Python 3.14 and drop support for 3.10

Pablu23:
Add support for Python 3.14, and remove support for Python 3.10

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -140,7 +140,7 @@ logprep-ng = "logprep.run_ng:cli"
 
 [tool.black]
 line-length = 100
-target-version = ['py312', 'py313', 'py314']
+target-version = ['py311', 'py312', 'py313', 'py314']
 extend-exclude = '''
 # A regex preceded with ^/ will apply only to files and directories
 # in the root of the project.


examples/compose/docker-compose.yml
mhoff: There seem to be a few formatting changes. Please make sure that your formatter extensions and their configuration are in line with the project. I can say that at least I didn't have these formatting changes when saving the file, but I wouldn't guarantee that my dev environment is correctly set up right now. Happy to discuss!

logprep/run_ng.py
```diff
@@ -16,6 +17,8 @@
 EPILOG_STR = "Check out our docs at https://logprep.readthedocs.io/en/latest/"
 init_yaml_loader_tags("safe", "rt")
 
+set_start_method("fork", True)
```
mhoff: Same comment as for the non-ng above.

examples/compose/docker-compose.yml
Pablu23: As this is a yml file I would like to know if you have a yaml lsp / formatter installed, if not thats probably the cause. Also the way the test is written is afaik not the recommended way to write Arrays, I think the updated version is better, if it shouldnt be completly swapped to yml multiline array notation, eg:
```yaml
test:
  - "CMD-SHELL"
  - "kafka-topics.sh ..."
```

mhoff (dismissed):
LGTM, thanks!

tests/unit/util/test_getter.py
```diff
@@ -208,7 +209,12 @@ def test_get_returns_binary_content(self):
         file_getter = GetterFactory.from_string("/my/file")
         with mock.patch("pathlib.Path.open", mock.mock_open(read_data=b"my content")) as mock_open:
             content = file_getter.get_raw()
-            mock_open.assert_called_with(mode="rb")
+            # Python version depending: In Python 3.14, the buffering parameter was added to open() calls in pathlib.Path.read_bytes()
+            # pythonVersion >= 3.14.0 (https://docs.python.org/3/c-api/apiabiversion.html#bit-packing-macros)
+            if sys.hexversion >= 0x030E0000:
+                mock_open.assert_called_with(mode="rb", buffering=0)
+            else:
+                mock_open.assert_called_with(mode="rb")
```
mhoff: Can be improved like this: https://github.com/fkie-cad/Logprep/pull/900/files#diff-8b74cd3fe276caa691eb23f914e14a2616e893852ecb2b59671b34b4c198dbdfR172-R173

tests/acceptance/test_config_refresh.py
```diff
@@ -45,23 +45,25 @@ def test_two_times_config_refresh_after_5_seconds(tmp_path, config):
...
-    proc = start_logprep(config_path)
-    wait_for_output(proc, "Config refresh interval is set to: 5 seconds", test_timeout=8)
-    wait_for_output(proc, "Successfully reloaded configuration", test_timeout=7)
-    stop_logprep(proc)
+    with run_logprep(config_path) as proc:
+        wait_for_output(proc, "Config refresh interval is set to: 5 seconds", test_timeout=5)
+        wait_for_output(
+            proc,
+            "Successfully reloaded configuration",
+            test_timeout=1000,
```
mhoff: That is a tad long in case something goes wrong

tests/unit/util/test_getter.py
Pablu23: Can you explain what call_args[1] is in that case? If there is a logical reason to why thats a 1 and it doesnt change on different args, like buffering being there, im glad to change that

pyproject.toml
```diff
@@ -91,6 +91,8 @@ dependencies = [
   "falcon>=3.1.3",
   "uvloop",
   "httptools",
+  "types-psutil",
+  "types-PyYaml"
 ]
```
mhoff: I think this rather belongs in the dev dependencies below

Pablu23: Yes, but two questions, in our pipeline do we install the dev dependencies or not? If not, the Problem from mypy would still happen, or am I understanding something wrong? If not I can move them from the core dependencies to dev

pyproject.toml
```diff
@@ -150,7 +153,7 @@ extend-exclude = '''
 profile = "black"
 
 [tool.cibuildwheel]
-build = "cp310-* cp311-* cp312-*"
+build = "cp312-* cp313-* cp314-*"
```
mhoff: cp311-* missing

mhoff (approved):
Many thanks for the extensive changes!

PR #896, opened 2025-11-11
feat: allow pre-detector to copy extra fields from log to detection event

mhoff:


logprep/util/helper.py
```diff
@@ -195,50 +229,166 @@ def add_fields_to(
         raise FieldExistsWarning(rule, event, unsuccessful_targets_resolved)
 
 
-def _get_slice_arg(slice_item):
+def _get_slice_arg(slice_item) -> Optional[int]:
```
Pablu23: I think we had a discussion about Optional, wasnt it better to use "int | None" in this case, as optionals are getting phased out?

logprep/util/helper.py
```diff
@@ -364,13 +514,58 @@ def snake_to_camel(snake: str) -> str:
 append_as_list = partial(add_fields_to, merge_with_target=True)
 
 
+def copy_fields_to_event(
+    target_event: dict,
+    source_event: dict,
+    dotted_field_names: Iterable[str],
+    *,
```
Pablu23: What is this? A singular lone standing * ?

Pablu23: Ahh thank you for the explanation :)

kaya-david (commented):
Great work! Please look at my comments. :)

kaya-david (dismissed):

kaya-david (approved):

PR #902, opened 2025-11-18
fix: repair automated check for the compose configuration

mhoff:


Pablu23 (approved):
Looks very good thanks for the changes :)

PR #911, opened 2025-11-28
feat: add acceptance tests for HTTP input metadata collection

kaya-david:
* test cover for HTTP input meta data collection

### DIFF
--- a/tests/acceptance/test_http_input_with_requests.py
+++ b/tests/acceptance/test_http_input_with_requests.py
@@ -117,12 +117,12 @@ def test_http_input_respects_collect_meta_flag(
         )
 
         time.sleep(0.5)
-        content_str = output_path.read_text()
-        content_json = json.loads(content_str)
+        output_str = output_path.read_text()
+        output_json = json.loads(output_str)
 
         if expect_metadata:
-            assert "@metadata" in content_json, content_str
+            assert "@metadata" in output_json, output_str
             for field in ("url", "remote_addr", "user_agent"):
-                assert field in content_json["@metadata"], f"{field} not in {content_str}"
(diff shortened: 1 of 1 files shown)


tests/acceptance/test_http_input_with_requests.py
```diff
@@ -77,3 +78,52 @@ def test_http_input_accepts_message_for_multiple_pipelines(tmp_path: Path, confi
...
+            verify=False,
+            timeout=5,
+        )
+
+        time.sleep(0.5)
+        content_str = output_path.read_text()
+        content_json = json.loads(content_str)
+
+        if expect_metadata:
+            assert "@metadata" in content_json, content_str
```
mhoff: Maybe add `metafield_name` to the parameterization (unset leads to "@metadata")

mhoff (changes requested):
Thanks, I got just a few minor remarks

tests/acceptance/test_http_input_with_requests.py
kaya-david: This parametrization now covers all relevant configuration scenarios:

- collect_meta=True with no metafield_name set → ensures the default @metadata field is used and populated.
- collect_meta=False → ensures that no metadata field is emitted at all.
- collect_meta=True with an explicit metafield_name → ensures metadata is written under the custom field name.

This should validate both the default behaviour and the explicit configuration paths.

mhoff (approved):
Many thanks, looks good to me now!

Approved assuming all GHA run successfully. 

Please sync with https://github.com/fkie-cad/Logprep/pull/912 as they extend the exact same functionality which is under test here

PR #921, opened 2026-01-09
fix dissector curly braces problem

ekneg54:
this handles curly braces one char before end of section.

feel free to have a look at the pipeline output of the first commit to understand the problem.

I had to remove the feature to handle curly braces in field names.
But ensured that curly braces in field content still works.
I think the tests for handling special fields in field names are more
sience and fiction than reality.


logprep/processor/dissector/rule.py
```diff
@@ -113,7 +113,7 @@
 ACTION = r"(?P<action>[+])?"
 STRIP_CHAR = r"(-\((?P<strip>.)\))?"
 SEPERATOR = r"(\((?P<separator>\\\)|[^)]+)\))?"
-TARGET_FIELD = r"(?P<target_field>[^\/\|-]*)"
+TARGET_FIELD = r"(?P<target_field>[^\/\|\}-]*)"
```
mhoff: @ekneg54 Maybe I have not properly understood the implementation, but shouldn't `TARGET_FIELD` and `VALID_TARGET_FIELD` be semantically equivalent?

Maybe not exactly, but something like this?:

```suggestion
TARGET_FIELD = rf"(?P<target_field>[^\/\|\}}-]{VALID_TARGET_FIELD}*)"
```

ekneg54: from semantics side you should be right. but from code site this doesn't work. And to be honest I have no clue why this ist written this way. I only came here to fix this bug. I have no time to dig deep into the implementation. Sorry for that. 

The only thing I remember is that I had to design the validation regex a little bit broader. It is used in rule validation via attrs. But I guess it is worth it to have a look into. But not in conjunction with this bugfix.

Pablu23 (approved):
Looks good so far, but I also definitly see why we need so many tests for this Processor, in the future we would like to have a deep dive into it and maybe rework / refactor it.

PR #951, opened 2026-03-31
feat: use nix for image/package build and remove rust

Pablu23:
## Description

* Remove unused rust dependencies, remove setuptools_scm and use uv_build, add nix flake for container image, dev shell and application building

## Assignee

### Documentation

### Code Quality

* nix develop
* nix build .#docker.python311 && docker load < result && docker run logprep:py3.11
* nix build .#dist.python311 && ls result/

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--951.org.readthedocs.build/).


.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
mhoff, line 1: This is only required for `nix build`, right? I am a bit hesitant to add this new file solely for builds and I would assume that it would also be loaded for all other operations (like running logprep etc.)

.github/workflows/check-examples.yml
```diff
@@ -17,14 +17,23 @@ jobs:
         uses: docker/setup-compose-action@364cc21a5de5b1ee4a7f5f9d3fa374ce0ccde746 # https://github.com/docker/setup-compose-action/releases/tag/v1.2.0
         with:
           version: latest
+      - name: Set up Nix
+        uses: cachix/install-nix-action@v31
```
mhoff: Commit SHA. I assume you did not do this as the PR is not finalized yet, but I still wanted to add the comment as a reminder not to forget for future reviews

.github/workflows/check-examples.yml
```diff
@@ -17,14 +17,23 @@ jobs:
...
+          # to find out the version its running
+          # you would see the latest released version
+          nix build .#packages.x86_64-linux.docker.python${PY_VERSION//.} -o image
+          docker load < image
         env: &compose-env
           COMPOSE_PROFILES: ${{ matrix.profile }}
       - name: Run Compose setup
         working-directory: examples/compose
-        run: docker compose up -d
+        run: PYTHON_VERSION="${{ matrix.python-version }}" docker compose up -d
```
mhoff: This was a bug right? ... nice catch

.github/workflows/check-examples.yml
```diff
@@ -17,14 +17,23 @@ jobs:
         uses: docker/setup-compose-action@364cc21a5de5b1ee4a7f5f9d3fa374ce0ccde746 # https://github.com/docker/setup-compose-action/releases/tag/v1.2.0
         with:
           version: latest
+      - name: Set up Nix
+        uses: cachix/install-nix-action@v31
       - name: Build images
         working-directory: examples/compose
-        run: docker compose build --build-arg PYTHON_VERSION=${{ matrix.python-version }}
+        run: |
+          export PY_VERSION="${{ matrix.python-version }}"
```
mhoff: Why not use the same wording as below (`PYTHON_VERSION`)? Does `nix build` use `PY_VERSION` internally? If not, let's simply use `PYTHON_VERSION` and put in under `env` together with `COMPOSE_PROFILES`. This way (as we are reusing `compose-env` in "Run Compose setup" we have it all in one go.

.github/workflows/container-build.yml
```diff
@@ -58,40 +70,26 @@ jobs:
...
-          import os
-          imageid = "${{ steps.build-and-push.outputs.ImageID }}"
-          digest = "${{ steps.build-and-push.outputs.Digest }}"
-          with open(os.environ['GITHUB_OUTPUT'], 'a') as file:
-              file.write(f"digest={digest.split(':')[1]}")
-              file.write(f"imageid={imageid.split(':')[1]}")
-          EOF
+          export PY_VERSION="${{ matrix.python-version }}"
+
+          uv version "$(uv version --short).dev0+${GITHUB_HEAD_REF}.$(git rev-parse --short HEAD)"
```
mhoff, line 68: Please add a comment how this should look like in practice. Like "# 20.0.0.dev0+....."

.github/workflows/container-build.yml
```diff
@@ -58,40 +70,26 @@ jobs:
...
         run: |
-          python - <<EOF
-          import os
-          imageid = "${{ steps.build-and-push.outputs.ImageID }}"
-          digest = "${{ steps.build-and-push.outputs.Digest }}"
-          with open(os.environ['GITHUB_OUTPUT'], 'a') as file:
-              file.write(f"digest={digest.split(':')[1]}")
-              file.write(f"imageid={imageid.split(':')[1]}")
-          EOF
+          export PY_VERSION="${{ matrix.python-version }}"
```
mhoff: Maybe put `PYTHON_VERSION` here as well, if you change the name in check-examples.yml

.github/workflows/container-build.yml
```diff
@@ -58,40 +70,26 @@ jobs:
...
+          for tag in $(echo "${{ steps.image-tag.outputs.tags }}" | tr ", " "\n")
+          do
+            docker image tag logprep:py${PY_VERSION} $tag
+            docker image push $tag
+
+            digest=$(docker inspect $tag | jq --raw-output ".[].Id")
+
+            # What happens with multiple tags?
+            echo "Digest=$digest" >> $GITHUB_OUTPUT
+            echo "Digest_no_sha=${digest#"sha256:"}" >> $GITHUB_OUTPUT
```
mhoff: Why uppercase?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * centralize trivy cache across branches using a daily workflow
 * fix vulnerable python dependencies
 * add .editorconfig for python
+* add flake.nix and change container building to nix build aswell
```
```suggestion
* add flake.nix and change container building to nix build as well
```

Dockerfile
```diff
@@ -50,7 +46,7 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 
 # Uninstall dependencies from pyproject.toml:build-system.requires
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv pip uninstall setuptools setuptools-scm setuptools-rust wheel
+    uv pip uninstall setuptools setuptools-scm wheel
```
mhoff: This file can be deleted, right?

doc/README.md
```diff
@@ -2,3 +2,6 @@
 
 * install pandoc with `apt-get install pandoc`
 * install requirements with `uv sync --frozen --extra doc` in project root
+
+## nix version
+* nix develop
```
mhoff, line 8: I don't get what this means. Maybe still a placeholder. Needs more context because people are typically not familiar with nix

Edit: installation.rst makes it more clear, but still

pyproject.toml
```diff
@@ -1,32 +1,19 @@
 [build-system]
 requires = [
-  "setuptools>=82.0.0",
-  "setuptools-scm>=9.2.2",
-  "wheel",
-  "setuptools-rust",
+"uv_build>=0.11.6,<0.12"
```
mhoff:
```suggestion
  "uv_build>=0.11.6,<0.12"
```

mhoff (changes requested):
Many thanks for the work! I did a first review pass

.envrc
Pablu23: this is for direnv, if you have direnv installed and allow the directory, it auto loads the nix flake. I definitly need it, I previously did not commit it, but I have seen it on nearly every big open source project, so I thought it would be best practice to commit aswell. This by itself does nothing

.github/workflows/check-examples.yml
Pablu23: Yeah it probably was, I think before we actually always just used py3.11 for the checks it seems. I just changed this, because there now is no image called logprep:py3.11 when running in py3.12 matrix as we dont build it directly in the docker file

.github/workflows/check-examples.yml
Pablu23: This has to do with the nameing of the docker file, .#docker.python311 is the current name for building 3.11, using .#docker.3.11 would then be something different as .#docker.311, thats why we export it as a variable and then do some Bash magic on it to transform 3.11 to python311, without the dot

.github/workflows/container-build.yml
Pablu23: it was uppercase before, but Im not opposed to changing that

.github/workflows/check-examples.yml
Pablu23: Ahh no now I get what you mean, yeah ill change that. I was a little confused

doc/README.md
mhoff: @Pablu23 I tried to run it on my machine and simply installing Nix and running `nix develop` does not seem to work. Apparently `nix develop` is an experimental feature `nix-command` which is disabled by default. I think the primary instructions should use non-experimental features

Dockerfile
```diff
@@ -1,78 +0,0 @@
-# syntax=docker/dockerfile:1.7
```
mhoff, line 1: .dockerignore has to go as well. Did we make sure that we don't copy extranous stuff (like examples) into the nix-baxed images?

.github/workflows/container-build.yml
```diff
@@ -17,10 +17,6 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # https://github.com/actions/checkout/releases/tag/v6.0.2
```
mhoff, line 17: We need to check that `publish-release-to-pypi` still works as expected. I'd assume it doesn't, as `uv_build` does not support extension modules and the old workflow explicitly builds wheels. You can temporarily copy the job `build-wheel-and-tarball` into `ci.yml` to test building the PyPI package (without publishing it)

Dockerfile
Pablu23: I have inspected the Dockerfile with dive, and we dont copy the source code directly, to the docker image, we copy the venv content with the built binaries to the image

.github/workflows/publish-release-to-pypi.yml
```diff
@@ -8,20 +8,13 @@ jobs:
...
       - name: Build wheels
-        run: python -m cibuildwheel --output-dir wheelhouse --archs x86_64
-      - name: Build binary wheel and a source tarball
-        run: pipx run build --sdist
-      - name: copy artifacts to dist folder
-        run: cp ./wheelhouse/* ./dist/
+        run: |
+          mkdir -p dist
+          nix build .#dist.python311 -o dist_sl
+          cp -rL dist_sl/* dist
```
mhoff, line 18:
```suggestion
          nix build .#dist.python311 -o dist
```

Why not like this?

mhoff: If there is a specific reason for using `cp` in particular with flags `-rL` a comment would be great to point that out

Pablu23: yes, nix only ever symlinks everything, the resulting dist_sl is a symlink, because I dont know how well the github upload action plays with symlinks I create an actual folder and copy the files from the symlinked directory into that dir. I will write a comment for that

mhoff: Feel free to resolve after that

.github/workflows/check-examples.yml
```diff
@@ -17,14 +17,23 @@ jobs:
...
+        run: |
+          export PYTHON_VERSION="${{ matrix.python-version }}"
+          # We dont set the actual version here,
+          # so if you would look into the container,
+          # to find out the version its running
+          # you would see the latest released version
+          nix build .#packages.x86_64-linux.docker.python${PYTHON_VERSION//.} -o image
+          docker load < image
         env: &compose-env
           COMPOSE_PROFILES: ${{ matrix.profile }}
```
mhoff, line 29: You can move the env variable into the environment block like this:

```suggestion
        run: |
          # We dont set the actual version here,
          # so if you would look into the container,
          # to find out the version its running
          # you would see the latest released version
          nix build .#packages.x86_64-linux.docker.python${PYTHON_VERSION//.} -o image
          docker load < image
        env: &compose-env
          COMPOSE_PROFILES: ${{ matrix.profile }}
          PYTHON_VERSION: ${{ matrix.python-version }}
```

Also, like this you can simply remove "PYTHON_VERSION=..." from the next step as this env variable is already propagated through the yaml anchor &compose-env

.github/workflows/container-build.yml
```diff
@@ -58,54 +70,50 @@ jobs:
...
+          docker load < image
+
+          for tag in $(echo "${{ steps.image-tag.outputs.tags }}" | tr ", " "\n")
+          do
+            docker image tag logprep:py${PYTHON_VERSION} $tag
+            docker image push $tag
+
+            digest=$(docker inspect $tag | jq --raw-output ".[].Id")
+
+            # What happens with multiple tags?
```
mhoff: Do we have an answer for that question? :)

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * centralize trivy cache across branches using a daily workflow
 * fix vulnerable python dependencies
 * add .editorconfig for python
+* add flake.nix and change container building to nix build as well
```
mhoff, line 11: remove rust

.github/workflows/container-build.yml
Pablu23: Well not 100% but different tags are the same image producing the same digest, so using the latest is fine here (atleast thats what I observed) and it was the same before. But ill look into this some more

Pablu23: This kind of confirms that the same image should always have the same digest: https://github.com/distribution/distribution/issues/1811#issuecomment-229433060

.github/workflows/check-examples.yml
```diff
@@ -17,11 +17,20 @@ jobs:
...
+      - name: Set up Nix
+        uses: cachix/install-nix-action@616559265b40713947b9c190a8ff4b507b5df49b # https://github.com/cachix/install-nix-action/releases/tag/v31.10.4
       - name: Build images
         working-directory: examples/compose
-        run: docker compose build --build-arg PYTHON_VERSION=${{ matrix.python-version }}
+        run: |
+          # We dont set the actual version here,
+          # so if you would look into the container,
+          # to find out the version its running
+          # you would see the latest released version
```
mhoff: I don't fully understand what this comment is trying to say here. First of all, the only version we are explicitly setting one line below is the python version. This seems to be proper and to have effect. So I figure you mean the Logprep version. But what exactly is the "latest released version"? "released" as in GitHub release? Probably not, as this is independent. So I guess you mean the version which is set in `pyproject.toml` from the last release. If that is the case (and this is also my understanding) then maybe simply do the following:
```suggestion
          # logprep version is assumed to be set in pyproject.toml
```

.github/workflows/container-build.yml
```diff
@@ -46,6 +32,12 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
```
mhoff: Commit SHA missing

.github/workflows/container-build.yml
```diff
@@ -58,54 +50,51 @@ jobs:
...
+          # version for dev builds looks like this
+          # 19.0.0.dev0+feat-abc.h31hd73
+          # ^----^      ^-------^      ^
+          # |           |       |------|
+          # |           |       |
+          # |           |       commit hash
+          # |           |
+          # |           branch name
+          # |
+          # Previous latest released version
```
mhoff:
```suggestion
          # Last release version (as of pyproject.toml)
```

.github/workflows/container-build.yml
mhoff: The question should be gone before merging

.github/workflows/publish-release-to-pypi.yml
```diff
@@ -8,20 +8,15 @@ jobs:
...
-        run: python -m cibuildwheel --output-dir wheelhouse --archs x86_64
-      - name: Build binary wheel and a source tarball
-        run: pipx run build --sdist
-      - name: copy artifacts to dist folder
-        run: cp ./wheelhouse/* ./dist/
+        run: |
+          mkdir -p dist
+          nix build .#dist.python311 -o dist_sl
+          # Nix by default always creates symlinks, because im unsure how well upload artifact plays with symlinks
+          # hard copy the resulting whl and tar.gz from the dist_sl symlink into a normal dist directory
```
mhoff:
```suggestion
          # hard copy to avoid issues with symlinks (which Nix uses by default)
```

doc/source/installation.rst
```diff
@@ -60,6 +60,21 @@ contribute to them.
...
     uv sync --frozen --extra dev # if you intend to contribute
 
+There is also the option of using nix flakes.
+
+Flakes are theoretically experimental so depending on which nix installer you used,
+you might have to follow this short guide:
+
+https://wiki.nixos.org/wiki/Flakes#Nix_standalone
+
+after that, you should be able to run the following code and just start developing.
```
mhoff:
```suggestion
After that, you should be able to run the following code and just start developing.
```

mhoff:
Should be removed or adapted:
https://logprep--951.org.readthedocs.build/en/951/installation.html#docker

mhoff:
This requires an image build first: https://github.com/fkie-cad/Logprep/blob/deaeece523825c436b7185d1c5b25a4875288b9e/doc/source/examples/compose.rst#run-with-getting-config-from-http-server-with-basic-authentication

.github/workflows/container-build.yml
```diff
@@ -46,6 +32,12 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install uv
+        uses: astral-sh/setup-uv@eb1897b8dc4b5d5bfe39a428a8f2304605e0983c #https://github.com/astral-sh/setup-uv/releases/tag/v7.0.0
```
mhoff: This version differs from other uses of this action

doc/source/examples/compose.rst
```diff
@@ -42,6 +42,18 @@ Run without Logprep Container (default)
...
+
+    .. code-block:: bash
+
+      # To build the image locally, from root project dir run
+      nix build .#docker.python311
+      docker load < result
+
+      # Or to download it
+      docker pull ghcr.io/fkie-cad/logprep:py3.11-v19.0.0
+      docker image tag ghcr.io/fkie-cad/logprep:py3.11-v19.0.0 logprep:py3.11
```
mhoff, line 59: This step is necessary to run `docker compose` with our example, right? I see that the image name there has changed from `logprep` to `logprep:py${PYTHON_VERSION:-3.11}`.

I think we should at least explain why we need to `image tag` the pulled image here

doc/source/installation.rst
```diff
@@ -2,112 +2,127 @@
...
+To see if the installation was successful run :code:`logprep --version`.
 
-https://docs.astral.sh/uv/getting-started/installation/#installing-uv
+Container Image
+===============
 
-To install Logprep you have the following options:
+This option is recommended for Users just wanting to run logprep without the hassle of compiling.
+We provide a number of pre compiled Images for all officially supported python versions.
+A list of every image can be found here https://github.com/fkie-cad/Logprep/pkgs/container/logprep/versions?filters%5Bversion_type%5D=tagged.
```
mhoff:
```suggestion
This option is recommended for users wanting to run logprep without the hassle of compiling.
We provide precompiled images for all python versions supported by logprep.
A list of available images can be found here https://github.com/fkie-cad/Logprep/pkgs/container/logprep/versions?filters%5Bversion_type%5D=tagged.
```

doc/source/installation.rst
```diff
@@ -2,112 +2,127 @@
...
 
-**1. Option:** latest stable release
+To install a specific version the general structure of our tags follows this example:
 
-This option is recommended if you just want to use the latest stable release of Logprep.
+logprep:py{PYTHON_VERSION}-v{LOGPREP_VERSION}
 
-.. code-block:: bash
+So python 3.11 with version 19.0.0 would be:
```
mhoff:
```suggestion
Python 3.11 with version 19.0.0 would be called:
```

doc/source/installation.rst
```diff
@@ -2,112 +2,127 @@
...
 
-    git clone https://github.com/fkie-cad/Logprep.git logprep && cd logprep && uv sync --frozen
+logprep:py3.11-v19.0.0
 
-Alternative: directly from PyPI via pip:
+To download (pull) these Images an OCI compliant Puller (in this example Docker) can be used like this
```
mhoff:
```suggestion
To download (pull) these images an OCI compliant puller (like docker) can be used like this
```

mhoff (commented):
Some more small remarks

mhoff (approved):
Many thanks for the extensive work!

PR #957, opened 2026-04-17
Fix unhandled luceneerror

IncorrigibleCoffeeCode:
## Description

Fixed unhandled error for filter definitions / rules ending unexpectedly. 
Added tests for those changes, fixed some tests that caused mypy errors.

## Assignee

### Documentation

### Code Quality

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--957.org.readthedocs.build/).

### DIFF
--- a/logprep/util/configuration.py
+++ b/logprep/util/configuration.py
@@ -194,12 +194,13 @@ from __future__ import annotations  # Fixes mypy reporting wrong lines
 import json
 import logging
 import os
+import typing
 from copy import deepcopy
 from importlib.metadata import version
 from itertools import chain
 from logging.config import dictConfig
 from pathlib import Path
-from typing import Any, cast, Iterable, List, Optional, Sequence, Tuple
+from typing import Any, Iterable, List, Optional, Sequence, Tuple
 from attrs import asdict, define, field, fields, validators
 from requests import RequestException
 from ruamel.yaml import YAML
@@ -256,7 +257,7 @@ class InvalidConfigurationErrors(InvalidConfigurationError):
 
     errors: List[InvalidConfigurationError]
 
-    def __init__(self, errors: Sequence[Exception]) -> None:
+    def __init__(self, errors: List[Exception]) -> None:
         unique_errors = []
         for error in errors:
             if not isinstance(error, InvalidConfigurationError):
@@ -1027,8 +1028,8 @@ class Configuration:
         for processor_config in self.pipeline:
             processor = None
             try:
-                processor_component = Factory.create(deepcopy(processor_config))
-                processor = cast(Processor, processor_component)
+                processor = Factory.create(deepcopy(processor_config))
+                processor = typing.cast(processor)
                 processor.setup()
                 self._verify_rules(processor)
             except (
--- a/logprep/util/configuration.py
+++ b/logprep/util/configuration.py
@@ -190,7 +190,6 @@ The following config file will be valid by setting the given environment variabl
                 group.id: test"
 """
 
-from __future__ import annotations  # Fixes mypy reporting wrong lines
 import json
 import logging
 import os
--- a/logprep/util/configuration.py
+++ b/logprep/util/configuration.py
@@ -1030,6 +1030,7 @@ class Configuration:
         for processor_config in self.pipeline:
             processor = None
             try:
+                processor = Factory.create(deepcopy(processor_config))
                 if isinstance(processor, Processor):
                     processor.setup()
                     self._verify_rules(processor)
--- a/logprep/util/configuration.py
+++ b/logprep/util/configuration.py
@@ -200,7 +200,7 @@ from importlib.metadata import version
 from itertools import chain
 from logging.config import dictConfig
 from pathlib import Path
-from typing import Any, Iterable, List, Optional, Sequence, Tuple
+from typing import Any, Iterable, Sequence
 from attrs import asdict, define, field, fields, validators
 from requests import RequestException
 from ruamel.yaml import YAML
@@ -255,9 +255,9 @@ yaml = MyYAML(pure=True)
 class InvalidConfigurationErrors(InvalidConfigurationError):
     """Raise for multiple Configuration related exceptions."""
 
-    errors: List[InvalidConfigurationError]
+    errors: list[InvalidConfigurationError]
 
-    def __init__(self, errors: List[Exception]) -> None:
+    def __init__(self, errors: list[Exception]) -> None:
         unique_errors = []
         for error in errors:
             if not isinstance(error, InvalidConfigurationError):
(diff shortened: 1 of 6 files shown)


logprep/util/configuration.py
```diff
@@ -255,9 +257,11 @@ class InvalidConfigurationErrors(InvalidConfigurationError):
 
     errors: List[InvalidConfigurationError]
 
-    def __init__(self, errors: List[Exception]) -> None:
+    def __init__(self, errors: List) -> None:
```
Pablu23: Please rather than removing specific annotations add them. Was there a reason why you removed the specification here?

logprep/util/configuration.py
```diff
@@ -255,9 +257,11 @@ class InvalidConfigurationErrors(InvalidConfigurationError):
...
         unique_errors = []
         for error in errors:
+            if not isinstance(error, Exception):
+                raise TypeError(f'Error {error} is not of expected type "Exception"')
```
Pablu23: This is probably unneeded if you add back the specification, well atleast it should be but python doesnt have a type system

logprep/util/configuration.py
```diff
@@ -1026,14 +1031,16 @@ def _verify(self) -> None:
             processor = None
             try:
                 processor = Factory.create(deepcopy(processor_config))
-                processor.setup()
-                self._verify_rules(processor)
+                if isinstance(processor, Processor):
+                    processor.setup()
```
Pablu23: What happens if this is not a processor? :P use something like typing.cast here to assert that what we get is a processor, because our world would break apart if this wasnt a processor

tests/unit/util/test_configuration.py
```diff
@@ -52,11 +53,11 @@ class TestConfiguration:
             ("config_refresh_interval", type(None), None),
             ("process_count", int, 1),
             ("timeout", float, 5.0),
-            ("logger", LoggerConfig, LoggerConfig(**{"level": "INFO"})),
+            ("logger", LoggerConfig, LoggerConfig(level="INFO")),
```
Pablu23, line 56: I think your right in doing this and it should be the same, but why was this changed? Was mypy being angry about it?

Pablu23 (changes requested):
Thanks for your work, I left a few Comments after adressing them we can merge this :)

logprep/util/configuration.py
```diff
@@ -257,11 +256,9 @@ class InvalidConfigurationErrors(InvalidConfigurationError):
 
     errors: List[InvalidConfigurationError]
 
-    def __init__(self, errors: List) -> None:
+    def __init__(self, errors: Sequence[Exception]) -> None:
```
Pablu23: This introduces a slight but meaningful difference to before, now we would not allow modification of this list, which might actually be fine, but before changing something like this we should always know exactly why we want to do that. As a little hint, just using lowercase "list" should work here, List and list are two different things

tests/unit/util/test_configuration.py
IncorrigibleCoffeeCode: Yes, mypy really doesn't like unpacking arguments and this seemed to be the cleanest solution

logprep/util/configuration.py
IncorrigibleCoffeeCode: Okay, I'll swap this out for typing.cast. The Type of "processor" is Component|None, which is why mypy didn't like this. Processor inherits from Component, so it should be fine, just needed some kind of confirmation.

logprep/util/configuration.py
```diff
@@ -1030,10 +1027,10 @@ def _verify(self) -> None:
...
             processor = None
             try:
-                processor = Factory.create(deepcopy(processor_config))
-                if isinstance(processor, Processor):
-                    processor.setup()
-                    self._verify_rules(processor)
+                processor_component = Factory.create(deepcopy(processor_config))
+                processor = cast(Processor, processor_component)
+                processor.setup()
+                self._verify_rules(processor)
```
Pablu23, line 1033: Thanks, this works and we can leave it like that, to avoid having processor_component and processor, we can do something called shadowing here, so what this means is we do
```py
processor = Factory.create()
processor = typing.cast()
```
also we nearly everywhere use typing.cast as a qualified name and dont import it, so I would like to keep it that way

logprep/util/configuration.py
IncorrigibleCoffeeCode: Agreed.

logprep/util/configuration.py
IncorrigibleCoffeeCode: mypy was having an issue with the list being invariant.
I now found a better solution by using Sequence instead of List.

Pablu23 (commented):
Thanks for your changes, I left a few more comments, feel free to resolve the conversations which you answered or changed

logprep/util/configuration.py
Pablu23: Yes just fyi, we are already planning on making this function more concrete, but right now factory.create is used for alot of things where you just have to know that the resulting type will be what you requested, thats why typing cast is fine here

logprep/util/configuration.py
IncorrigibleCoffeeCode: Hmm... No, "list" doesn't work either. I already suspected that substituting "List" or "list" with "Sequence" might be problematic, that's why I originally opted for the if-check. The tests don't seem to mind Sequence.
Here's the Error I get (List/list doesn't change anything here):
```
logprep/util/configuration.py:812: error: Argument 1 to "InvalidConfigurationErrors" has incompatible type "list[InvalidConfigurationError]"; expected "list[Exception]"  [arg-type]
logprep/util/configuration.py:812: note: "list" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
logprep/util/configuration.py:812: note: Consider using "Sequence" instead, which is covariant
```
I could also do a # type: ingore, but I guess that should be avoided where possible, too.

Pablu23 (dismissed):
LGTM thank you for the work

logprep/util/configuration.py
```diff
@@ -190,16 +190,17 @@
                 group.id: test"
 """
 
+from __future__ import annotations  # Fixes mypy reporting wrong lines
```
mhoff:
```suggestion
from __future__ import annotations
```

The comment is not required.

FYI, this annotation is deprecated in 3.14 but fine for now as we need to support older versions: https://docs.python.org/3/whatsnew/3.14.html#from-future-import-annotations

IncorrigibleCoffeeCode: I have removed it and there's been no problems for now.

Pablu23 (approved):
Thank you looks good to me

PR #958, opened 2026-04-20
dev: json support httpgetter

kaya-david:
## Description

* Added json support for HttpGetter

## Assignee

### Documentation

### Code Quality

* Added tests to verify json list support

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--958.org.readthedocs.build/).

### DIFF
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -5,8 +5,9 @@ import os
 import re
 from abc import ABC, abstractmethod
 from copy import deepcopy
+from email.headerregistry import ContentTypeHeader
 from string import Template
-from typing import Dict, List, Union
+from typing import Dict, List, TypeAlias, Union
 
 from attrs import define, field, validators
 from ruamel.yaml import YAML, YAMLError
@@ -21,6 +22,11 @@ BLOCKLIST_VARIABLE_NAMES = [
 
 VALID_PREFIXES = ["LOGPREP_", "CI_", "GITHUB_", "PYTEST_"]
 
+# TODO: get config informations for related processor and pass it to the _get method:
+CONTENT_JSON_KEY = "content"
+
+ContentType: TypeAlias = str
+
 
 @define(kw_only=True)
 class Getter(ABC):
@@ -30,13 +36,13 @@ class Getter(ABC):
         """Template class for uppercase only template variables"""
 
         pattern = r"""
-        \$(?:
-            (?P<escaped>\$\$\$)|
-            (?P<named>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)|
-            {(?P<braced>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)}|
-            (?P<invalid>)
-        )
-        """
+            \$(?:
+                (?P<escaped>\$\$\$)|
+                (?P<named>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)|
+                {(?P<braced>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)}|
+                (?P<invalid>)
+            )
+            """  # type: ignore[assignment]
         flags = re.VERBOSE
 
     protocol: str = field(validator=validators.instance_of(str))
@@ -54,10 +60,36 @@ class Getter(ABC):
     )
     """used variables in content but not set in environment"""
 
-    def get(self) -> str:
+    content_type: ContentType | None = None
+    """content/mime type of the underlying content, defaults to None"""
+
+    def get(self) -> str | list:
         """calls the get_raw method, decodes the bytes to string and
         enriches by environment variables.
         """
+
+        match self.content_type:
+            case "application/json":
+                json_content: dict | list = self.get_json()
+
+                if isinstance(json_content, list):
+                    return json_content
+                elif isinstance(json_content, dict):
+                    content = json_content.get(CONTENT_JSON_KEY)
+
+                    if content is None:
+                        raise ValueError(f"Content key not found: '{CONTENT_JSON_KEY}'")
+
+                    return content
+                else:
+                    raise ValueError(f"Expected json dict in Getter content")
+
+            case "text/plain":
+                return self._substitute_raw_content().splitlines()
+            case _:
+                return self._substitute_raw_content()
+
+    def _substitute_raw_content(self):
         content = self.get_raw().decode("utf8")
         template = self.EnvTemplate(content)
         kwargs = self._get_kwargs(template, content)
@@ -81,7 +113,7 @@ class Getter(ABC):
 
     def get_yaml(self) -> Union[Dict, List]:
         """gets and parses the raw content to yaml"""
-        raw = self.get()
+        raw = self._substitute_raw_content()
         parsed_yaml = list(yaml.load_all(raw))
         if not parsed_yaml:
             return {}
@@ -91,7 +123,7 @@ class Getter(ABC):
 
     def get_json(self) -> Union[Dict, List]:
         """gets and parses the raw content to json"""
-        return json.loads(self.get())
+        return json.loads(self._substitute_raw_content())
 
     def get_collection(self) -> Union[Dict, List]:
         """Gets and parses the raw content to yaml or json if yaml fails"""
@@ -117,7 +149,7 @@ class Getter(ABC):
     def get_jsonl(self) -> List:
         """Gets and parses the raw content to jsonl"""
         parsed_events = []
-        for json_string in self.get().splitlines():
+        for json_string in self._substitute_raw_content().splitlines():
             if json_string.strip() != "":
                 event = json.loads(json_string)
                 parsed_events.append(event)
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -5,9 +5,8 @@ import os
 import re
 from abc import ABC, abstractmethod
 from copy import deepcopy
-from email.headerregistry import ContentTypeHeader
 from string import Template
-from typing import Dict, List, TypeAlias, Union
+from typing import TypeAlias
 
 from attrs import define, field, validators
 from ruamel.yaml import YAML, YAMLError
@@ -21,9 +20,6 @@ BLOCKLIST_VARIABLE_NAMES = [
 ]
 
 VALID_PREFIXES = ["LOGPREP_", "CI_", "GITHUB_", "PYTEST_"]
-
-# TODO: get config informations for related processor and pass it to the _get method:
-CONTENT_JSON_KEY = "content"
 
 ContentType: TypeAlias = str
 
@@ -60,36 +56,13 @@ class Getter(ABC):
     )
     """used variables in content but not set in environment"""
 
-    content_type: ContentType | None = None
-    """content/mime type of the underlying content, defaults to None"""
-
-    def get(self) -> str | list:
+    def get(self) -> dict | list | str:
         """calls the get_raw method, decodes the bytes to string and
         enriches by environment variables.
         """
+        return self._substitute_raw_content()
 
-        match self.content_type:
-            case "application/json":
-                json_content: dict | list = self.get_json()
-
-                if isinstance(json_content, list):
-                    return json_content
-                elif isinstance(json_content, dict):
-                    content = json_content.get(CONTENT_JSON_KEY)
-
-                    if content is None:
-                        raise ValueError(f"Content key not found: '{CONTENT_JSON_KEY}'")
-
-                    return content
-                else:
-                    raise ValueError(f"Expected json dict in Getter content")
-
-            case "text/plain":
-                return self._substitute_raw_content().splitlines()
-            case _:
-                return self._substitute_raw_content()
-
-    def _substitute_raw_content(self):
+    def _substitute_raw_content(self) -> str:
         content = self.get_raw().decode("utf8")
         template = self.EnvTemplate(content)
         kwargs = self._get_kwargs(template, content)
@@ -111,21 +84,29 @@ class Getter(ABC):
         used_braced_env_vars = map(lambda x: x[2], found_variables)
         return (item for item in {*used_named_env_vars, *used_braced_env_vars} if item)
 
-    def get_yaml(self) -> Union[Dict, List]:
+    def get_yaml(self) -> dict | list:
         """gets and parses the raw content to yaml"""
-        raw = self._substitute_raw_content()
-        parsed_yaml = list(yaml.load_all(raw))
+        content = self._substitute_raw_content()
+        return self.parse_yaml(content)
+
+    def parse_yaml(self, content: str) -> dict | list:
+        parsed_yaml = list(yaml.load_all(content))
         if not parsed_yaml:
             return {}
         if len(parsed_yaml) > 1:
             return parsed_yaml
         return parsed_yaml.pop()
 
-    def get_json(self) -> Union[Dict, List]:
+    def get_json(self) -> dict | list:
         """gets and parses the raw content to json"""
-        return json.loads(self._substitute_raw_content())
+        content = self._substitute_raw_content()
+        return self.parse_json(content)
 
-    def get_collection(self) -> Union[Dict, List]:
+    def parse_json(self, content: str) -> dict | list:
+        """parses the content to json"""
+        return json.loads(content)
+
+    def get_collection(self) -> dict | list:
         """Gets and parses the raw content to yaml or json if yaml fails"""
         try:
             return self.get_yaml()
@@ -139,21 +120,22 @@ class Getter(ABC):
             raise ValueError("Value is not a dictionary")
         return result
 
-    def get_list(self) -> list:
-        """Gets list and fails otherwise"""
-        result = self.get_collection()
-        if not isinstance(result, list):
-            raise ValueError("Value is not a list")
-        return result
+    def parse_list(self, content: str) -> list:
+        """parses the content to list"""
+        return content.splitlines()
 
-    def get_jsonl(self) -> List:
+    def get_jsonl(self) -> list:
         """Gets and parses the raw content to jsonl"""
         parsed_events = []
-        for json_string in self._substitute_raw_content().splitlines():
-            if json_string.strip() != "":
-                event = json.loads(json_string)
+        for content in self.parse_list(self._substitute_raw_content()):
+            if content.strip() != "":
+                event = self.parse_json(content)
                 parsed_events.append(event)
         return parsed_events
+
+    def parse_jsonl(self, content: str) -> list:
+        """parses the content to jsonl"""
+        return json.loads(content)
 
     @abstractmethod
     def get_raw(self) -> bytes:
--- a/logprep/processor/list_comparison/rule.py
+++ b/logprep/processor/list_comparison/rule.py
@@ -125,8 +125,13 @@ class ListComparisonRule(FieldManagerRule):
             http_getter.add_callback(self._update_compare_sets_via_http, http_getter, list_path)
 
     def _update_compare_sets_via_http(self, http_getter: HttpGetter, list_path: str) -> None:
-        compare_elements: list[str] = http_getter.get()
-        self._compare_sets.update({list_path: set(compare_elements)})
+        compare_elements = http_getter.get()
+
+        if isinstance(compare_elements, str):
+            compare_elements = http_getter.parse_list(compare_elements)
+
+        file_elem_tuples = (elem for elem in compare_elements if not elem.startswith("#"))
+        self._compare_sets.update({list_path: set(file_elem_tuples)})
 
     def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> None:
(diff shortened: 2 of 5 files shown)


logprep/processor/list_comparison/rule.py
```diff
@@ -120,8 +125,10 @@ def _init_list_comparison_from_http(self, list_search_base_path: str) -> None:
             http_getter.add_callback(self._update_compare_sets_via_http, http_getter, list_path)
 
     def _update_compare_sets_via_http(self, http_getter: HttpGetter, list_path: str) -> None:
-        compare_elements = http_getter.get().splitlines()
-        file_elem_tuples = (elem for elem in compare_elements if not elem.startswith("#"))
+        compare_elements: list[str] = http_getter.get_list()
+        file_elem_tuples = (
+            elem for elem in compare_elements if not elem.startswith("#")
```
mhoff: I am not sure if we still require this. In any case, this is rather content-type specific I'd say

logprep/util/getter.py
```diff
@@ -392,6 +395,7 @@ def credentials(self) -> Credentials:
 
     def _get_from_target(self) -> tuple[bytes | None, bool]:
         response = self._do_request()
+        self._headers["Content-Type"] = response.headers["Content-Type"]
```
mhoff: I think `_headers` is rather meant as the headers for _sending_ the request. Putting the response header here is first of all an unexpected side-effect and would affect further requests made through this client (see `_do_request`). I think this information should rather be returned through the `return` statement

logprep/util/getter.py
```diff
@@ -435,6 +439,46 @@ def _handle_http_error(error: requests.exceptions.HTTPError):
...
+        match ct := self._headers.get("Content-Type"):
+            case "application/json":
+                json_content = self.get_json()
+
+                if isinstance(json_content, dict):
+                    content = json_content.get(content_json_key)
+                    if content is None:
+                        raise RefreshableGetterError(f"Content key not found: '{content_json_key}'")
+                else:
+                    raise RefreshableGetterError(f"Expected json dict in HttpGetter content")
```
mhoff: I think this makes sense (and could even be improved with a nested match case?):

```suggestion
                if isinstance(json_content, list):
                    content = json_content
                elif isinstance(json_content, dict):
                    content = json_content.get(content_json_key)
                    if content is None:
                        raise RefreshableGetterError(f"Content key not found: '{content_json_key}'")
                else:
                    raise RefreshableGetterError(f"Expected json dict in HttpGetter content")
```

logprep/util/getter.py
```diff
@@ -435,6 +439,46 @@ def _handle_http_error(error: requests.exceptions.HTTPError):
...
+
+        self._update_cache()
+
+        # TODO: get config informations for related processor and pass it to the _get method:
+        content_json_key = "content"
+
+        return self._get(content_json_key=content_json_key)
+
+    def _get(self, content_json_key: str) -> list[str]:
+        match ct := self._headers.get("Content-Type"):
```
mhoff: I am not sure whether these inline assignments are very readable. I have no strong opinion, so happy to discuss. @Pablu23 feel free to pitch in.

Walrus operator aside, `ct` might be a tad short as a variable name

mhoff (commented):
Many thanks, I have left a few comments

logprep/processor/list_comparison/rule.py
kaya-david: Done. Removed the line.

logprep/util/getter.py
kaya-david: Done. This case is a bit tricky, so I implemented it in a way that minimizes interface changes.

logprep/util/getter.py
kaya-david: Removed by another review fix

logprep/util/getter.py
kaya-david: Done. Added list result support as well.

logprep/util/getter.py
```diff
@@ -26,7 +27,12 @@
     CredentialsEnvNotFoundError,
     CredentialsFactory,
 )
-from logprep.util.defaults import ENV_NAME_LOGPREP_CREDENTIALS_FILE, ENV_NAME_LOGPREP_GETTER_CONFIG
+from logprep.util.defaults import (
+    ENV_NAME_LOGPREP_CREDENTIALS_FILE,
+    ENV_NAME_LOGPREP_GETTER_CONFIG,
+)
+
+ContentType = str
```
mhoff: Explicit annotation as a TypeAlias is [preferred](https://typing.python.org/en/latest/spec/aliases.html#typealias)

```suggestion
ContentType: TypeAlias = str
```

logprep/util/getter.py
```diff
@@ -390,10 +396,11 @@ def credentials(self) -> Credentials:
             creds = CredentialsFactory.from_target(self.uri)
         return creds if creds else Credentials()
 
-    def _get_from_target(self) -> tuple[bytes | None, bool]:
+    def _get_from_target(self) -> tuple[bytes | None, bool, ContentType]:
         response = self._do_request()
+        content_type = response.headers["Content-Type"]
```
mhoff: I wonder if this header is always set. We should either verify that this is the case or handle the edge case as well

logprep/util/getter.py
```diff
@@ -435,6 +442,49 @@ def _handle_http_error(error: requests.exceptions.HTTPError):
...
+        in the corresponding `list_comparison` or `network_comparison`
+        processor configuration.
+
+        See the configuration examples of these processors for details.
+        """
+
+        _, content_type = self._update_cache()
+
+        # TODO: get config informations for related processor and pass it to the _get method:
+        content_json_key = "content"
```
mhoff: Should be a global constant for now I think

logprep/util/getter.py
```diff
@@ -435,6 +442,49 @@ def _handle_http_error(error: requests.exceptions.HTTPError):
...
+
+        return self._get(content_type=content_type, content_json_key=content_json_key)
```
mhoff: Just an optinionated suggestion but I think the keyword-based arguments do not improve readability for this line

```suggestion
        return self._get(content_type, content_json_key)
```

logprep/util/getter.py
```diff
@@ -435,6 +442,49 @@ def _handle_http_error(error: requests.exceptions.HTTPError):
...
+    def get_list(self) -> list[str]:
+        """Get the content as list based on header information.
+
+        This method calls the target URL to retrieve the initial header
+        information and initialize the cache state.
+
+        For the content type `application/json`, it defaults to using the
+        `content` key. This can also be configured via `content_json_key`
+        in the corresponding `list_comparison` or `network_comparison`
+        processor configuration.
```
mhoff: Not true as of now. Should not be in the PR as long as the the other part is not implemented (otherwise we might forget to take it out)

logprep/util/getter.py
```diff
@@ -435,6 +442,49 @@ def _handle_http_error(error: requests.exceptions.HTTPError):
...
+
+        See the configuration examples of these processors for details.
+        """
+
+        _, content_type = self._update_cache()
```
mhoff: I just realized how you hook into the class hierarchy:

- HttpGetter.get_list()
  - HttpGetter._update_cache() -> _get_from_target() -> _do_request() -> session.get()
  - HttpGetter._get() -> get_json() -> get() -> get_raw() -> _update_cache() but with cache boundary -> [... see above...]

This effectively disables caching and does the http request every time `get_list` is called. Also, we are effectively avoiding the problem of modifying the class hierarchy in the way, that the content type is properly passed to the top-level handler.

We should rather modify the abstract getter (and probably the class hierarchy beneath as well) to allow for extraneous information to be passed along the call chain, while keeping the public interfaces intact.

I also wonder why the tests didn't catch that. There is probably one missing for ensuring the cache is actually being used and not circumvented by other implementations

logprep/util/getter.py
```diff
@@ -435,6 +442,49 @@ def _handle_http_error(error: requests.exceptions.HTTPError):
...
+
+        return self._get(content_type=content_type, content_json_key=content_json_key)
+
+    def _get(self, content_type: ContentType, content_json_key: str) -> list[str]:
+        match content_type:
+            case "application/json":
+                json_content: dict | list = self.get_json()
+
+                if isinstance(json_content, list):
+                    return json_content
```
mhoff: Validation for element type str would be nice

logprep/util/getter.py
```diff
@@ -435,6 +442,49 @@ def _handle_http_error(error: requests.exceptions.HTTPError):
...
+                elif isinstance(json_content, dict):
+                    content = json_content.get(content_json_key)
```
mhoff: Validation for object type list and element type str would be nice

mhoff (changes requested):
Hello @kaya-david, I am afraid I have found a challenge in the implementation. Happy to discuss

logprep/util/getter.py
kaya-david: Done

logprep/util/getter.py
kaya-david: Done

logprep/util/getter.py
kaya-david: Done

logprep/util/getter.py
kaya-david: As discussed verbally, I refactored this, so this comment should now be obsolete and the change requires a new review.

logprep/util/getter.py
kaya-david: What do you exactly mean? Is this obsolete?

logprep/util/getter.py
kaya-david: What do you exactly mean? Is this obsolete?

logprep/abc/getter.py
```diff
@@ -54,10 +60,36 @@ class EnvTemplate(Template):
     )
     """used variables in content but not set in environment"""
 
-    def get(self) -> str:
+    content_type: ContentType | None = None
```
mhoff: I would like to keep the class stateless and avoid having separate paths for transporting content and content type. Both are set at the same time, are tightly coupled and should a similar path in the call hierarchy

logprep/abc/getter.py
```diff
@@ -54,10 +60,36 @@ class EnvTemplate(Template):
...
+    """content/mime type of the underlying content, defaults to None"""
+
+    def get(self) -> str | list:
```
mhoff: It's suprising that a `get` method can return `str` or `list` but not a `dict`.

Also, the automagic access of the field `CONTENT_JSON_KEY` is only legitimate if we explicitly want a `list` (`get_list` rather than `get`) and thus navigate inside the `dict`. In other cases this behavior is not warranted nor valid

logprep/processor/list_comparison/rule.py
```diff
@@ -137,8 +141,13 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
         ]
         list_paths = [*absolute_list_paths, *converted_absolute_list_paths]
         for list_path in list_paths:
-            content = GetterFactory.from_string(list_path).get()
-            compare_elements = content.splitlines()
+            content: str | list = GetterFactory.from_string(list_path).get()
+
+            if isinstance(content, str):
+                compare_elements = content.splitlines()
```
mhoff: This would be great to have in a `get_list` inside the getter

logprep/processor/list_comparison/rule.py
```diff
@@ -120,9 +125,8 @@ def _init_list_comparison_from_http(self, list_search_base_path: str) -> None:
             http_getter.add_callback(self._update_compare_sets_via_http, http_getter, list_path)
 
     def _update_compare_sets_via_http(self, http_getter: HttpGetter, list_path: str) -> None:
-        compare_elements = http_getter.get().splitlines()
-        file_elem_tuples = (elem for elem in compare_elements if not elem.startswith("#"))
```
mhoff, line 124: This filtering logic is now gone for the http case, even though we could still get plaintext from the server

logprep/util/getter.py
```diff
@@ -435,6 +441,17 @@ def _handle_http_error(error: requests.exceptions.HTTPError):
             )
         ) from error
 
+    def get(self) -> list[str]:
+        """Get the content as list based on header information."""
+
+        self._update_cache()
```
mhoff: This circumvents the scheduler and triggers an http request on each get

logprep/util/getter.py
mhoff: No this is not obsolete, but also not strictly required.

What I meant is that we could check if all elements of the list are actually strings, as the has returned list[str] in the past

logprep/util/getter.py
mhoff: Same here, is content actually a list?

logprep/util/getter.py
kaya-david: I’ve decided not to do that at this point.

logprep/util/getter.py
kaya-david: I added that check.

logprep/abc/getter.py
kaya-david: I removed that and integrated it into the shared object.

logprep/abc/getter.py
kaya-david: I changed the return type of `Abc.Getter` to `dict | list | str`.

I also removed `HttpGetter.get()` and implemented `RefreshableGetter.get()` instead, as mentioned below in the last review comment: “...The branching logic now lives at the `RefreshableGetter` level, which is also where the shared object with its cache and `content_type` resides.”

Side note: on the `Abc.Getter` level, this may be a bit confusing in the code. However, mypy requires the same annotation here as on the overridden child implementations of `get()`.

I deliberately chose `CONTENT_JSON_KEY`, because in the `elif isinstance(json_content, dict):` case we already have the dict at hand, and implementing a separate method that only performs a `dict[key]` lookup feels like overengineering here.

As mentioned above, I got rid of `get_list` and removed the related tests, because the method never really fit in, was not used anywhere, and only caused confusion.

Instead, there are now `parse_...` methods, which make it possible to avoid duplicate URL requests and continue working directly with content that was already retrieved via get in the child classes, and then parse it further from there (see `RefreshableGetter.get()`).

On the `RefreshableGetter` level, I need a populated cache in order to branch based on the `content_type`. Because of that, I could never knowingly call the parent methods `get_list, get_json`, etc., since at that point I would not yet have the headers available.

So in the code I had to strike a balance between `get_...` and `parse_...`, as you will see.

logprep/processor/list_comparison/rule.py
kaya-david: Done. Implemented `parse_list` instead of `get_list`.

logprep/processor/list_comparison/rule.py
kaya-david: I had removed that because I thought our earlier discussion about those lines was that filtering for `#` was no longer necessary.

I added it back in, since `http_getter.get()` has changed and now returns a list in the tests. To rule out the possibility that it does not return a `str` after all, I will also add a check that falls back to `splitlines()` via the inherited method `Getter.parse_list()`

logprep/util/getter.py
kaya-david: Done. The "branching logic" now lives at the `RefreshableGetter` level, which is also where the shared object with its `cache` and `content_type` resides.

logprep/abc/getter.py
```diff
@@ -5,8 +5,9 @@
 import re
 from abc import ABC, abstractmethod
 from copy import deepcopy
+from email.headerregistry import ContentTypeHeader
```
mhoff: Is this unused?

kaya-david: Yes, unfortunately that was introduced by me in commit https://github.com/fkie-cad/Logprep/pull/958/changes/a47f01be3eaa74b6cb07b7617a6cf90aabb548e3

logprep/util/getter.py
```diff
@@ -318,6 +340,34 @@ def refresh(cls):
...
+                    content = cast(list, json_content.get(CONTENT_JSON_KEY))
+
+                    if content is None:
+                        raise ValueError(f"Content key not found: '{CONTENT_JSON_KEY}'")
+                    if not isinstance(content, list):
+                        raise ValueError(f"Content is not a list.")
+
+                    return content
+                else:
+                    raise ValueError(f"Expected json dict in Getter content")
```
mhoff, line 364:
```suggestion
                        raise ValueError(f"Content is not a list")

                    return content
                else:
                    raise ValueError(f"Expected json object in getter content")
```

logprep/util/getter.py
```diff
@@ -318,6 +340,34 @@ def refresh(cls):
             if shared_target_data.scheduler:
                 shared_target_data.scheduler.run_pending()
 
+    def get(self) -> dict | list | str:
+        """Get the content as list based on header information."""
```
mhoff, line 344: The comment is not correct

logprep/abc/getter.py
```diff
@@ -79,21 +84,29 @@ def _get_used_env_vars(self, content, template):
...
-    def get_yaml(self) -> Union[Dict, List]:
+    def get_yaml(self) -> dict | list:
         """gets and parses the raw content to yaml"""
-        raw = self.get()
-        parsed_yaml = list(yaml.load_all(raw))
+        content = self._substitute_raw_content()
+        return self.parse_yaml(content)
+
+    def parse_yaml(self, content: str) -> dict | list:
+        parsed_yaml = list(yaml.load_all(content))
```
mhoff, line 93: Ale `parse_*` methods are rather staticmethods

logprep/abc/getter.py
```diff
@@ -54,10 +56,13 @@ class EnvTemplate(Template):
     )
     """used variables in content but not set in environment"""
 
-    def get(self) -> str:
+    def get(self) -> dict | list | str:
```
mhoff, line 59: I would prefer to do the content-type dependent deserialization already up here in the abstract getter. This makes the implementation less spread out and easier to understand for other developers. As a developer you statically always use the abstract interface class and then you see an implementation which returns only strings. You obviously could describe what is happening through comments and pydoc, but I would prefer to simply put it here.

I only see one argument against, which is that not each subclass of `Getter` might have content/mime types. Well, we need to handle the `None` case for content types either way (right now via simply returning `super.get()`), so these implementations could simply return an empty content type, if they truly have none.

mhoff, line 59: Maybe call this function `get_parsed` as we automatically resolve env variables and deserialize the format.

logprep/abc/getter.py
```diff
@@ -54,10 +56,13 @@ class EnvTemplate(Template):
...
         """calls the get_raw method, decodes the bytes to string and
         enriches by environment variables.
         """
+        return self._substitute_raw_content()
```
mhoff, line 63: `substitute_raw_content` rather sounds like a method which takes something and returns something. If we stick with this wording, we should rather pass the results of `get_raw()` as a parameter to it. This might be even better then the current solution, because there is less implicit behavior.

Alternatively you could call this method `get_raw_with_resolved_placeholders` or something like this. Yes, that is very verbose and there are for sure better names. The point I am trying to make is rather with the `get_` prefix

logprep/processor/list_comparison/rule.py
```diff
@@ -137,8 +146,16 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
...
-            compare_elements = content.splitlines()
+            getter = GetterFactory.from_string(list_path)
+            content: str | list | dict = getter.get()
+
+            if isinstance(content, str):
+                compare_elements = getter.parse_list(cast(str, content))
+            elif isinstance(content, list):
+                compare_elements = content
+            else:
+                raise TypeError(f"The content must be a string or a list")
```
mhoff, line 157: Be careful with `TypeError` when it involves dynamic data. These kind of errors are seldomly handled gracefully because they are not expected if everything is implemented properly. A `ValueError` seems more reasonable here, as the wrong type stems from a wrong value.

```suggestion
                raise ValueError(f"The content must be a string or a list")
```

mhoff, line 157: This makes a point apparent, which I dislike about making the `parse_*` methods available as static utility methods: they are not connected anymore to the `Getter` and could (or rather should) be moved to another place if we stick with this design. But this also adds complexity to the users of `Getter` we can already see here that conditional parsing logic flows into the calling code which will become a pattern over time.

Imagine you are a processor and want to support data sources for lists of many kinds. Like newline-separated text, json lists, yaml lists or any other common format for lists. Would you implement this functionality in each processor? I actually liked the construct `get_list` which does content-type dependent parsing logic, attempts to convert to a list where reasonable and configured ("content" sub-field) and actually validates that the result is a list or fails otherwise. The calling code becomes very simply and only needs to handle the error case where a list could not be obtained.

Don't get me wrong. I don't argue for magic list conversion logic. Obtaining a list from the above mentioned data formats seems reasonable to me. The automatic sub-field selection ("content" etc.) should _only_ take place if the getter has been configured in this way (e.g. by passing the field name to the Getter itself or the `get_list` method. Happy to discuss

mhoff (commented):
Hello @kaya-david, I have added a few remarks. We are getting there! Happy to discuss

logprep/abc/getter.py
mhoff: I have replied to you in the conversations below

kaya-david:
Clean start by PR: https://github.com/fkie-cad/Logprep/pull/959

PR #959, opened 2026-04-23
feat: add json support to getters + refactoring

kaya-david:
## Description

* add json support for list_comparison by refactoring the getters
* add _get_raw abstract method interface
* refactor getters and adapt public interfaces + introduce private helper methods
* extend cache and handle content_type

## Assignee

### Documentation

### Code Quality

* added test cases

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--959.org.readthedocs.build/).

### DIFF
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -91,17 +91,17 @@ class Getter(ABC):
         """Get content with enriched environment variables parsed based on content type."""
 
         raw, content_type = self._get_raw()
-        resolved_content = self._resolve_content(raw)
+        content = self._resolve_content(raw)
 
         match content_type:
             case "application/json":
-                return self._to_json(resolved_content)
+                return self._to_json(content)
             case "application/yaml":
-                return self._to_yaml(resolved_content)
+                return self._to_yaml(content)
             case "text/plain":
-                return resolved_content
+                return content
             case _:
-                return resolved_content
+                return content
 
     @staticmethod
     def _to_yaml(content: str) -> dict | list:
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -187,9 +187,9 @@ class Getter(ABC):
     def get_jsonl(self) -> list:
         """Gets and parses the raw content to jsonl"""
         parsed_events = []
-        for json_string in self.get().splitlines():
+        for json_string in self._parse_list(self.get()):
             if json_string.strip() != "":
-                event = json.loads(json_string)
+                event = self._parse_json(json_string)
                 parsed_events.append(event)
         return parsed_events
 
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -198,7 +198,7 @@ class Getter(ABC):
         content: dict | list | str = self._get_parsed_content()
 
         if isinstance(content, str):
-            content = content.splitlines()
+            content = self._to_list(content)
 
         match content:
             case dict():
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -136,7 +136,10 @@ class Getter(ABC):
         return json.loads(content)
 
     def get_json(self) -> dict | list:
-        """Gets and parses the raw content as json"""
+        """Gets and parses the raw content as json
+        Content may already be parsed according to its actual content type,
+        e.g. YAML content is returned as-is.
+        """
         content = self._resolve_content_by_content_type()
 
         if isinstance(content, str):
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -1,6 +1,7 @@
 """Module for getter interface"""
 
 import json
+import logging
 import os
 import re
 from abc import ABC, abstractmethod
@@ -11,6 +12,7 @@ from typing import TypeAlias
 from attrs import define, field, validators
 from ruamel.yaml import YAML, YAMLError
 
+logger = logging.getLogger("Getter")
 yaml = YAML(typ="safe", pure=True)
 
 BLOCKLIST_VARIABLE_NAMES = [
@@ -60,8 +62,7 @@ class Getter(ABC):
     def get(self) -> str:
         """Returns content with enriched environment variables."""
         raw, _ = self._get_raw()
-        content = self._resolve_content(raw)
-        return content
+        return self._resolve_content(raw)
 
     def _resolve_content(self, raw_content: bytes) -> str:
         content = raw_content.decode("utf8")
@@ -93,14 +94,13 @@ class Getter(ABC):
 
         match content_type:
             case "application/json":
-                json_content: dict | list = self._parse_json(content)
-                return json_content
+                return self._parse_json(content)
             case "application/yaml":
-                yaml_content: dict | list = self._parse_yaml(content)
-                return yaml_content
-            case "text/plain":
+                return self._parse_yaml(content)
+            case "text/plain" | None:
                 return content
             case _:
+                logger.warning("Unexpected content type.")
                 return content
 
     @staticmethod
@@ -117,7 +117,12 @@ class Getter(ABC):
         return parsed_yaml.pop()
 
     def get_yaml(self) -> dict | list:
-        """Gets and parses the raw content to yaml"""
+        """Gets and parses the raw content as yaml
+
+        Content may already be parsed according to its actual content type,
+        e.g. JSON content is returned as-is.
+        """
+
         content = self._resolve_content_by_content_type()
 
         if isinstance(content, str):
@@ -174,9 +179,6 @@ class Getter(ABC):
 
         if isinstance(content, str):
             content = self._parse_list(content)
-
-            if not content:
-                return []
 
         match content:
             case list():
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -1,19 +1,17 @@
 """Module for getter interface"""
 
 import json
-import logging
 import os
 import re
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from string import Template
-from typing import TypeAlias, cast
+from typing import TypeAlias
 
 from attrs import define, field, validators
 from ruamel.yaml import YAML, YAMLError
 
 yaml = YAML(typ="safe", pure=True)
-logger = logging.getLogger("Getter")
 
 BLOCKLIST_VARIABLE_NAMES = [
     "",
@@ -176,7 +174,6 @@ class Getter(ABC):
 
         if isinstance(content, str):
             content = self._parse_list(content)
-            # content = self._try_parse(content)
 
             if not content:
                 return []
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -105,7 +105,7 @@ class Getter(ABC):
 
     @staticmethod
     def _parse_yaml(content: str) -> dict | list:
-        """Converts yaml content to dict or json.
+        """Parse yaml content to dict or json.
 
         Note: if parsed_yaml is empty, an empty dict will be returned."""
 
@@ -132,12 +132,12 @@ class Getter(ABC):
 
     @staticmethod
     def _parse_json(content: str) -> dict | list:
-        """Converts content to json"""
+        """Parse content to json"""
         return json.loads(content)
 
     def get_json(self) -> dict | list:
-        """Gets and parses the raw content to json"""
-        content: dict | list | str = self._resolve_content_by_content_type()
+        """Gets and parses the raw content as json"""
+        content = self._resolve_content_by_content_type()
 
         if isinstance(content, str):
             content = self._parse_json(content)
@@ -149,12 +149,12 @@ class Getter(ABC):
         content = self._resolve_content_by_content_type()
 
         if isinstance(content, str):
-            content = self._try_parse(content)
+            content = self._parse_yaml_or_json(content)
 
         return content
 
-    def _try_parse(self, content: str) -> dict | list:
-        """Helper which tries to convert content to list or dict"""
+    def _parse_yaml_or_json(self, content: str) -> dict | list:
+        """Helper which tries to parse content to list or dict"""
         try:
             return self._parse_yaml(content)
         except YAMLError:
@@ -168,17 +168,17 @@ class Getter(ABC):
         return result
 
     @staticmethod
-    def _parse_list(content: str) -> list:
+    def _parse_newline_seperated_list(content: str) -> list:
         """Helper which tries to convert content to list"""
         return content.splitlines()
 
     def get_list(self) -> list:
         """Gets list and fails otherwise"""
 
-        content: dict | list | str = self._resolve_content_by_content_type()
+        content = self._resolve_content_by_content_type()
 
         if isinstance(content, str):
-            content = self._parse_list(content)
+            content = self._parse_newline_seperated_list(content)
 
         match content:
             case list():
@@ -187,9 +187,9 @@ class Getter(ABC):
                 raise ValueError("Content is not a list")
 
     def get_jsonl(self) -> list:
-        """Gets and parses the raw content to jsonl"""
+        """Gets and parses the raw content as jsonl"""
         parsed_events = []
-        for json_string in self._parse_list(self.get()):
+        for json_string in self._parse_newline_seperated_list(self.get()):
             if json_string.strip() != "":
                 event = self._parse_json(json_string)
                 parsed_events.append(event)
@@ -206,12 +206,5 @@ class Getter(ABC):
         """
 
     def get_raw(self) -> bytes:
-        """Get the content.
-
-        Returns
-        -------
-        str
-            The raw serialized content.
-        """
-
+        """Get the raw content."""
         return self._get_raw()[0]
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -6,12 +6,12 @@ import re
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from string import Template
-from typing import Dict, List, Union
+from typing import TypeAlias
 
 from attrs import define, field, validators
 from ruamel.yaml import YAML, YAMLError
 
-yaml = YAML(typ="safe", pure=True)
+yaml = YAML(pure=True, typ="safe")
 
 BLOCKLIST_VARIABLE_NAMES = [
     "",
@@ -20,6 +20,11 @@ BLOCKLIST_VARIABLE_NAMES = [
 ]
 
 VALID_PREFIXES = ["LOGPREP_", "CI_", "GITHUB_", "PYTEST_"]
+
+# TODO: get config informations for related processor
+CONTENT_JSON_KEY = "content"
+
+ContentType: TypeAlias = str
 
 
 @define(kw_only=True)
@@ -30,13 +35,14 @@ class Getter(ABC):
         """Template class for uppercase only template variables"""
 
         pattern = r"""
-        \$(?:
-            (?P<escaped>\$\$\$)|
-            (?P<named>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)|
-            {(?P<braced>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)}|
-            (?P<invalid>)
-        )
-        """
+            \$(?:
+                (?P<escaped>\$\$\$)|
+                (?P<named>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)|
+                {(?P<braced>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)}|
+                (?P<invalid>)
+            )
+        """  # type: ignore[assignment]
+
         flags = re.VERBOSE
 
     protocol: str = field(validator=validators.instance_of(str))
@@ -55,10 +61,12 @@ class Getter(ABC):
     """used variables in content but not set in environment"""
 
     def get(self) -> str:
-        """calls the get_raw method, decodes the bytes to string and
-        enriches by environment variables.
-        """
-        content = self.get_raw().decode("utf8")
+        """Returns content with enriched environment variables."""
+        raw = self.get_raw()
+        return self._resolve_content(raw)
+
+    def _resolve_content(self, raw_content: bytes) -> str:
+        content = raw_content.decode("utf8")
         template = self.EnvTemplate(content)
         kwargs = self._get_kwargs(template, content)
         return template.safe_substitute(**kwargs)
@@ -79,26 +87,66 @@ class Getter(ABC):
         used_braced_env_vars = map(lambda x: x[2], found_variables)
         return (item for item in {*used_named_env_vars, *used_braced_env_vars} if item)
 
-    def get_yaml(self) -> Union[Dict, List]:
-        """gets and parses the raw content to yaml"""
-        raw = self.get()
-        parsed_yaml = list(yaml.load_all(raw))
+    def _get_parsed_content(self) -> dict | list | str:
+        """Get content with enriched environment variables parsed based on content type."""
+
+        raw, content_type = self._get_raw()
+        resolved_content = self._resolve_content(raw)
+
+        match content_type:
+            case "application/json":
+                return self._to_json(resolved_content)
+            case "application/yaml":
+                return self._to_yaml(resolved_content)
+            case "text/plain":
+                return resolved_content
+            case _:
+                return resolved_content
+
+    @staticmethod
+    def _to_yaml(content: str) -> dict | list:
+        """Converts content to json"""
+        parsed_yaml = list(yaml.load_all(content))
         if not parsed_yaml:
             return {}
         if len(parsed_yaml) > 1:
             return parsed_yaml
         return parsed_yaml.pop()
 
-    def get_json(self) -> Union[Dict, List]:
-        """gets and parses the raw content to json"""
-        return json.loads(self.get())
+    def get_yaml(self) -> dict | list:
+        """Gets and parses the raw content to yaml"""
+        content = self._get_parsed_content()
 
-    def get_collection(self) -> Union[Dict, List]:
+        if isinstance(content, dict | list):
+            return content
+
+        return self._to_yaml(content)
+
+    @staticmethod
+    def _to_json(content: str) -> dict | list:
+        """Converts content to json"""
+        return json.loads(content)
+
+    def get_json(self) -> dict | list:
+        """gets and parses the raw content to json"""
+        return self._to_json(self.get())
+
+    def get_collection(self) -> dict | list:
         """Gets and parses the raw content to yaml or json if yaml fails"""
+        content = self._get_parsed_content()
+
+        if isinstance(content, str):
+            # handles unknown content_type or text/plain
+            content = self._try_parse(content)
+
+        return content
+
+    def _try_parse(self, content: str) -> dict | list:
+        """Helper which tries to convert content to list or dict"""
         try:
-            return self.get_yaml()
+            return self._to_yaml(content)
         except YAMLError:
-            return self.get_json()
+            return self._to_json(content)
 
     def get_dict(self) -> dict:
         """Gets dict and fails otherwise"""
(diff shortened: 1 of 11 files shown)


logprep/abc/getter.py
```diff
@@ -54,11 +61,13 @@ class EnvTemplate(Template):
...
 
-    def get(self) -> str:
-        """calls the get_raw method, decodes the bytes to string and
-        enriches by environment variables.
-        """
-        content = self.get_raw().decode("utf8")
+    def get(self) -> dict | list | str:
+        """Returns content with enriched environment variables."""
+        content = self._get_parsed_content()
+        return content
```
mhoff: Does it make sense to break the interface here? This might break some processor implementations and lead to test failures. Maybe these processor implementations are outdated and should be improved, but I would rather define that out-of-scope for this PR.

We could simply return `raw, _ = self._get_raw(); return self._resolve_content(raw)` and be done with it, I think

logprep/abc/getter.py
```diff
@@ -79,26 +88,101 @@ def _get_used_env_vars(self, content, template):
...
+    def _get_parsed_content(self) -> dict | list | str:
+        """Get content with enriched environment variables parsed based on content type."""
+
+        raw, content_type = self._get_raw()
+        content = self._resolve_content(raw)
+
+        match content_type:
+            case "application/json":
+                if isinstance(content, dict | list):
+                    return content
```
mhoff: How can this be the case? `_resolve_content` returns a `str`

logprep/abc/getter.py
```diff
@@ -79,26 +88,101 @@ def _get_used_env_vars(self, content, template):
...
+
+                json_content: dict | list = self._to_json(content)
+
+                if isinstance(json_content, dict) and CONTENT_FIELD is not None:
+                    if CONTENT_FIELD in json_content:
+                        json_content = json_content[CONTENT_FIELD]
+                    else:
+                        raise ValueError(f"Field '{CONTENT_FIELD}' not found in content.")
```
mhoff: I would still tend to only do this if a list has explicitly been requested (and not in the general `_get_parsed_content` utility method). But I would stick with `CONTENT_FIELD = None` for now (as you have it), so this functionality is a stub for the proper implementation in another PR.

logprep/abc/getter.py
```diff
@@ -79,26 +88,101 @@ def _get_used_env_vars(self, content, template):
...
+                if isinstance(content, dict | list):
+                    return content
+
+                yaml_content: dict | list = self._to_yaml(content)
+
+                if isinstance(yaml_content, dict) and CONTENT_FIELD is not None:
+                    if CONTENT_FIELD in yaml_content:
+                        yaml_content = yaml_content[CONTENT_FIELD]
+                    else:
+                        raise ValueError(f"Field '{CONTENT_FIELD}' not found in content.")
```
mhoff: See the two comments above

logprep/abc/getter.py
```diff
@@ -79,26 +88,101 @@ def _get_used_env_vars(self, content, template):
...
+
+                return yaml_content
+            case "text/plain":
+                if isinstance(content, str):
+                    return content
```
mhoff: It can only be `str` here, as `_to_resolve_content` only returns `str`

logprep/abc/getter.py
```diff
@@ -79,26 +88,101 @@ def _get_used_env_vars(self, content, template):
...
+                raise ValueError(
+                    f"Content type '{content_type}' does not match with parsed content"
+                )
+            case _:
+                if isinstance(content, str):
+                    return content
+                raise ValueError(
+                    f"Content type not set. Expecting a simple str (decoded raw content from bytes)"
+                )
```
mhoff: Same as above

logprep/abc/getter.py
```diff
@@ -107,23 +191,62 @@ def get_dict(self) -> dict:
...
 
-    def get_jsonl(self) -> List:
+        content = self._get_parsed_content()
+
+        if isinstance(content, str):
+            content = self._try_parse(content)
+
+        match content:
+            case dict():
+                any_content = content.get(CONTENT_FIELD)
```
mhoff: `CONTENT_FIELD is None` currently, which is why this will fail. The failure is fine as a `list` has been requested and there is a `dict`, but the error message is improper.

logprep/abc/getter.py
```diff
@@ -107,23 +191,62 @@ def get_dict(self) -> dict:
...
-                parsed_events.append(event)
+        content = self._get_parsed_content()
+
+        if isinstance(content, str):
+            for content in self._to_list(content):
+                if content.strip() != "":
+                    event = self._to_json(content)
+                    parsed_events.append(event)
+        elif isinstance(content, list | dict):
+            parsed_events.append(content)
```
mhoff: I think this implementation would allow for a json list of jsons which is not proper jsonl
```
"[ \"{ \\"key\\": \\"abc\\" }\" ]"
```

We should rather do `splitlines` on the resolved (not parsed) content and then process each item with `_to_json`

mhoff (commented):
Hello @kaya-david I have added a few remarks

logprep/abc/getter.py
kaya-david: The following test, which passes, shows the behavior I would expect for JSONL: one JSON object per line, parsed independently.

Is this the behavior you are referring to?

```python
@mock.patch.dict("os.environ", {"PYTEST_TEST_TOKEN": "mytoken"})
def test_get_jsonl_parses_json_object_per_line(self, tmp_path):
    testfile = tmp_path / "test_getter.jsonl"
    testfile.write_text(
        '{"key": "abc"}\n{"key": "def"}\n'
    )

    my_getter = GetterFactory.from_string(str(testfile))

    assert my_getter.get_jsonl() == [
        {"key": "abc"},
        {"key": "def"},
    ]

Comment (kaya-david) [reply logprep/abc/getter.py]: I added another test case that also covers your example:

```python
@pytest.mark.parametrize(
    ("content", "expected"),
    [
        ('[ "{ \\"key\\": \\"abc\\" }" ]', [['{ "key": "abc" }']]),
        (
            '{"key": "abc"}\n{"key": "def"}\n',
            [{"key": "abc"}, {"key": "def"}],
        ),
    ],
)
@mock.patch.dict("os.environ", {"PYTEST_TEST_TOKEN": "mytoken"})
def test_get_jsonl_parses_json_object_per_line(self, tmp_path, content, expected):
    testfile = tmp_path / "test_getter.jsonl"
    testfile.write_text(content)

    my_getter = GetterFactory.from_string(str(testfile))

    assert my_getter.get_jsonl() == expected
```

This shows that your case is currently parsed as one JSONL line containing a JSON list with a JSON string inside, while proper JSONL with one object per line is parsed into separate events.

mhoff: > The following test, which passes, shows the behavior I would expect for JSONL: one JSON object per line, parsed independently.
> 
> Is this the behavior you are referring to?
> 
> ```python
> @mock.patch.dict("os.environ", {"PYTEST_TEST_TOKEN": "mytoken"})
> def test_get_jsonl_parses_json_object_per_line(self, tmp_path):
>     testfile = tmp_path / "test_getter.jsonl"
>     testfile.write_text(
>         '{"key": "abc"}\n{"key": "def"}\n'
>     )
> 
>     my_getter = GetterFactory.from_string(str(testfile))
> 
>     assert my_getter.get_jsonl() == [
>         {"key": "abc"},
>         {"key": "def"},
>     ]
> ```

This is how it should be, yes.

But what I meant to say is that I suspect the above implementation _also_ allows for a json list of serialized jsons. An easier example might be `'["[1, 2, 3]", "[3, 4, 5]"]'`. I think this would be parsed as `[[1, 2, 3], [3, 4, 5]]`, even though it is not JSONL

logprep/abc/getter.py
```diff
@@ -79,26 +91,94 @@ def _get_used_env_vars(self, content, template):
...
+                    if CONTENT_FIELD in yaml_content:
+                        yaml_content = yaml_content[CONTENT_FIELD]
+                    else:
+                        raise ValueError(f"Field '{CONTENT_FIELD}' not found in content.")
+
+                return yaml_content
+            case "text/plain":
+                return content
+            case _:
+                logger.warning("Unexpected content type.")
```
mhoff, line 103: `FileGetter` would always produce this warning. Not sure if this is helpful

logprep/abc/getter.py
kaya-david: Done

logprep/abc/getter.py
kaya-david: Code has changed a little in the meantime. Maybe we should get together specifically about this point if you see the need.

At the moment, I generally see a need to take content_field into account:

- None means: use the content as it is.
- If content_field is set: extract the relevant part from the content and process only that part further.

Happy to discuss this verbally as well.

logprep/abc/getter.py
kaya-david: Done.

logprep/abc/getter.py
kaya-david: Done.

logprep/abc/getter.py
kaya-david: Should be done, or we may need to clarify it verbally.

logprep/abc/getter.py
kaya-david: Done.

logprep/abc/getter.py
```diff
@@ -79,26 +91,94 @@ def _get_used_env_vars(self, content, template):
...
+                json_content: dict | list = self._to_json(content)
+
+                if CONTENT_FIELD is not None:
+                    if isinstance(json_content, list):
+                        raise ValueError(f"'{CONTENT_FIELD}' set, but content is a list.")
+
+                    if CONTENT_FIELD in json_content:
+                        json_content = json_content[CONTENT_FIELD]
+                    else:
+                        raise ValueError(f"Field '{CONTENT_FIELD}' not found in content.")
```
mhoff: I think we should remove this functionality from the `_get_parsed_content` method and only apply the `CONTENT_FIELD` logic in higher-level methods like `get_list`. I'd actually even remove `CONTENT_FIELD` from this PR completely, I think this functionality should either be implemented properly from end to end or not at all. A proper implementation would include passing the `CONTENT_FIELD` from outside and/or making it configurable via `Configuration`. 

Apart from this general remark, the concrete problem with the code in `_get_parsed_content` is that if the constant `CONTENT_FIELD` were set, the high-level API method `get_list` would not work with contents of type application/json which are plain json lists (`"[1,2,3]"`). The code would raise a `ValueError` as it does not expect the parsed content to be a list already.

logprep/abc/getter.py
```diff
@@ -79,26 +91,94 @@ def _get_used_env_vars(self, content, template):
...
+                yaml_content: dict | list = self._to_yaml(content)
+
+                if CONTENT_FIELD is not None:
+                    if isinstance(yaml_content, list):
+                        raise ValueError(f"'{CONTENT_FIELD}' set, but content is a list.")
+
+                    if CONTENT_FIELD in yaml_content:
+                        yaml_content = yaml_content[CONTENT_FIELD]
+                    else:
+                        raise ValueError(f"Field '{CONTENT_FIELD}' not found in content.")
```
mhoff: Same here, see above

logprep/abc/getter.py
```diff
@@ -79,26 +91,94 @@ def _get_used_env_vars(self, content, template):
...
+                logger.warning("Unexpected content type.")
+                return content
+
+    @staticmethod
+    def _to_yaml(content: str) -> dict | list:
+        """Converts yaml content to dict or json."""
+        parsed_yaml = list(yaml.load_all(content))
         if not parsed_yaml:
-            return {}
+            raise YAMLError("Expected content to be a yaml object.")
```
mhoff: I like this semantic change -- we just need to make sure that this does not lead to unexpected behavior

logprep/abc/getter.py
```diff
@@ -79,26 +91,94 @@ def _get_used_env_vars(self, content, template):
...
+        return json.loads(content)
+
+    def get_json(self) -> dict | list:
+        """Gets and parses the raw content to json"""
+        content = self._get_parsed_content()
+
+        if isinstance(content, dict | list):
+            return content
+
+        return self._to_json(content)
```
mhoff: Maybe rather use `_parse_json` instead of `_to_json`. The parameter of the method is expected to be serialized json and that makes the naming counter intuitive.

logprep/abc/getter.py
```diff
@@ -107,23 +187,68 @@ def get_dict(self) -> dict:
...
+                        raise ValueError(f"Field '{CONTENT_FIELD}' not found in content.")
+
+                if isinstance(content, list):
+                    return content
+
+                if isinstance(content, str):
+                    try:
+                        content = self._try_parse(content)
+                        return content
+                    except Exception as ex:
```
mhoff: That's a very broad catch

logprep/abc/getter.py
```diff
@@ -107,23 +187,68 @@ def get_dict(self) -> dict:
...
+                        raise ValueError("Content is not a list") from ex
```
mhoff: The error wording is weird as the result of `_try_parse` is either a `dict` or a `list` or thrown exceptions. So we raise an error that `content` does not represent a `list` in the error case, but we silently pass through a `dict` without further validation.

I'd say either report "could not be parsed" (instead of "is not a list") or also validate that the output is actually a list (which I think we should)

logprep/abc/getter.py
mhoff: Nevermind my above comment, I have reviewed the changed code. What remains is, that yaml content would be allowed as jsonl:

   ```
    @responses.activate
    def test_get_jsonl_treats_json_list_as_one_item(self):
        http_getter: HttpGetter = GetterFactory.from_string("http://some/test/path")

        responses.add(
            responses.GET,
            "http://some/test/path",
            "abc: xyz",
            content_type="application/yaml",
        )
        content = http_getter.get_jsonl() # should raise instead
        assert len(content) == 1 
```

**Review by mhoff — COMMENTED** [review]
Hello @kaya-david, I did another review pass

File: logprep/abc/getter.py [hunk 16]
```diff
@@ -107,23 +162,47 @@ def get_dict(self) -> dict:
...
-        return result
 
-    def get_jsonl(self) -> List:
+        content: dict | list | str = self._resolve_content_by_content_type()
+
+        if isinstance(content, str):
+            content = self._parse_list(content)
+
+            if not content:
+                return []
```
mhoff: Why do we need this? Can this even happen?

logprep/abc/getter.py
```diff
@@ -55,10 +58,13 @@ class EnvTemplate(Template):
...
 
     def get(self) -> str:
-        """calls the get_raw method, decodes the bytes to string and
-        enriches by environment variables.
-        """
-        content = self.get_raw().decode("utf8")
+        """Returns content with enriched environment variables."""
+        raw, _ = self._get_raw()
+        content = self._resolve_content(raw)
+        return content
```
mhoff: Why not immediately return and save the extra line? Do you find that more readable? Some more assignments below follow this pattern.

I would prefer the condensed variant, as the method name (`_resolve_content`) should already speak for itself and if we need the variable name to clarify what is happening, maybe we should improve the name of the helper method instead. Just my opinion and only a recommendation.

logprep/abc/getter.py
```diff
@@ -79,26 +85,75 @@ def _get_used_env_vars(self, content, template):
...
+    def _resolve_content_by_content_type(self) -> dict | list | str:
+        """Get content with enriched environment variables parsed based on content type."""
+
+        raw, content_type = self._get_raw()
+        content = self._resolve_content(raw)
+
+        match content_type:
+            case "application/json":
+                json_content: dict | list = self._parse_json(content)
+                return json_content
```
mhoff: Similar case as above but here with type annotation. Why not simply return the result of `_parse_json` immediately? Because of readability?

logprep/abc/getter.py
```diff
@@ -79,26 +85,75 @@ def _get_used_env_vars(self, content, template):
...
+            case "application/yaml":
+                yaml_content: dict | list = self._parse_yaml(content)
+                return yaml_content
+            case "text/plain":
+                return content
+            case _:
+                return content
```
mhoff:
```suggestion
            case "text/plain" | None:
                return content
            case _:
                # add warning here, as content_type is not None but unknown
                return content
```

logprep/abc/getter.py
```diff
@@ -79,26 +85,75 @@ def _get_used_env_vars(self, content, template):
...
             return {}
         if len(parsed_yaml) > 1:
             return parsed_yaml
         return parsed_yaml.pop()
 
-    def get_json(self) -> Union[Dict, List]:
-        """gets and parses the raw content to json"""
-        return json.loads(self.get())
+    def get_yaml(self) -> dict | list:
+        """Gets and parses the raw content to yaml"""
```
mhoff:
```suggestion
        """Gets and parses the raw content as yaml"""
```

Maybe also document the edge case that json content with content_type "application/json" would be parsed and returned without problems, which might be unexpected for a `get_yaml` method. Same for `get_json`.

logprep/abc/getter.py
```diff
@@ -107,23 +162,47 @@ def get_dict(self) -> dict:
             raise ValueError("Value is not a dictionary")
         return result
 
+    @staticmethod
+    def _parse_list(content: str) -> list:
```
mhoff: In lack of a better name:

```suggestion
    def _parse_newline_separated_list(content: str) -> list:
```

Other methods (like `_parse_json`) parse from a specified source format which is in the name (=json). `_parse_list` on the other hand does not specify such a format. Just a suggestion

logprep/abc/getter.py
```diff
@@ -107,23 +162,47 @@ def get_dict(self) -> dict:
...
+            case list():
+                return content
+            case _:
+                raise ValueError("Content is not a list")
+
+    def get_jsonl(self) -> list:
         """Gets and parses the raw content to jsonl"""
         parsed_events = []
-        for json_string in self.get().splitlines():
+        for json_string in self._parse_list(self.get()):
```
mhoff: Maybe rather don't reuse the public method `get()` and use `_get_raw` and `_resolve_content` instead. Otherwise an API change in one method does affect another API method which is unrelated at first glance

logprep/abc/getter.py
```diff
@@ -132,3 +211,5 @@ def get_raw(self) -> bytes:
         str
             The raw serialized content.
```
mhoff: Maybe change pydoc that we return the content **as raw bytes**. I think one sentence is enough and the `Returns` block might be unnecessary. Your call

logprep/processor/list_comparison/rule.py
```diff
@@ -138,7 +143,14 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
...
         for list_path in list_paths:
             content = GetterFactory.from_string(list_path).get()
-            compare_elements = content.splitlines()
+
+            if isinstance(content, str):
+                compare_elements = content.splitlines()
+            elif isinstance(content, list):
+                compare_elements = content
+            else:
+                raise ValueError("Expected list or str as content")
```
mhoff: `get()` can only return `str` and thus these two cases do not make sense. Maybe rather use `get_list()`?

logprep/util/getter.py
```diff
@@ -11,22 +11,25 @@
 from importlib.metadata import version
 from pathlib import Path
 from string import Template
-from typing import Tuple, ClassVar
+from typing import ClassVar, Tuple
```
mhoff: `tuple`

tests/unit/processor/list_comparison/test_list_comparison.py
```diff
@@ -287,6 +287,37 @@ def test_list_comparison_loads_rule_with_http_template_in_list_search_base_path(
...
+            content_type="application/json",
+        )
+        rule_dict = {
+            "filter": "user",
+            "list_comparison": {
+                "source_fields": ["user"],
+                "target_field": "user_results",
+                "list_file_paths": ["bad_users.list"],
+            },
+            "description": "",
```
mhoff, line 305: `description` can be omitted I think

tests/unit/processor/list_comparison/test_list_comparison.py
```diff
@@ -287,6 +287,37 @@ def test_list_comparison_loads_rule_with_http_template_in_list_search_base_path(
...
+            "type": "list_comparison",
+            "rules": [],
+            "list_search_base_path": "http://localhost:8080/v2/valuestore/test_4",
+        }
+
+        HttpGetter._shared.clear()
+
+        processor = Factory.create({"custom_lister": config})
+        rule = processor.rule_class.create_from_dict(rule_dict)
+        processor._rule_tree.add_rule(rule)
```
mhoff, line 317: Why not directly put `rule_dict` in `rules` and avoid the manual rule setup?

tests/unit/test_run_logprep.py
```diff
@@ -238,7 +238,7 @@ def test_logprep_exits_on_any_exception_during_verify(self):
             assert result.exit_code == 1
 
     def test_logprep_exits_on_request_exception(self):
-        with mock.patch("logprep.util.getter.HttpGetter.get_raw") as mock_verify:
+        with mock.patch("logprep.util.getter.HttpGetter._get_raw") as mock_verify:
```
mhoff: Maybe rather call the mock `mock_get_raw`? I don't see how `mock_verify` makes sense as a name

tests/unit/test_run_ng.py
```diff
@@ -177,7 +177,7 @@ def test_logprep_exits_on_any_exception_during_verify(self):
             assert result.exit_code == 1
 
     def test_logprep_exits_on_request_exception(self):
-        with mock.patch("logprep.util.getter.HttpGetter.get_raw") as mock_verify:
+        with mock.patch("logprep.util.getter.HttpGetter._get_raw") as mock_verify:
```
mhoff: Same here

tests/unit/util/test_getter.py
```diff
@@ -1310,19 +1364,26 @@ def test_refresh_while_with_refreshable_getter_error_logs(self, caplog):
...
         with pytest.raises(ValueError, match="HttpGetter cache is empty"):
             http_getter._update_cache()
 
-    @mock.patch("logprep.abc.getter.Getter.get_yaml", side_effect=YAMLError)
-    @mock.patch("logprep.abc.getter.Getter.get_json")
-    def test_get_collection_parses_json_if_yaml_fails(self, mock_get_json, _):
+    @mock.patch("logprep.abc.getter.Getter._parse_yaml", side_effect=YAMLError)
+    @mock.patch("logprep.abc.getter.Getter._parse_json")
+    @responses.activate
+    def test_get_collection_parses_json_if_yaml_fails(self, mock_to_json, _):
```
mhoff: mock_to_json --> mock_parse_json, why does this even work?

mhoff (commented):
I did another review pass

logprep/abc/getter.py
kaya-david: I originally had a logging.warning here, but you wanted it removed. What should be done here now?

tests/unit/test_run_logprep.py
kaya-david: Done.

tests/unit/util/test_getter.py
kaya-david: Done.

logprep/abc/getter.py
mhoff: Before you had a warning on `case _` which included the `None` case. My suggestion merges `None` with "text/plain" allowing us to address the case which merits a warning.

Code [before](https://github.com/fkie-cad/Logprep/pull/959/commits/6e1117474f780f613e6e4f3607edb9a4a44921a7):

```
                return yaml_content
            case "text/plain":
                return content
            case _:
                logger.warning("Unexpected content type.")
                return content
```

tests/unit/test_run_ng.py
kaya-david: Done.

logprep/abc/getter.py
kaya-david: I see, I had overlooked `None`. That makes sense.

kaya-david: Done.

tests/unit/processor/list_comparison/test_list_comparison.py
kaya-david: Like in many other test cases, I copied the same structure to keep it consistent. This pattern appears in quite a few tests.

tests/unit/processor/list_comparison/test_list_comparison.py
kaya-david: Same here, I based this on the existing tests.

logprep/abc/getter.py
kaya-david: I agree that in this case the helper method name is clear enough, so the extra variable does not add much. I’ll simplify this to `return _resolve_content(raw)`

logprep/abc/getter.py
kaya-david: Done.

logprep/abc/getter.py
kaya-david: I see, I had overlooked `None`. That makes sense.

logprep/abc/getter.py
kaya-david: That’s true, this is not necessary. I had the case where specific inputs parsed an empty string to `{}` within `try_parse`, but that is no longer relevant for our current use cases.

kaya-david: Done.

logprep/abc/getter.py
```diff
@@ -79,26 +86,79 @@ def _get_used_env_vars(self, content, template):
...
+
+    @staticmethod
+    def _parse_json(content: str) -> dict | list:
+        """Parse content to json"""
+        return json.loads(content)
 
-    def get_collection(self) -> Union[Dict, List]:
-        """Gets and parses the raw content to yaml or json if yaml fails"""
+    def get_json(self) -> dict | list:
+        """Gets and parses the raw content as json"""
```
mhoff, line 131:
```suggestion
        """Gets and parses the raw content as json
        Content may already be parsed according to its actual content type,
        e.g. YAML content is returned as-is.
        """
```

Same as `get_yaml`

tests/unit/util/test_getter.py
```diff
@@ -1470,3 +1549,61 @@ def test_refresh_getters(self, tmp_path):
...
+    ):
+        responses.add(
+            responses.GET,
+            "http://something",
+            body=body,
+            content_type=content_type,
+        )
+
+        http_getter: HttpGetter = GetterFactory.from_string("http://something")
+        assert http_getter._resolve_content_by_content_type() == expected
```
mhoff, line 1609: Please add CHANGELOG entries

mhoff (commented):
Hey @kaya-david, just one pydoc remark and please add an CHANGELOG entry

mhoff (dismissed):
LGTM!

logprep/abc/getter.py
```diff
@@ -79,26 +86,75 @@ def _get_used_env_vars(self, content, template):
...
 
-    def get_json(self) -> Union[Dict, List]:
-        """gets and parses the raw content to json"""
-        return json.loads(self.get())
+    def get_yaml(self) -> dict | list:
+        """Gets and parses the raw content as yaml
+
+        Content may already be parsed according to its actual content type,
+        e.g. JSON content is returned as-is.
+        """
```
mhoff:
```suggestion
        """Gets and parses the raw content as yaml"""
```

logprep/abc/getter.py
```diff
@@ -79,26 +86,75 @@ def _get_used_env_vars(self, content, template):
...
+    @staticmethod
+    def _parse_json(content: str) -> dict | list:
+        """Parse content to json"""
+        return json.loads(content)
+
+    def get_json(self) -> dict | list:
+        """Gets and parses the raw content as json
+        Content may already be parsed according to its actual content type,
+        e.g. YAML content is returned as-is.
+        """
```
mhoff:
```suggestion
        """Gets and parses the raw content as json"""
```

mhoff (commented):
Minor pydoc changes missing

logprep/abc/getter.py
mhoff: This is not resolved

```suggestion
        """Gets and parses the raw content as json"""
```

mhoff (changes requested):

mhoff (approved):
LGTM

PR #961, opened 2026-04-29
feat: continue processing list and network comparison with failure tags 

kaya-david:
## Description

* list_comparison, network_comparison: continue processing with failure tags after HTTP list get or refresh errors
* add tests for rule error state switches
* add fail state on top Rule level and fail checks on processor top level for ng- and non-ng path

## Assignee

### Documentation

### Code Quality

* implemented new tests an run pytest 

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--961.org.readthedocs.build/).


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 ### Features
 
 ### Improvements
-
+* allow `list_comparison` to continue processing if comparison list initialization fails and log a warning instead
```
kaya-david: I am not sure if this is a breaking change, an improvement, or a feature. :D

mhoff (changes requested):
As discussed

logprep/util/rule_loader.py
```diff
@@ -40,20 +42,34 @@ def rules(self) -> list[Rule]:
...
+                    path = Path(self.source)
+                    if path.is_dir():
+                        rules.extend(DirectoryRuleLoader(self.source, self.processor_name).rules)
+                    if path.is_file():
+                        rules.extend(FileRuleLoader(self.source, self.processor_name).rules)
+                case list():
+                    for element in self.source:
+                        rules.extend(RuleLoader(element, self.processor_name).rules)
+        except Exception as ex:
+            self._handle_error(error=ex)
```
mhoff: Why do we need to silence errors during rule collection? The error that a list could not be retrieved happens during `setup` which should be after this. Wouldn't we just drop and forget about rules following this approach?

logprep/processor/list_comparison/processor.py
```diff
@@ -56,7 +60,10 @@ class Config(Processor.Config):
     def setup(self):
         super().setup()
         for rule in self.rules:
-            rule.init_list_comparison(self._config.list_search_base_path)
+            try:
+                rule.init_list_comparison(self._config.list_search_base_path)
+            except Exception as ex:
+                self._collect_rule_error(rule=rule, ex=ex)
```
mhoff: How would the rule recover from this? So when would the list retrieval be retried in order to make the rule healthy again?

mhoff (commented):
I did a first high-level pass

logprep/abc/processor.py
```diff
@@ -282,7 +282,9 @@ def load_rules(self, rules_targets: Sequence[str | dict]) -> None:
     def _field_exists(event: dict, dotted_field: str) -> bool:
         return has_dotted_field(event, dotted_field)
 
-    def _handle_warning_error(self, event, rule, error, failure_tags=None):
+    def _handle_warning_error(
+        self, event: dict, rule: "Rule", error: Exception, failure_tags: list | None = None
```
mhoff: Please add the generic type annotations for `event: dict` and `failure_tags: list`

logprep/processor/list_comparison/processor.py
```diff
@@ -73,6 +73,11 @@ def _apply_rules(self, event, rule):
             Currently applied list comparison rule.
 
         """
+
+        if rule.fail:
```
mhoff: I would rather like an adjective here. Like "failed" or "disabled". I see that this conflicts with the attribute also holding the exception. Maybe have an `is_failed` method implemented as `self._data_error is not None` to make that differentiation

logprep/processor/list_comparison/rule.py
```diff
@@ -97,6 +100,22 @@ def __init__(
...
+            self.__class__.__name__,
+            error,
+        )
+
+    def clear_failed(self) -> None:
+        self._fail = None
+
+    @property
+    def fail(self):
+        return self._fail
```
mhoff: This would actually be something I would consider putting in the base class and not the concrete rule type

logprep/processor/list_comparison/rule.py
```diff
@@ -119,15 +138,21 @@ def _init_list_comparison_from_http(self, list_search_base_path: str) -> None:
                 {**os.environ, **{"LOGPREP_LIST": list_path}}
             )
             http_getter = GetterFactory.from_string(list_search_base_path_resolved)
+
```
mhoff, line 122: Blank line on purpose? If so, fine by me

logprep/processor/list_comparison/rule.py
```diff
@@ -119,15 +138,21 @@ def _init_list_comparison_from_http(self, list_search_base_path: str) -> None:
...
-        content = http_getter.get_list()
-        file_elem_tuples = (elem for elem in content if not elem.startswith("#"))
-        self._compare_sets.update({list_path: set(file_elem_tuples)})
+        try:
+            content = http_getter.get_list()
+            file_elem_tuples = (elem for elem in content if not elem.startswith("#"))
+            self._compare_sets.update({list_path: set(file_elem_tuples)})
+            self.clear_failed()
+        except Exception as ex:
+            self.mark_failed(error=ex)
```
mhoff, line 138:
```suggestion
        try:
            content = http_getter.get_list()
            file_elem_tuples = (elem for elem in content if not elem.startswith("#"))
            self._compare_sets.update({list_path: set(file_elem_tuples)})
        except Exception as ex:
            self.mark_failed(error=ex)
        else:
            self.clear_failed()
```

logprep/processor/list_comparison/rule.py
```diff
@@ -119,15 +138,21 @@ def _init_list_comparison_from_http(self, list_search_base_path: str) -> None:
...
             http_getter.add_callback(self._update_compare_sets_via_http, http_getter, list_path)
+            self._update_compare_sets_via_http(http_getter, list_path)
 
     def _update_compare_sets_via_http(self, http_getter: HttpGetter, list_path: str) -> None:
-        content = http_getter.get_list()
-        file_elem_tuples = (elem for elem in content if not elem.startswith("#"))
-        self._compare_sets.update({list_path: set(file_elem_tuples)})
+        try:
+            content = http_getter.get_list()
+            file_elem_tuples = (elem for elem in content if not elem.startswith("#"))
```
mhoff: I think these aren't tuples. Consider renaming `file_elem_tuples`

logprep/processor/list_comparison/rule.py
```diff
@@ -150,3 +175,8 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
     @property
     def compare_sets(self) -> dict:  # pylint: disable=missing-docstring
         return self._compare_sets
+
+    @property
+    def failure_tags(self):
```
mhoff: Missing type annotation

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Upcoming Changes
 ### Breaking
 * chart: change default deployment strategy to RollingUpdate
+* handle list_comparison and network_comparison errors as warnings
```
mhoff: A bit too general

mhoff (commented):
Hello @kaya-david, many thanks. I have added some remarks. Please also consider the ng-path. I have not yet checked the tests

logprep/abc/processor.py
kaya-david: From my point of view, the annotation should be applied consistently everywhere. At this point, I do not know what concrete type event actually has underneath. My idea would be `event: dict[str, Any]`.

Either we keep the signature consistent with all other methods in this class, or I add event: dict[str, Any] only partially here. What do you think?

logprep/processor/list_comparison/rule.py
kaya-david: I avoid placing statements directly after blocks without a blank line.

The blank line is meant to make it clear that the statement does not belong to the if block. Unfortunately, in Python it can easily happen that statements are accidentally moved into the wrong branch due to indentation.

kaya-david:
> Hello [@kaya-david](https://github.com/kaya-david), many thanks. I have added some remarks. Please also consider the ng-path. I have not yet checked the tests

I see a problem here. It is not only that I made changes to the processors, but also that some previous changes in the non-ng path are missing in the ng path. For example, the non-ng path calls the helper function `join_dotted_fields(...)`, but this call is missing in the ng path.

If we synchronize the paths now, I think we might introduce new issues. Therefore, I would suggest handling this synchronization as a follow-up task.

logprep/abc/processor.py
mhoff: `dict[str, FieldValue]`

mhoff:
> > Hello [@kaya-david](https://github.com/kaya-david), many thanks. I have added some remarks. Please also consider the ng-path. I have not yet checked the tests
> 
> I see a problem here. It is not only that I made changes to the processors, but also that some previous changes in the non-ng path are missing in the ng path. For example, the non-ng path calls the helper function `join_dotted_fields(...)`, but this call is missing in the ng path.
> 
> If we synchronize the paths now, I think we might introduce new issues. Therefore, I would suggest handling this synchronization as a follow-up task.

`join_dotted_fields` is the only delta I see (except for typing) between ng and non-ng for the `list_comparison/processor.py`:

```
31c31
< from logprep.abc.processor import Processor
---
> from logprep.ng.abc.processor import Processor
33,37c33
< from logprep.util.helper import (
<     add_fields_to,
<     get_dotted_field_value,
<     join_dotted_fields,
< )
---
> from logprep.util.helper import add_fields_to, get_dotted_field_value
56c52
<     def setup(self):
---
>     def setup(self) -> None:
62,63c58,59
<         """
<         Apply matching rule to given log event.
---
>         """Apply matching rule to given log event.
> 
78c74
<             fields = {join_dotted_fields((rule.target_field, comparison_key)): comparison_result}
---
>             fields = {f"{rule.target_field}.{comparis31c31
< from logprep.abc.processor import Processor
---
> from logprep.ng.abc.processor import Processor
33,37c33
< from logprep.util.helper import (
<     add_fields_to,
<     get_dotted_field_value,
<     join_dotted_fields,
< )
---
> from logprep.util.helper import add_fields_to, get_dotted_field_value
56c52
<     def setup(self):
---
>     def setup(self) -> None:
62,63c58,59
<         """
<         Apply matching rule to given log event.
---
>         """Apply matching rule to given log event.
> 
78c74
<             fields = {join_dotted_fields((rule.target_field, comparison_key)): comparison_result}
---
>             fields = {f"{rule.target_field}.{comparison_key}": comparison_result}
81c77
<     def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list, str]:
---
>     def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list[str], str]:on_key}": comparison_result}
81c77
<     def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list, str]:
---
>     def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list[str], str]:
```

`join_dotted_fields` is doing exactly the same as what is happening in non-ng:

```
def join_dotted_fields(dotted_fields: Iterable[str]) -> str:
    """Combines a list of dotted fields to a single dotted field by joining them
    in order and without escaping already existing dots or other characters additionally.
    For instance, :code:`x.y` and :code:`z` would be mapped to :code:`x.y.z` and
    joining :code:`x\\.y` and :code:`z` would result in :code:`x\\.y.z`.

    Parameters
    ----------
    dotted_fields : Iterable[str]
        The list of fields to be joined.

    Returns
    -------
    str
        A single dotted field.
    """
    return ".".join(dotted_fields)
```

Which means ng and non-ng are still functionally equivalent.

So, you can easily replace 

```
fields = {f"{rule.target_field}.{comparison_key}": comparison_result}
```

with

```
fields = {join_dotted_fields((rule.target_field, comparison_key)): comparison_result}
```

I therefore see no good reason why we should diverge paths functionally with this PR

logprep/processor/base/rule.py
```diff
@@ -302,6 +310,21 @@ def __repr__(self) -> str:
...
+            "%s failed: %s",
+            self.__class__.__name__,
+            error,
+        )
+
+    def clear_failed(self) -> None:
+        self._data_error = None
+
+    def is_failed(self) -> tuple[bool, Exception | None]:
+        failed = True if self._data_error is not None else False
```
mhoff:
```suggestion
        failed = self._data_error is not None
```

logprep/processor/list_comparison/processor.py
```diff
@@ -73,6 +73,12 @@ def _apply_rules(self, event, rule):
             Currently applied list comparison rule.
 
         """
+
+        failed, data_error = rule.is_failed()
+        if failed:
+            self._handle_warning_error(event=event, rule=rule, error=data_error)
+            return
```
mhoff: The returned tuple does not appear useful to me. We produce a `bool` which signifies whether the `Exception` is set or not but the type cannot convey this constraint, meaning application code needs to again differentiate `Exception | None`.

If we (for most use cases) need the exception together with the info _if_ the rule failed, then we can drop the helper method and simply check the property. Otherwise we are not really adding benefit or readability

```suggestion
        if rule.data_error is not None:
            self._handle_warning_error(event=event, rule=rule, error=rule.data_error)
            return
```

logprep/ng/processor/list_comparison/processor.py
```diff
@@ -49,14 +57,19 @@ class Config(Processor.Config):
 
     rule_class = ListComparisonRule
 
+    @property
+    def config(self) -> Config:
+        """Provides the properly typed rule configuration object"""
```
mhoff:
```suggestion
        """Provides the properly typed configuration object"""
```

logprep/ng/processor/list_comparison/processor.py
```diff
@@ -49,14 +57,19 @@ class Config(Processor.Config):
...
         super().setup()
-        for rule in self.rules:
-            rule.init_list_comparison(self._config.list_search_base_path)
-
-    def _apply_rules(self, event, rule):
-        """Apply matching rule to given log event.
+        for rule in typing.cast(list[ListComparisonRule], self.rules):
+            rule.init_list_comparison(self.config.list_search_base_path)
 
+    def _apply_rules(self, event: dict[str, FieldValue], rule: "Rule"):
```
mhoff:
```suggestion
    def _apply_rules(self, event: dict[str, FieldValue], rule: Rule):
```

tests/unit/ng/processor/list_comparison/test_list_comparison.py
```diff
@@ -409,3 +410,170 @@ def test_match_list_field(self, testcase, system, result):
...
+            session = original_get_requests_session(self)
+            captured_sessions.append(session)
+            return session
+
+        processor = Factory.create({"custom_lister": config})
+        rule = processor.rule_class.create_from_dict(rule_dict)
+        processor._rule_tree.add_rule(rule)
+
+        data_error = rule.data_error
+        assert data_error is None
```
mhoff:
```suggestion
        assert rule.data_error is None
```

tests/unit/ng/processor/list_comparison/test_list_comparison.py
```diff
@@ -409,3 +410,170 @@ def test_match_list_field(self, testcase, system, result):
...
+        with mock.patch.object(
+            HttpGetter,
+            "_get_requests_session",
+            autospec=True,
+            side_effect=capture_session,
+        ):
+            processor.setup()
+
+        data_error = rule.data_error
+        assert isinstance(data_error, RefreshableGetterError)
```
mhoff:
```suggestion
        assert isinstance(rule.data_error, RefreshableGetterError)
```

tests/unit/ng/processor/list_comparison/test_list_comparison.py
```diff
@@ -409,3 +410,170 @@ def test_match_list_field(self, testcase, system, result):
...
+
+        processor = Factory.create({"custom_lister": config})
+        rule = processor.rule_class.create_from_dict(rule_dict)
+        processor._rule_tree.add_rule(rule)
+
+        processor.setup()
+        processor.process(log_event)
+
+        data_error = rule.data_error
+        assert isinstance(data_error, RefreshableGetterError)
```
mhoff:
```suggestion
        assert isinstance(rule.data_error, RefreshableGetterError)
```

tests/unit/ng/processor/list_comparison/test_list_comparison.py
```diff
@@ -409,3 +410,170 @@ def test_match_list_field(self, testcase, system, result):
...
+            "user_results": {"in_list": [list_name]},
+        }
+
+        HttpGetter._shared.clear()
+
+        processor = Factory.create({"custom_lister": config})
+        rule = processor.rule_class.create_from_dict(rule_dict)
+        processor._rule_tree.add_rule(rule)
+
+        processor.setup()
```
mhoff, line 574: I am missing another test case which does not recreate the processor but uses the RefreshableGetter to recover in-place.

I think it's simply a copy of this one with the following changes applied:

```suggestion
        refresh_getters()
```

tests/unit/ng/processor/list_comparison/test_list_comparison.py
```diff
@@ -409,3 +410,170 @@ def test_match_list_field(self, testcase, system, result):
...
+        processor.process(log_event)
+
+        data_error = rule.data_error
+        assert data_error is None
```
mhoff:
```suggestion
        assert rule.data_error is None
```

tests/unit/processor/list_comparison/test_list_comparison_rule.py
```diff
@@ -103,11 +102,14 @@ def test_compare_set_not_empty_for_valid_rule_def_after_init_list_comparison(
         assert isinstance(rule.compare_sets, dict)
         assert len(rule.compare_sets.keys()) > 0
 
-    @mock.patch(
-        "logprep.processor.list_comparison.rule.GetterFactory.from_string",
-        return_value=mock.MagicMock(),
+    @pytest.mark.parametrize(
+        ("url", "will_fail"),
```
mhoff, line 106: You can supply a tuple here? Nice! Didn't know that

tests/unit/processor/list_comparison/test_list_comparison_rule.py
```diff
@@ -117,6 +119,13 @@ def test_expects_refreshable_getter(self, _):
             },
         }
         rule = ListComparisonRule.create_from_dict(rule_definition)
-        url = "http://something"
-        with pytest.raises(TypeError, match=f"The target {url} must be a url"):
-            rule._init_list_comparison_from_http(url)
```
mhoff, line 122: I don't understand why this should failed before or what we actually test here. Does it make change that the call succeeds now? Can you explain the reasoning to me?

Also, we are changing http in https. On purpose? Why not include http and https?

mhoff (commented):
I added some more final remarks. The code suggestions typically always apply to both, ng and non-ng.

Edit: Please also fill out the whole PR description and update the title

tests/unit/processor/list_comparison/test_list_comparison_rule.py
kaya-david: discussed

mhoff (approved):
LGTM, many thanks!

PR #976, opened 2026-06-18
feat: support ranges in Lucene filters and field groups

kaya-david:
## Description

* add support for inclusive numeric Lucene range expressions using square brackets, such as age:[18 TO 65].
* support integer and floating-point ranges for both regular search fields and field group syntax.
* support negative range boundaries by normalizing Luqum's Prohibit nodes.
* reject exclusive ranges, non-numeric or non-finite boundaries, and reversed ranges with descriptive errors.
* add parametrized tests covering valid ranges, boundary values, invalid expressions, and both supported syntax variants.

## Assignee

### Documentation

### Code Quality

* pytest

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--976.org.readthedocs.build/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 ### Breaking
 
 ### Features
+* add support for inclusive numeric Lucene range filters using square brackets.
+* add support for integer and floating-point ranges in regular fields and field groups.
+* add validation for reversed, non-numeric, non-finite, and unsupported exclusive ranges.
 * add `deduplicator` processor that removes duplicates from lists
 
 ### Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@
 ### Breaking
 
 ### Features
-* add support for validated integer, floating-point, and lexicographic string Lucene range filters in regular fields and field groups.
-* add support for quoted string range boundaries, allowing values with Lucene special characters such as ISO-8601 timestamps with timezone offsets.
+* add Lucene range support for integer, floating-point, and lexicographic string values, including quoted ISO-8601 timestamps with timezone offsets.
 * add `deduplicator` processor that removes duplicates from lists
 
 ### Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,8 @@
 ### Breaking
 
 ### Features
-* add support for inclusive numeric Lucene range filters using square brackets.
-* add support for integer and floating-point ranges in regular fields and field groups.
-* add validation for reversed, non-numeric, non-finite, and unsupported exclusive ranges.
+
+* add support for validated integer, floating-point, and lexicographic string Lucene range filters in regular fields and field groups.
 * add `deduplicator` processor that removes duplicates from lists
 
 ### Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 ### Breaking
 
 ### Features
-
 * add support for validated integer, floating-point, and lexicographic string Lucene range filters in regular fields and field groups.
+* add support for quoted string range boundaries, allowing values with Lucene special characters such as ISO-8601 timestamps with timezone offsets.
 * add `deduplicator` processor that removes duplicates from lists
 
 ### Improvements
--- a/logprep/filter/lucene_filter.py
+++ b/logprep/filter/lucene_filter.py
@@ -61,9 +61,13 @@ This example is not complete, since rules are specific to processors and require
 Range-Filter
 ------------
 
-It is possible to use inclusive range expressions to match numeric values.
-The lower and upper boundaries are included and must be specified using square
-brackets. Both integer and floating-point boundaries are supported.
+It is possible to use range expressions to match integer, floating-point, and
+string values. Square brackets include a boundary, while curly brackets exclude
+a boundary. String ranges are compared lexicographically.
+
+The lower and upper boundaries of a range must have the same type. Mixed ranges
+such as integer-to-float, integer-to-string, or float-to-string are not
+supported.
 
 
 ..  code-block:: yaml
@@ -76,6 +80,39 @@ brackets. Both integer and floating-point boundaries are supported.
 The example matches log messages in which the value of :code:`age` is greater
 than or equal to :code:`18` and less than or equal to :code:`65`.
 
+Exclusive boundaries are also supported:
+
+
+..  code-block:: yaml
+    :linenos:
+    :caption: Example
+
+    filter: 'age:{18 TO 65}'
+
+
+The example matches log messages in which the value of :code:`age` is greater
+than :code:`18` and less than :code:`65`.
+
+Floating-point ranges are supported as well:
+
+
+..  code-block:: yaml
+    :linenos:
+    :caption: Example
+
+    filter: 'temperature:[18.5 TO 25.0]'
+
+
+String ranges are compared lexicographically:
+
+
+..  code-block:: yaml
+    :linenos:
+    :caption: Example
+
+    filter: 'status:[alpha TO stable]'
+
+
 Range expressions can also be used within field groups:
 
 
@@ -86,9 +123,8 @@ Range expressions can also be used within field groups:
     filter: 'temperature:([18.5 TO 25.0])'
 
 
-Only closed numeric ranges using square brackets are currently supported.
-Exclusive ranges using curly brackets, open boundaries using :code:`*`, and
-non-numeric boundaries are not supported.
+Open boundaries using :code:`*`, non-finite numeric boundaries, and mixed
+boundary types are not supported.
 
 RegEx-Filter
 ------------
--- a/logprep/filter/lucene_filter.py
+++ b/logprep/filter/lucene_filter.py
@@ -159,9 +159,11 @@ from logprep.filter.expression.filter_expression import Not as NotExpression
 from logprep.filter.expression.filter_expression import (
     Null,
     Or,
+    RangeBoundary,
     RegExFilterExpression,
     SigmaFilterExpression,
     StringFilterExpression,
+    StringRangeFilterExpression,
 )
 from logprep.util.helper import field_list_to_dotted_field, get_dotted_field_list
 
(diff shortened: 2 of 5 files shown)


logprep/filter/lucene_filter.py
```diff
@@ -58,6 +58,38 @@
...
+..  code-block:: yaml
+    :linenos:
+    :caption: Example
+
+    filter: 'temperature:([18.5 TO 25.0])'
+
+
+Only closed numeric ranges using square brackets are currently supported.
+Exclusive ranges using curly brackets, open boundaries using :code:`*`, and
+non-numeric boundaries are not supported.
```
Pablu23: Do we want to allow square brackets? I think that should be a relativly easy implementation

logprep/filter/lucene_filter.py
```diff
@@ -275,6 +313,14 @@ def _parse_tree(self, tree: luqum.tree) -> FilterExpression:
             # pylint: enable=no-value-for-parameter
         if isinstance(tree, Group):
             return self._parse_tree(tree.children[0])
+
+        if isinstance(tree, Range):
+            if self._last_search_field is None:
+                raise LuceneFilterError(f'The expression "{str(tree)}" is invalid!')
+
+            key = get_dotted_field_list(self._last_search_field)
+            return self._parse_range(key, tree)
```
Pablu23, line 390: Is this actually ever run? In my testing I think this couldnt be reached / wasnt, but im not sure

logprep/filter/lucene_filter.py
```diff
@@ -288,6 +334,97 @@ def _parse_tree(self, tree: luqum.tree) -> FilterExpression:
...
 
+    def _parse_range(self, key: Sequence[str], expr: Range) -> FilterExpression:
+        expression = str(expr)
+        lower_token, upper_token = expr.children
+
+        if not expression.startswith("[") or not expression.endswith("]"):
+            raise LuceneFilterError(
+                f'The expression "{expression}" is invalid. '
+                "Only inclusive ranges using square brackets are supported."
+            )
```
Pablu23: This seems, overengineered, does expression not expose this in a nicer way?

logprep/filter/lucene_filter.py
```diff
@@ -288,6 +334,97 @@ def _parse_tree(self, tree: luqum.tree) -> FilterExpression:
...
+                key,
+                lower_bound,
+                upper_bound,
+            )
+
+        try:
+            lower_bound_float = float(lower_value)
+            upper_bound_float = float(upper_value)
+        except ValueError as error:
+            raise LuceneFilterError(f'The expression "{expression}" is invalid!') from error
```
Pablu23: We do want string ranges, so timestamps of ISO 8601 Format, are possible

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 ### Breaking
 
 ### Features
+* add support for inclusive numeric Lucene range filters using square brackets.
+* add support for integer and floating-point ranges in regular fields and field groups.
+* add validation for reversed, non-numeric, non-finite, and unsupported exclusive ranges.
```
Pablu23: I would summarize this all as one point if possible

Pablu23 (commented):
Looks good so far, had a few questions and notes

logprep/filter/lucene_filter.py
kaya-david: It does. Setting a breakpoint in this code and running the tests shows it.

logprep/filter/lucene_filter.py
kaya-david: Implemented square brackets

logprep/filter/lucene_filter.py
```diff
@@ -288,6 +402,211 @@ def _parse_tree(self, tree: luqum.tree) -> FilterExpression:
...
+        upper_value: str,
+        include_lower_bound: bool,
+        include_upper_bound: bool,
+        expr: Range,
+    ) -> FloatRangeFilterExpression | None:
+        """Create a float range expression if both boundaries are finite floats."""
+        if LuceneTransformer._is_integer_range_boundary(
+            lower_value
+        ) or LuceneTransformer._is_integer_range_boundary(upper_value):
+            return None
```
Pablu23: What is this for? Through ordering alone this shoulndt be possible, and If I call this with an int explicitly I might want it to behave like a float. also this doesnt really make a difference for floats to ints?

logprep/filter/lucene_filter.py
```diff
@@ -288,6 +402,211 @@ def _parse_tree(self, tree: luqum.tree) -> FilterExpression:
...
+        key: Sequence[str],
+        lower_value: str,
+        upper_value: str,
+        include_lower_bound: bool,
+        include_upper_bound: bool,
+        expr: Range,
+    ) -> StringRangeFilterExpression:
+        """Create a lexicographic string range expression for non-numeric boundaries."""
+        if lower_value == "*" or upper_value == "*":
+            raise LuceneFilterError(f'The expression "{expr}" is invalid!')
```
Pablu23, line 496: What is this check for?

logprep/filter/lucene_filter.py
```diff
@@ -288,6 +402,211 @@ def _parse_tree(self, tree: luqum.tree) -> FilterExpression:
...
+
+        if LuceneTransformer._is_numeric_range_boundary(
+            lower_value
+        ) or LuceneTransformer._is_numeric_range_boundary(upper_value):
+            raise LuceneFilterError(f'The expression "{expr}" is invalid!')
```
Pablu23: Same as above, I would remove this as ordering will disallow it anyway and calling it explicitly should not raise an error, as it is possible

logprep/filter/lucene_filter.py
```diff
@@ -288,6 +402,211 @@ def _parse_tree(self, tree: luqum.tree) -> FilterExpression:
...
+        """Return whether a range boundary can be interpreted as a finite float."""
+        if LuceneTransformer._is_integer_range_boundary(value):
+            return False
+
+        try:
+            numeric_value = float(value)
+        except ValueError:
+            return False
+
+        return math.isfinite(numeric_value)
```
Pablu23: This is unnecessary then

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 ### Breaking
 
 ### Features
+* add support for validated integer, floating-point, and lexicographic string Lucene range filters in regular fields and field groups.
+* add support for quoted string range boundaries, allowing values with Lucene special characters such as ISO-8601 timestamps with timezone offsets.
```
Pablu23: I would put them both into one summary of what happened, "add lucene range support for integer, float and strings, supporting ISO-8601 timestamps"

logprep/filter/lucene_filter.py
```diff
@@ -288,6 +402,211 @@ def _parse_tree(self, tree: luqum.tree) -> FilterExpression:
...
+                upper_value,
+                include_lower_bound,
+                include_upper_bound,
+                expr,
+            )
+
+            if range_filter_expression is not None:
+                return range_filter_expression
+
+        raise LuceneFilterError(f'The expression "{expression}" is invalid!')
```
Pablu23:
```suggestion
    def _parse_range(self, key: Sequence[str], expr: Range) -> FilterExpression:
        lower_value = self._get_range_boundary_value(expr.low)
        upper_value = self._get_range_boundary_value(expr.high)

        for range_parser in (
            self._parse_integer_range,
            self._parse_float_range,
            self._parse_string_range,
        ):
            range_filter_expression = range_parser(
                key,
                lower_value,
                upper_value,
                expr.include_low,
                expr.include_high,
                expr,
            )

            if range_filter_expression is not None:
                return range_filter_expression

        raise LuceneFilterError(f'The expression "{expr}" is invalid!')
```

Pablu23 (commented):
Thanks for the work, looks good so far, there are a few points I would like to see a change or I dont agree with but thats it

logprep/filter/lucene_filter.py
kaya-david: This check explicitly rejects open-ended Lucene ranges such as `[* TO 10]` or `[10 TO *]`.

The current implementation only supports closed ranges with concrete lower and upper boundaries. Without this check, `*` could fall through to the string range parser and be treated as a normal lexicographic string boundary, which would make open ranges appear to be supported accidentally.

logprep/filter/lucene_filter.py
kaya-david: This check prevents numeric-looking boundaries from falling through to the string range parser.

Without it, expressions such as `[1 TO bar]` or `[0 TO 10.5]` could be interpreted as lexicographic string ranges after the integer and float parsers declined them. That would make mixed or ambiguous range boundaries appear to be supported accidentally.

The intention here is that unquoted numeric-looking boundaries are handled only by the numeric range parsers. String ranges should only use non-numeric boundaries, or quoted boundaries when the value contains Lucene special characters.

logprep/filter/lucene_filter.py
kaya-david: That make sense, removed it.

logprep/filter/lucene_filter.py
kaya-david: fixed

logprep/filter/lucene_filter.py
kaya-david: Fixed/changed

CHANGELOG.md
kaya-david: Done

logprep/filter/lucene_filter.py
Pablu23: I see, then im fine with this, do we have it in the docs that we dont allow open-ended ranges for now?

logprep/filter/lucene_filter.py
Pablu23: Okay than that seems good to me aswell, thanks for the clarification

logprep/filter/lucene_filter.py
```diff
@@ -288,6 +402,171 @@ def _parse_tree(self, tree: luqum.tree) -> FilterExpression:
...
+        expr: Range,
+    ) -> StringRangeFilterExpression | None:
+        """Create a lexicographic string range expression for non-numeric boundaries."""
+        if lower_value == "*" or upper_value == "*":
+            raise LuceneFilterError(f'The expression "{expr}" is invalid!')
+
+        if LuceneTransformer._is_finite_numeric_range_boundary(
+            lower_value
+        ) or LuceneTransformer._is_finite_numeric_range_boundary(upper_value):
+            return None
```
Pablu23, line 501: Maybe raise here instead of returning None, and then waiting for it to raise in the upper method, to keep consistent with the other methods

kaya-david: I would keep returning `None` here.

The range parsers are tried in order: integer, float, string. Returning `None` means “this parser is not responsible for this boundary type”, while raising is reserved for cases that are invalid for the parser that did accept the type.

Numeric-looking boundaries are not valid string range boundaries, but they are also not string-parser-specific errors. Returning `None` lets `_parse_range` produce the final generic invalid-range error if no parser can handle the expression. This keeps the parser chain consistent with `_parse_integer_range` and `_parse_float_range`, which also return `None` when the boundary type does not match.

Pablu23: Yeah true that is fair

Pablu23 (approved):
LGTM thank you for the work

PR #977, opened 2026-06-23
feat: make compare lists dynamic

Pablu23:
## Description

* allow `list_comparison` and `network_comparison` to use dynamic values from event to resolve list uris

## Assignee

### Documentation

### Code Quality

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--977.org.readthedocs.build/).


logprep/processor/list_comparison/rule.py
```diff
@@ -192,8 +205,78 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
...
+                identifier: get_dotted_field_value(event, identifier)
+                for identifier in resolved_tmpl.get_identifiers()
+            }
+            for _, val in key_val.items():
+                if not isinstance(val, (str, int)):
+                    raise ValueError("value for list comparison is not a scalar value")
+                pass
+
+            # Dont use safe substitute here, as we want it to fail if an event key is missing,
+            # might be redundant because the previous func checks for Scalar str, int and None is not that
```
mhoff: I'd prefer a try...except which makes this behavior explicit. Even raising the error again (possibly after generating a log) would be fine for me. That would make it clear that we expect Exceptions on this line

logprep/processor/list_comparison/rule.py
```diff
@@ -192,8 +205,78 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
...
+
+            compare_set = self._load_http_compare_set(dynamic_resolved)
+            if compare_set is not None:
+                compare_sets_result.update({dynamic_resolved: compare_set})
+
+        return compare_sets_result
+
     @property
-    def compare_sets(self) -> dict:  # pylint: disable=missing-docstring
+    def compare_sets(self) -> dict[str, set]:  # pylint: disable=missing-docstring
```
mhoff: Add a short docstring instead of pylint disable

logprep/processor/network_comparison/rule.py
```diff
@@ -85,7 +85,7 @@ class Config(FieldManagerRule.Config):
         You can also pass a template with keys from environment,
         e.g.,  :code:`${<your environment variable>}`. The special key :code:`${LOGPREP_LIST}`
         will be filled by this processor. """
-        mapping: dict = field(default="", init=False, repr=False, eq=False)
+        mapping: dict = field(default={}, init=False, repr=False, eq=False)
```
mhoff: factory=set, defaults should never be mutable object instances

logprep/util/getter.py
```diff
@@ -73,10 +75,8 @@ def from_string(cls, getter_string: str) -> "Getter":
             protocol = "file"
         if protocol == "file":
             return FileGetter(protocol=protocol, target=target)  # type: ignore
-        if protocol == "http":
-            return HttpGetter(protocol=protocol, target=target)  # type: ignore
-        if protocol == "https":
-            return HttpGetter(protocol=protocol, target=target)  # type: ignore
+        if protocol == "http" or protocol == "https":
+            return HttpGetter(protocol=protocol, target=f"{protocol}://{target}")  # type: ignore
```
mhoff:
```suggestion
            return HttpGetter(protocol=protocol, target=getter_string)  # type: ignore
```

Isn't this equivalent?

mhoff: Why do we need the type ignore here? `HttpGetter` is a subclass of `Getter`, so that should be fine, isn't it?

examples/exampledata/config/dynamic-compare-lists-getter-config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "http://localhost:8000/*": {
+    "refresh_interval": 10
+  }
+}
```
mhoff: I like that you did testing on the whole application level. However, this (including the pipeline below) feels more like an acceptance test to me

logprep/util/getter.py
```diff
@@ -226,44 +234,80 @@ def _default_return_value(self) -> bytes | None:
             self.shared.default_return_value = self._get_default_return_value()
         return self.shared.default_return_value
 
-    def add_callback(self, fnc, *args, **kwargs):
+    def add_callback(self, owner: str, fnc, *args, **kwargs):
```
mhoff: I'd prefer "tag" over "owner" as nobody enforces this parameter to actually encapsulate the owner of the callback. I think the wording "tag" is more precise. Also applies to cleanup callbacks

logprep/processor/list_comparison/rule.py
```diff
@@ -134,42 +142,47 @@ def __init__(
...
-            http_getter = GetterFactory.from_string(list_search_base_path_resolved)
-            if not isinstance(http_getter, HttpGetter):
-                raise TypeError(f"The target {list_search_base_path_resolved} must be a url")
-            self._update_compare_sets_via_http(http_getter, list_path)
-            http_getter.add_callback(self._update_compare_sets_via_http, http_getter, list_path)
+            self._init_static_http_list_comparison()
 
-    def _update_compare_sets_via_http(self, http_getter: HttpGetter, list_path: str) -> None:
+    def _update_compare_sets_via_http(
+        self, http_getter: HttpGetter, fully_resoled_uri: str
```
mhoff: Typo

logprep/processor/list_comparison/rule.py
```diff
@@ -192,8 +205,78 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
             filename = os.path.basename(list_path)
             self._compare_sets.update({filename: set(file_elements)})
 
+    def _init_static_http_list_comparison(self) -> None:
+        for list_path in self._config.list_file_paths:
+            resolved = Template(self._config.list_search_base_path).safe_substitute(
+                {**os.environ, **{"LOGPREP_LIST": list_path}}
+            )
+            if Template(resolved).get_identifiers():
+                continue
```
mhoff: Isn't this an error scenario?

logprep/processor/list_comparison/rule.py
```diff
@@ -192,8 +205,78 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
...
+            self._load_http_compare_set(resolved)
+
+    def _load_http_compare_set(self, resolved_uri: str) -> set[dict] | None:
+        http_getter = GetterFactory.from_string(resolved_uri)
+        if not isinstance(http_getter, HttpGetter):
+            raise TypeError(f"The target {resolved_uri} must be a url")
+
+        http_getter.signal_called()
```
mhoff: Maybe .keep_alive()?

logprep/processor/list_comparison/rule.py
```diff
@@ -192,8 +205,78 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
...
+
+        compare_set = self._update_compare_sets_via_http(http_getter, resolved_uri)
+        owner = self._callback_owner
+
+        refresh_key = (owner, resolved_uri, "list_comparison_refresh")
+        cleanup_key = (owner, resolved_uri, "list_comparison_cleanup")
```
mhoff: I don't think that we need this. `http_getter` knows about its uri and `add_callback` & `add_cleanup_callback` are different. Hence they key is reduced to owner. So as long as we are only using the key to deduplicate _inside_ a given getter (or rather _inside_ a DataSharedPerTarget which is proxied by the getter), the owner tag should be sufficient

logprep/processor/list_comparison/rule.py
```diff
@@ -192,8 +205,78 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
...
+
+        http_getter.add_callback_once(
+            refresh_key, owner, self._update_compare_sets_via_http, http_getter, resolved_uri
+        )
```
mhoff: _If_ the above complex key is not necessary, we can consolidate this function with `add_callback` and maybe have a parameter to turn on/off deduplication

logprep/processor/list_comparison/rule.py
```diff
@@ -192,8 +205,78 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
...
+
+        def cleanup():
+            self._compare_sets.pop(resolved_uri, None)
+            logger.debug("Deleted compare set for %s after cleanup", resolved_uri)
```
mhoff: Make this an instance method instead. That's more consistent with the other callback and `resolved_uri` can be passed as a parameter

logprep/processor/list_comparison/rule.py
```diff
@@ -192,8 +205,78 @@ def _init_list_comparison_from_local_file(self, list_search_base_path: str) -> N
...
+
+        http_getter.add_cleanup_callback_once(cleanup_key, owner, cleanup)
+        return compare_set
+
+    def get_dynamic_set(self, event: dict) -> dict[str, set]:
+        compare_sets_result: dict[str, set] = {}
+        if not self._config.list_search_base_path.startswith("http"):
```
mhoff: Hereby we are assuming that `FileGetters` are not `RefreshableGetters` which is true at the moment. But would there be a problem if we skip this check and rely on the getter implementation instead?

mhoff (commented):
Many thanks for the work. I did a quick (not exhaustive) review pass

logprep/util/getter.py
Pablu23: No it isnt as in the lines above, target get filled with environment variables, where getter_string still has placeholders, in the same vein though, im unsure if we ever need to supply the os.environ to the substitution below, because they shouldnt be there anymore

logprep/processor/list_comparison/rule.py
Pablu23: it is, aslong as we are not in a static context

logprep/processor/list_comparison/rule.py
Pablu23: You are right we dont, I somehow forgot that self already supplies the resolved_uri, without that we would still need it

logprep/processor/list_comparison/rule.py
Pablu23: We are, and I would keep it that way for now, we could practically change it to check for RefreshableGetter or not, but that brings other sideeffects with it because of attrs_post_init that we dont want here. I would mark this down as techincal debt but wouldnt bother with it for now

mhoff: Fine by me, thanks

logprep/processor/list_comparison/rule.py
Pablu23: I handle the missing fields case explicitly now

logprep/util/getter.py
```diff
@@ -226,44 +235,80 @@ def _default_return_value(self) -> bytes | None:
             self.shared.default_return_value = self._get_default_return_value()
         return self.shared.default_return_value
 
-    def add_callback(self, fnc, *args, **kwargs):
+    def add_callback(self, tag: str, fnc, deduplicate: bool = False, *args, **kwargs):
```
mhoff: This is problematic as the default and explicit value for `deduplicate` will be confused with *args and **kwargs. For instance would `fnc` not be allowed to receive a `deduplicate=whatever` parameter. We should stop using variadic arguments to capture `fnc` parameters. The same is true for `tag` and `fnc`.

Proposal:
`def add_callback(self, tag: str, fnc, *, deduplicate: bool = False, fnc_args: Iterable[Any], fnc_kwargs: dict[str, Any]):`

Also, do we want to enforce supplying a tag? What about existing code? Maybe adding a default here would not be bad. Except if we really want to nudge the developer to clean up

logprep/util/getter.py
```diff
@@ -226,44 +235,80 @@ def _default_return_value(self) -> bytes | None:
...
+        if ENV_NAME_LOGPREP_GETTER_CONFIG not in os.environ:
+            return {}
+
+        getter_file_path = os.environ.get(ENV_NAME_LOGPREP_GETTER_CONFIG)
+        if not getter_file_path or getter_file_path == self.target and self.protocol == "file":
+            return {}
+
+        getters_config = FileGetter(protocol="file", target=getter_file_path).get_dict()
+
+        legacy_target = self.target.removeprefix("https://").removeprefix("http://")
```
mhoff: Is there a difference between `target` and `legacy_target`? I thought `target` would already be stripped of the protocol

Pablu23: Yes that was the case but not anymore as we need that now, so legacy_target does exactly that to preserve the old way of writing them, to not incur a breaking change

logprep/util/getter.py
Pablu23: You are correct, I think I already changed that to a non defaulting argument, but your style looks alot better. And there are currently only two sites where add_callback is called from, I have edited both and because were good programmers we do want to cleanup behind ourselves d:

logprep/ng/util/configuration.py
```diff
@@ -653,7 +656,7 @@ class Configuration:
     _metrics: "Configuration.Metrics" = field(init=False, repr=False, eq=False)
 
     _getter: Getter = field(
-        validator=validators.instance_of(Getter),
+        validator=validators.instance_of((FileGetter | RefreshableGetter)),
```
mhoff, line 659: I am a bit surprised that this works. Per documentation a tuple type is expected, see the example here: [API Reference](https://www.attrs.org/en/stable/api.html#attrs.validators.not_)

CHANGELOG.md
```diff
@@ -2,10 +2,12 @@
 ### Breaking
 
 ### Features
+* allow `list_comparison` to use dynamic values from event to resolve list uris
```
mhoff: Also `network_comparison`

tests/conftest.py
```diff
@@ -1,9 +1,14 @@
 """Global configuration and fixtures for all pytest-based tests"""
 
+import os
 from multiprocessing import set_start_method
 
 import pytest
 
+os.environ.pop("LOGPREP_GETTER_CONFIG", None)
+os.environ.pop("PROMETHEUS_MULTIPROC_DIR", None)
+os.environ.pop("prometheus_multiproc_dir", None)
```
mhoff: This seems odd to me.

1) Why do we need to unset prometheus stuff for this PR?
2) I am not sure if `os.environ.pop` is the proper pytest way to modify environments, but if we stick with it, let's rather put it in a fixture like the start method below

Pablu23: we dont, but when you have PROMETHEUS_MULTIPROC_DIR set, some tests dont complete correctly same with LOGPREP_GETTER_CONFIG and I dislike that my tests are environment variable dependend so I popped them there

tests/unit/util/test_getter.py
```diff
@@ -1612,3 +1730,246 @@ def test_get_parsed_content_with_content_field_returns_expected_content(
...
+        ]
+
+    def test_callbacks_are_not_shared_between_targets(self):
+        http_getter_1 = HttpGetter(protocol="http", target="http://example.test/one")
+        http_getter_2 = HttpGetter(protocol="http", target="http://example.test/two")
+
+        callback = mock.MagicMock()
+        cleanup_callback = mock.MagicMock()
+
+        http_getter_1.add_callback("owner", callback)
```
mhoff: Rather use "tag123" or something

tests/unit/util/test_getter.py
```diff
@@ -40,7 +40,8 @@
 
 
 @pytest.fixture(autouse=True)
-def fixture_clear_getter_cache():
+def fixture_clear_getter_cache(monkeypatch):
+    monkeypatch.delenv(ENV_NAME_LOGPREP_GETTER_CONFIG, raising=False)
     getter = HttpGetter(protocol="http", target="anything")
     getter._shared.clear()
```
mhoff: I wonder if this fixture shouldn't be somewhere in conftest for all tests. We never want the cache to influence following tests. 

Then we could remove it here and also remove at least some of the extra `delenv` statements

logprep/util/getter.py
mhoff: Nevermind, makes sense

logprep/util/getter.py
```diff
@@ -219,51 +233,183 @@ def _refresh_interval(self, value: int) -> None:
...
+        deduplication_key: tuple | None,
+        fnc_args: Iterable[Any] | None,
+        fnc_kwargs: dict[str, Any] | None,
+    ) -> None:
+        callbacks = getattr(shared, callback_list_name)
+        callback = cls._build_callback(tag, fnc, fnc_args, fnc_kwargs)
+
+        if deduplication_key is not None:
+            if any(existing.get("key") == deduplication_key for existing in callbacks):
+                return
```
mhoff, line 297: I wonder if updating the existing callback makes more sense in this case. That's a bit implicit, but currently it'd be silently ignored that I changed some `fnc_args`

logprep/util/getter.py
```diff
@@ -219,51 +233,183 @@ def _refresh_interval(self, value: int) -> None:
...
+
+        getter_file_path = os.environ.get(ENV_NAME_LOGPREP_GETTER_CONFIG)
+        if not getter_file_path or getter_file_path == self.target and self.protocol == "file":
+            return {}
+
+        getters_config = FileGetter(protocol="file", target=getter_file_path).get_dict()
+
+        legacy_target = self.target.removeprefix("https://").removeprefix("http://")
+
+        for candidate in (self.uri, self.target, legacy_target):
```
mhoff: So, `uri` always has the protocol. `target` sometimes has the protocol (judging from `uri` property implementation) and `legacy_target` never has the http-protocol? I feel like it should only be two cases: (1) with protocol, (2) without protocol

logprep/util/getter.py
```diff
@@ -219,51 +233,183 @@ def _refresh_interval(self, value: int) -> None:
...
+                return getters_config[candidate]
+
+        for configured_target, config in getters_config.items():
+            if self._target_matches(configured_target):
+                return config
+
+        return {}
+
+    def _target_matches(self, configured_target: str) -> bool:
+        legacy_target = self.target.removeprefix("https://").removeprefix("http://")
```
mhoff: Maybe rather have a (cached) property for this one

logprep/util/getter.py
```diff
@@ -195,6 +206,9 @@ def hash(self, value: str) -> None:
     @property
     def uri(self) -> str:
         """Returns the URI of the target"""
+        # Protocol already in target
+        if re.match(r"^[a-zA-Z][a-zA-Z0-9+.-]*://", self.target):
+            return self.target
```
mhoff, line 220: Not sure if this makes sense, but see discussion below. If we stick with this, a cached property would make sense here I think

logprep/util/getter.py
```diff
@@ -330,10 +476,68 @@ def _get_raw(self) -> tuple[bytes, ContentType | None]:
...
+        target_shared = cls._shared.get(target)
+        if target_shared:
+            if target_shared.last_called:
+                elapsed = time.monotonic() - target_shared.last_called
+                timeout_interval = target_shared.timeout_interval
+                if timeout_interval is None:
+                    timeout_interval = 60
+
+                if elapsed > timeout_interval:
+                    return True
```
mhoff: This functionality is kind of spread out and partially redundant

- logic in `RefreshableGetter.timed_out_for_target` and in  `refreshableGetter.timed_out`
- default timeout interval in `RefreshableGetter.timed_out_for_target` and in `refreshableGetter._get_timeout_interval`

I would propose to move functionality into `DataSharedPerTarget` (maybe needs a better name then)

logprep/util/getter.py
```diff
@@ -330,10 +476,68 @@ def _get_raw(self) -> tuple[bytes, ContentType | None]:
...
+
+    @classmethod
+    def timed_out_for_target(cls, target: str):
+        target_shared = cls._shared.get(target)
+        if target_shared:
+            if target_shared.last_called:
+                elapsed = time.monotonic() - target_shared.last_called
+                timeout_interval = target_shared.timeout_interval
+                if timeout_interval is None:
+                    timeout_interval = 60
```
mhoff: I am a bit hesitant about the default interval in general. What happens if `timeout_inverval` is lower than `refresh_interval`? Wouldn't that effectively disable the refreshable getter (if not keep_alive is called in the meantime)? I'd argue that we should refrain from using a default interval for cleanups and delegate this to the configuring user to avoid weird and unintended side effects

logprep/util/getter.py
Pablu23: Yeah would be fine by me, but im also unsure, that has an very implicit part to it

logprep/util/getter.py
Pablu23: Correct

logprep/util/getter.py
Pablu23: How would we make this into a property if we need to accept input?

mhoff: I mean `legacy_target`, same as `target`

logprep/processor/network_comparison/rule.py
```diff
@@ -135,10 +135,15 @@ def init_list_comparison(self, owner: str, list_search_base_path: str | None = N
...
-        compare_set = super()._update_compare_sets_via_http(http_getter, list_path)
+    def _update_compare_sets_via_http(
+        self, http_getter: HttpGetter, list_path: str, *, mark_rule_failed: bool = True
+    ) -> set | None:
+        compare_set = super()._update_compare_sets_via_http(
+            http_getter, list_path, mark_rule_failed=mark_rule_failed
+        )
         self._convert_compare_sets_to_networks()
         if compare_set is None:
+            # TODO: Mark rule failed if set?
```
mhoff: I think not. `super._update_compare_sets_via_http` is responsible for that. Except if an error occured in `_convert_compare_sets_to_networks`, that might be a reason to fail the rule

logprep/ng/util/configuration.py
Pablu23: Honestly I think this is the way to do it, I dont know where I saw it, but Im pretty sure I saw it in the documentation, and using a tuple like the link you send shows, does NOT work (atleast my lsp complains)

logprep/util/getter.py
Pablu23: I changed this but I disagree with the no default, I agree that we should tell our users to set a timeout_interval themselves which fits, but its better to have one defaulted than have none because someone forgot and have a memory leak. The timeout_interval being lower than the refresh_interval is weird but fine, it gets cleaned up the next time around and wont be refreshed then anyways. Say we have:
refresh_interval: 10s
timeout_interval: 5s

0s: refreshes
5s: nothing happens but is timed out theoretically
10s: refreshes, sees its timed out, doesnt refresh that target and goes on

tests/unit/util/test_getter.py
Pablu23: That I dont know, let me know what you think would be the right place / implementation, I could probably put that in there then, im unsure of how exactly our / python testing in general works / loads

logprep/processor/network_comparison/rule.py
Pablu23: So I can probably just remove the todo and close this conversation I guess?

logprep/ng/util/configuration.py
mhoff: I just saw this:
```
            ),
            value_validator=attrs.validators.instance_of((str, int)),
        )
```
But probably outdated. If it works, it works and I agree, I like your version better

logprep/processor/network_comparison/rule.py
mhoff: Yes!

logprep/util/getter.py
mhoff: Imagine a component using refreshable getter like the list_comparison before we gave it keep_alive and cleanup. In the beginning, some static lists are initialized and getter callbacks registered. Imagine these callbacks are meant to update entries every 5 minutes.

Now, after 60s the timeout_interval kicks in, removing the DataSharedPerTarget and therefore the scheduled callback entries. The component entries would be frozen in time as the update callback will never be called and the component by itself will also not call the getter again. It assumes that the auto update mechanism is in place.

`generic_adder` and `generic_resolver` might fall exactly in this category.

If my understanding is correct and assuming there are components out there without cleanup and dynamic (re-)registering logic, we can not find a default value for timeout_interval which is big enough. It might always be the case that hours, days, weeks are taking until a certain entry is re-used.

Does that make sense?

logprep/util/getter.py
mhoff: Let's maybe offload this discussion. We can keep this for now

logprep/util/getter.py
Pablu23: Yes that makes sense, and maybe this is an unclean way to do it, but right now we wont get to that point, because the timeout only happens if the keep alive gets called atleast once on it. Components like generic_adder and generic_resolver where content stays static should not call keep alive on the Refreshable getter. Not interacting with the keep alive / timeout mechanism "deactivates" it

mhoff: Let's implement this advanced feature in a separate task.

mhoff (approved):
LGTM

kaya-david (approved):

Issue #984, opened 2026-07-02
Timestamper: UNIX timestamps with fractional seconds cannot be parsed


SDO-CN:
The `timestamper` processor (and the underlying `TimeParser`) cannot parse UNIX timestamps that contain fractional seconds (e.g. `1700000000.123`). Whole-second UNIX timestamps work, but any value with a decimal point fails to parse, even though fractional UNIX timestamps are a common and valid representation.

**Expected behavior**

`TimeParser.parse_datetime("1700000000.123", "UNIX", ...)` (and the `timestamper` processor with `source_format: [UNIX]`) should parse the fractional UNIX timestamp and normalize it to an ISO8601 datetime, preserving the sub-second component.

**Current behavior**

Parsing raises a `TimeParserException: invalid literal for int() with base 10: '1700000000.123'`. In the `timestamper` processor this surfaces as a `ProcessingWarning` ("Could not parse timestamp"), the event is tagged `_timestamper_failure`, and the target field is not written.

The root cause is in `TimeParser._normalize_unix_timestamp` (`logprep/util/time.py`), which normalizes the timestamp string via `int(timestamp)`:

```python
return (
    int(timestamp)
    if len(timestamp) <= 10
    else int(timestamp) / 10 ** (len(timestamp) - 10)
)
```

`int("1700000000.123")` raises `ValueError`, which is re-raised as `TimeParserException`. So any UNIX timestamp string with a decimal point can never be parsed.

**Steps to reproduce**

Minimal reproduction using the public `TimeParser` API:

```python
from datetime import timezone
from logprep.util.time import TimeParser

# Whole-second UNIX timestamp -> works
print(TimeParser.parse_datetime("1700000000", "UNIX", timezone.utc).isoformat())
# -> 2023-11-14T22:13:20+00:00

# Fractional-second UNIX timestamp -> raises
print(TimeParser.parse_datetime("1700000000.123", "UNIX", timezone.utc).isoformat())
# -> TimeParserException: invalid literal for int() with base 10: '1700000000.123'
```

The same failure occurs through the `timestamper` processor when an event field holds a fractional UNIX timestamp (as `float` or `str`) and the rule uses `source_format: [UNIX]`: the event ends up tagged `_timestamper_failure` instead of being normalized.

**Environment**

Logprep version: 19.3.0
Python version: 3.12.10

**Possible solution**

[...]

PR #985, opened 2026-07-03
fix: parse fractional UNIX timestamps

kaya-david:
## Description

* Parse UNIX timestamps with fractional seconds as seconds with sub-second precision.
Keep existing length-based normalization for integer-only UNIX timestamps.
Update the _normalize_unix_timestamp docstring to document fractional timestamp handling.
Add regression tests for fractional UNIX timestamps in the timestamper processor.
* Relates to #984

## Assignee

### Documentation

### Code Quality
- [ ] Patch test coverage > 95% and does not decrease
- [ ] New code uses correct & specific type hints

* pytest: Added test cases for UNIX timestamps with fractional seconds.

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--985.org.readthedocs.build/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## Upcoming Changes
 ### Breaking
+* add support for fractional UNIX timestamps in the `timestamper` processor while preserving integer timestamp normalization
 
 ### Features
-* add support for fractional UNIX timestamps in the `timestamper` processor while preserving integer timestamp normalization
 
 ### Improvements
 
@@ -60,34 +60,6 @@
 * improve robustness of compose tests
 
 ### Bugfix
-* fix `dissector` not dissecting multiline strings
-* fix `grokker` dropping matches on duplicate named capture groups
-
-## 19.4.1
-
-### Breaking
-
-### Features
-* add support for `dict` in `generic_adder`
-
-### Improvements
-* fix Pyparsing deprecation warnings
-
-### Bugfix
-* fix `add_fields_to` injecting identical objects instead of copies
-* fix `generic_adder` accumulating state by inserting identical objects in events
-
-## 19.4.0
-### Breaking
-
-### Features
-* allow `list_comparison` and `network_comparison` to use dynamic values from event to resolve list uris
-
-### Improvements
-
-### Bugfix
-* make `list_search_base_path` actually optional and correctly overrideable by rule config
-* add missing documentation for the `network_comparison` processor
 * fix `dissector` not dissecting multiline strings
 * fix `grokker` dropping matches on duplicate named capture groups
 * fix `list_comparison` & `network_comparison` to actually work with dotted field notation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 ### Breaking
 
 ### Features
-* add support for fractional UNIX timestamps in the timestamper processor while preserving integer timestamp normalization.
+* add support for fractional UNIX timestamps in the `timestamper` processor while preserving integer timestamp normalization
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,34 @@
 ### Bugfix
 * fix `dissector` not dissecting multiline strings
 * fix `grokker` dropping matches on duplicate named capture groups
+
+## 19.4.1
+
+### Breaking
+
+### Features
+* add support for `dict` in `generic_adder`
+
+### Improvements
+* fix Pyparsing deprecation warnings
+
+### Bugfix
+* fix `add_fields_to` injecting identical objects instead of copies
+* fix `generic_adder` accumulating state by inserting identical objects in events
+
+## 19.4.0
+### Breaking
+
+### Features
+* allow `list_comparison` and `network_comparison` to use dynamic values from event to resolve list uris
+
+### Improvements
+
+### Bugfix
+* make `list_search_base_path` actually optional and correctly overrideable by rule config
+* add missing documentation for the `network_comparison` processor
+* fix `dissector` not dissecting multiline strings
+* fix `grokker` dropping matches on duplicate named capture groups
 * fix `list_comparison` & `network_comparison` to actually work with dotted field notation
 * fix `list_comparison` & `network_comparison` rule failure handling for multiple lists
 * fix `list_comparison` & `network_comparison` to raise on http-urls without `LOGPREP_LIST`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 ## Upcoming Changes
 ### Breaking
-* add support for fractional UNIX timestamps in the `timestamper` processor while preserving integer timestamp normalization
+* Restrict UNIX timestamp normalization to supported precisions: seconds, milliseconds, microseconds, and nanoseconds. Integer and fractional UNIX timestamps with unsupported digit lengths are now rejected.
 
 ### Features
+* Add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ### Breaking
 
 ### Features
+* add support for fractional UNIX timestamps in the timestamper processor while preserving integer timestamp normalization.
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 ## Upcoming Changes
 ### Breaking
-* Restrict UNIX timestamp normalization to supported precisions: seconds, milliseconds, microseconds, and nanoseconds. Integer and fractional UNIX timestamps with unsupported digit lengths are now rejected.
+* restrict UNIX timestamp normalization to seconds, milliseconds, microseconds, and nanoseconds.
 
 ### Features
-* Add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
+* add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
 
 ### Improvements
 
--- a/logprep/util/time.py
+++ b/logprep/util/time.py
@@ -11,7 +11,7 @@ class TimeParserException(LogprepException):
 
 
 class UnixTimestampLength(IntEnum):
-    """Digit lengths of common Unix timestamp representations."""
+    """Digit lengths of common UNIX timestamp representations."""
 
     SECONDS = 10
     MILLISECONDS = 13
--- a/logprep/util/time.py
+++ b/logprep/util/time.py
@@ -1,6 +1,6 @@
 """Logprep time helpers module"""
 
-from datetime import datetime, tzinfo, UTC
+from datetime import UTC, datetime, tzinfo
 
 from logprep.abc.exceptions import LogprepException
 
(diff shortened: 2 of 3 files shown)


logprep/util/time.py
```diff
@@ -144,13 +147,19 @@ def _normalize_unix_timestamp(timestamp: str) -> int | float:
...
-            Raised if input timestamp string could not be parsed as integer
+            Raised if input timestamp string could not be parsed as a number
         """
         try:
+            timestamp = timestamp.strip()
+
+            if "." in timestamp:
+                return float(timestamp)
+
+            unsigned_timestamp = timestamp.removeprefix("-").removeprefix("+")
```
mhoff: Why do we need it and do we have a testcase for that?

logprep/util/time.py
```diff
@@ -144,13 +147,19 @@ def _normalize_unix_timestamp(timestamp: str) -> int | float:
...
         ------
         TimeParserException
-            Raised if input timestamp string could not be parsed as integer
+            Raised if input timestamp string could not be parsed as a number
         """
         try:
+            timestamp = timestamp.strip()
+
+            if "." in timestamp:
+                return float(timestamp)
```
mhoff: I wonder if we need the 10-digit-before-the-decimal logic here as well but honestly, I don't fully understand why we even need that. Feel free to resolve this comment as a plain remark

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ### Improvements
 
 ### Bugfix
+* add support for fractional UNIX timestamps in the timestamper processor while preserving integer timestamp normalization.
```
```suggestion
* add support for fractional UNIX timestamps in the timestamper processor while preserving integer timestamp normalization
```

mhoff: We typically dont use dots here

mhoff (commented):
Thanks, I left some minor remarks

logprep/util/time.py
kaya-david: Good point. This handles signed UNIX timestamps correctly when applying the digit-count heuristic. Without removing the sign, values like `-1000000000` would be counted as 11 characters and therefore be scaled incorrectly, even though the timestamp has only 10 digits.

I will remove it from this PR to keep the change focused on fractional UNIX timestamps.

logprep/util/time.py
kaya-david: Good point. I changed the logic to apply the digit-count heuristic to fractional UNIX timestamps as well.

The length check now uses the integer part before the decimal point. So `1700000000.123` is treated as seconds with fractional seconds, while longer values like `1700000000123.456` are scaled according to the existing normalization logic.

logprep/util/time.py
```diff
@@ -144,17 +148,52 @@ def _normalize_unix_timestamp(timestamp: str) -> int | float:
...
+            else parsed_timestamp / 10 ** (len(integer_part) - 10)
+        )
+
+    @staticmethod
+    def _normalize_integer_unix_timestamp(timestamp: str, integer_part: str) -> int | float:
+        """Normalize an integer-only UNIX timestamp to seconds.
+
+        Existing behavior is preserved: timestamps with 10 digits or fewer are
+        interpreted as seconds. Longer values are scaled down so that 10 digits
+        remain before the decimal point.
```
mhoff: AI? "Existing behavior" makes no sense in pydoc as there is no "new" and "old" behavior after merging this change. It will all be part of the "current" behavior. It might be "legacy behavior" if there is a clear intension to change it in the future. But then I'd expect a different wording and a TODO associated with it

logprep/util/time.py
```diff
@@ -144,17 +148,52 @@ def _normalize_unix_timestamp(timestamp: str) -> int | float:
...
+        """Normalize a fractional UNIX timestamp to seconds.
+
+        Fractional UNIX timestamps are parsed as float to preserve sub-second
+        precision. The digit-count heuristic is applied to the integer part only.
+        """
+        parsed_timestamp = float(timestamp)
+        return (
+            parsed_timestamp
+            if len(integer_part) <= 10
+            else parsed_timestamp / 10 ** (len(integer_part) - 10)
```
mhoff: We have now effectively implemented this normalization twice. Assuming we need the heuristical scaling for the float code path as well, it would probably be better to calculate the scaling factor once and then branch depending on the existence of a dot.

```
int_part, decimal_part = timestamp.split(".", maxsplit=1) + [None]

factor = 1 if len(int_part) <= 10 else 10 ** (len(int_part - 10))

if decimal_part is not None:
    return float(timestamp) / factor
else:
    return int(timestamp) / factor
```

But I am honestly not sure if we need the normalization at all.....

mhoff (commented):
Further remarks

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ### Breaking
 
 ### Features
+* add support for fractional UNIX timestamps in the timestamper processor while preserving integer timestamp normalization.
```
```suggestion
* add support for fractional UNIX timestamps in the `timestamper` processor while preserving integer timestamp normalization
```

logprep/util/time.py
```diff
@@ -144,17 +158,82 @@ def _normalize_unix_timestamp(timestamp: str) -> int | float:
...
+            case length if length <= UnixTimestampLength.SECONDS:
+                return 1
+            case UnixTimestampLength.MILLISECONDS:
+                return 10 ** (UnixTimestampLength.MILLISECONDS - UnixTimestampLength.SECONDS)
+            case UnixTimestampLength.MICROSECONDS:
+                return 10 ** (UnixTimestampLength.MICROSECONDS - UnixTimestampLength.SECONDS)
+            case UnixTimestampLength.NANOSECONDS:
+                return 10 ** (UnixTimestampLength.NANOSECONDS - UnixTimestampLength.SECONDS)
+            case _:
+                raise ValueError(f"Unsupported Unix timestamp length: {integer_part_length}")
```
mhoff:
```suggestion
# constant
VALID_UNIX_TIMESTAMP_LENGTHS = set(UnixTimestampLength)

        if integer_part_length in VALID_UNIX_TIMESTAMP_LENGTHS:
            return 10 ** (integer_part_length - UnixTimestampLength.SECONDS)
        else:
            raise ValueError("...")
```

Just a variant which is a little less redundant

CHANGELOG.md
mhoff: Also, this is breaking change

logprep/util/time.py
```diff
@@ -144,17 +158,82 @@ def _normalize_unix_timestamp(timestamp: str) -> int | float:
...
+
+        Parameters
+        ----------
+        integer_part : str
+            The integer part of the UNIX timestamp string.
+
+        Returns
+        -------
+        int
+            Divisor used to normalize the timestamp to seconds.
```
mhoff: I feel like this part of the pydoc is largely redundant with the method name & parameters and return type. As this is a private method within a mathematical context, I'd argue that less (redundant) text gives you more clarity. Different story for public methods

logprep/util/time.py
```diff
@@ -144,17 +158,82 @@ def _normalize_unix_timestamp(timestamp: str) -> int | float:
...
-                else int(timestamp) / 10 ** (len(timestamp) - 10)
-            )
+            timestamp = timestamp.strip()
+            integer_part = TimeParser._get_unix_timestamp_integer_part(timestamp)
+            divisor = TimeParser._get_unix_timestamp_normalization_divisor(integer_part)
+
+            if "." in timestamp:
+                return TimeParser._normalize_fractional_unix_timestamp(timestamp, divisor)
+
+            return TimeParser._normalize_integer_unix_timestamp(timestamp, divisor)
```
mhoff:
```suggestion
            timestamp = timestamp.strip()

            integer_part, fractional_part = timestamp.split(".", maxsplit=1) + [None]
            divisor = TimeParser._get_unix_timestamp_normalization_divisor(integer_part)

            if fractional_part is not None:
                value = float(timestamp)
            else:
                value = int(timestamp)
                
            if divisor == 1:
                return value
            return value / divisor
```

I am also a friend of methods in the clean code style. However, I feel like 1-3 line methods are a bit fine-grained and rather increase complexity.

logprep/util/time.py
```diff
@@ -126,10 +136,14 @@ def _set_utc_if_timezone_is_missing(time_object: datetime) -> datetime:
     def _normalize_unix_timestamp(timestamp: str) -> int | float:
```
mhoff: Maybe put "to_seconds" somewhere in the name, because this is what we are essentially doing. Normalization might mean a lot

logprep/util/time.py
```diff
@@ -9,6 +10,15 @@ class TimeParserException(LogprepException):
     """Exception class for time parsing"""
 
 
+class UnixTimestampLength(IntEnum):
+    """Digit lengths of common Unix timestamp representations."""
```
mhoff:
```suggestion
    """Digit lengths of common UNIX timestamp representations."""
```

mhoff (commented):
Many thanks for the work. I have some important and some maybe rather opinionated suggestions. Feel free to push back on the latter if you feel differently about these points

CHANGELOG.md
kaya-david: Where do you see a breaking change here?

logprep/util/time.py
```diff
@@ -123,38 +133,78 @@ def _set_utc_if_timezone_is_missing(time_object: datetime) -> datetime:
...
         except ValueError as error:
             raise TimeParserException(str(error)) from error
 
+    @staticmethod
+    def _get_unix_timestamp_integer_part(timestamp: str) -> str:
+        """Return the integer part of a UNIX timestamp."""
+        return timestamp.split(".", maxsplit=1)[0]
+
+    @staticmethod
+    @staticmethod
```
mhoff: Redundant

logprep/util/time.py
```diff
@@ -123,38 +133,78 @@ def _set_utc_if_timezone_is_missing(time_object: datetime) -> datetime:
...
+        """Return the divisor for normalizing a supported UNIX timestamp to seconds."""
+
+        integer_part_length = len(integer_part)
+
+        if integer_part_length <= UnixTimestampLength.SECONDS:
+            return 1
+
+        scalable_unix_timestamp_lengths = frozenset(
+            int(length) for length in UnixTimestampLength if length > UnixTimestampLength.SECONDS
+        )
```
mhoff: That's a constant and should only be computed once

logprep/util/time.py
```diff
@@ -123,38 +133,78 @@ def _set_utc_if_timezone_is_missing(time_object: datetime) -> datetime:
...
+
+        if integer_part_length in scalable_unix_timestamp_lengths:
+            return 10 ** (integer_part_length - UnixTimestampLength.SECONDS)
+
+        raise ValueError(f"Unsupported Unix timestamp length: {integer_part_length}")
+
+    @staticmethod
+    def _normalize_fractional_unix_timestamp_to_seconds(timestamp: str, divisor: int) -> float:
```
mhoff: Unused

logprep/util/time.py
```diff
@@ -123,38 +133,78 @@ def _set_utc_if_timezone_is_missing(time_object: datetime) -> datetime:
...
+        """Normalize a fractional UNIX timestamp to seconds.
+
+        Fractional UNIX timestamps are parsed as float to preserve sub-second
+        precision.
+        """
+        return float(timestamp) / divisor
+
+    @staticmethod
+    def _normalize_integer_unix_timestamp_to_seconds(timestamp: str, divisor: int) -> int | float:
```
mhoff: Unused

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 ## Upcoming Changes
 ### Breaking
+* Restrict UNIX timestamp normalization to supported precisions: seconds, milliseconds, microseconds, and nanoseconds. Integer and fractional UNIX timestamps with unsupported digit lengths are now rejected.
```
mhoff: Always start with a lower-case letter in the changelog for consistency. Also this entry is inconsistently long

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
...
 
 ### Features
+* Add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
```
mhoff: Lower-case

logprep/util/time.py
```diff
@@ -123,38 +133,78 @@ def _set_utc_if_timezone_is_missing(time_object: datetime) -> datetime:
...
+
+            if divisor == 1:
+                return value
+
+            return value / divisor
         except ValueError as error:
             raise TimeParserException(str(error)) from error
 
+    @staticmethod
+    def _get_unix_timestamp_integer_part(timestamp: str) -> str:
```
mhoff: Unused

mhoff (commented):
I have added a few remarks

mhoff (approved):
LGTM, thanks!

PR #992, opened 2026-07-15
feat: rework opensearch output to enable retrying, circuit breaking and metrics

mhoff:
## Description

* ng: implement circuit breaker utility to protect downstream systems
* ng: enable opensearch output to retry on transport- and item-level failures + circuit breaker
* ng: add metrics for opensearch output

Configuration changes for Opensearch output:
- removed message_backlog_size
- removed flush_timeout
- removed queue_size
- removed thread_count
- renamed max_retries --> max_client_retries
- added max_bulk_item_retries
- added max_connection_retries
- added initial_backoff
- added max_backoff
- added bulk_concurrency
- added pool_maxsize
- added retryable_item_statuses
- added retryable_transport_statuses

## Assignee

### Documentation

### Code Quality

* pytest and benchmark

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--992.org.readthedocs.build/).

### DIFF
--- a/logprep/ng/connector/opensearch/output.py
+++ b/logprep/ng/connector/opensearch/output.py
@@ -2,13 +2,49 @@
 OpensearchOutput
 ================
 
-This section contains the connection settings for Opensearch, the default
-index, the error index and a buffer size. Documents are sent in batches to Opensearch to reduce
-the amount of times connections are created.
+This section contains the connection settings for Opensearch, the default index, the error index
+and a buffer size. Documents are sent in batches to Opensearch to reduce the amount of times
+connections are created.
 
-The documents desired index is the field :code:`_index` in the document. It is deleted afterwards.
+The documents desired index :code:`_index` can be set using :code:`event.output_target` (or
+:code:`_index` as data field).
+Data fields prefixed with an underscore (like :code:`_index`, :code:`_op_type`) will be fed to
+Opensearch as-is, meaning these will be interpreted as metadata by Opensearch.
+This functionality might change in the future to avoid unintended mixup between pure data and
+Opensearch metadata.
 If you want to send documents to data streams, you have to set the field :code:`_op_type: create` in
 the document.
+
+Sending the data involves different layers:
+- Chunking: split the batch into smaller chunks and process concurrently
+- Item-level reject: send a chunk and retry rejected items until all events are stored or failed
+- Circuit breaker: retry transient connection errors and limit concurrency
+- Opensearch client: manage connection pool and retry with another connection on errors
+
+Retry behavior can be configured in the following ways:
+- Item-level reject:
+  Retries rejected items for a configurable number of times
+  (:code:`max_bulk_item_retries`, indefinitely by default).
+  A retry is only attempted if the item status is deemed retryable
+  (:code:`retryable_item_statuses`, by default 429).
+  Other non-success statuses will lead to corresponding event failures.
+- Circuit breaker:
+  Retries connection-level failures for a configurable number of times
+  (:code:`max_connection_retries`, indefinitely by default).
+  This kind of failure includes retryable request-level status as of Opensearch,
+  timeouts and other connection errors.
+  The circuit breaker opens (=breaks) the circuit on failures and only allows for
+  single probes at a time until the downstream system is deemed healthy again.
+  The breaker also limits maximum concurrency to :code:`bulk_concurrency`.
+  Depending on the nature of failure, retries on this level might lead to duplicates.
+- Opensearch client:
+  Manages a pool of at maximum :code:`pool_maxsize` connections which will be used
+  to handle concurrent connections but also fail-over and retries in case of errors.
+  While timeouts are handled by the circuit breaker, connection errors and other
+  transport-level status codes (like 502, 503 and 504) get retried a maximum of
+  :code:`max_client_retries` times.
+
+Circuit breaker and opensearch client are instance-wide.
 
 Example
 ^^^^^^^
@@ -22,34 +58,43 @@ Example
             - 127.0.0.1:9200
         default_index: default_index
         error_index: error_index
-        message_backlog_size: 10000
         timeout: 10000
-        max_retries:
         user:
         secret:
         ca_cert: /path/to/cert.crt
+        max_bulk_item_retries:          # retry rejected items (429) forever
+        max_connection_retries:         # unbounded; bound it to limit duplicates
+        initial_backoff: 2
+        max_backoff: 600
+        bulk_concurrency: 4             # bulk requests in flight across the output
 """
 
-import contextlib
+import asyncio
 import logging
 import ssl
 import typing
-from collections.abc import Sequence
-from functools import cached_property
-from typing import Any, List, Optional
+from collections.abc import Iterator, Sequence
+from functools import cached_property, partial
+from typing import Any
 
 from attrs import define, field, validators
 from opensearchpy import (
     AsyncOpenSearch,
+)
+from opensearchpy import ConnectionError as OpenSearchConnectionError
+from opensearchpy import (
     OpenSearchException,
     SerializationError,
-    helpers,
+    TransportError,
 )
+from opensearchpy.helpers.actions import expand_action
 from opensearchpy.serializer import JSONSerializer
 
 from logprep.abc.exceptions import LogprepException
+from logprep.metrics.metrics import CounterMetric, HistogramMetric, Metric
 from logprep.ng.abc.event import OutputEvent
 from logprep.ng.abc.output import CriticalOutputError, Output
+from logprep.ng.util.retry import Backoff, BlockingCircuitBreaker, RetriesExhaustedError
 
 logger = logging.getLogger("OpenSearchOutput")
 
@@ -60,7 +105,7 @@ class BulkError(LogprepException):
     def __init__(
         self,
         message: str,
-        status: str | None = None,
+        status: int | str | None = None,
         exception: str | None = None,
     ) -> None:
         super().__init__(message)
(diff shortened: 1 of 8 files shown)


logprep/ng/connector/opensearch/output.py
```diff
@@ -71,17 +116,16 @@ def __str__(self) -> str:
         return f"BulkError: {self.message}, status_code: {self.status}, exception: {self.exception}"
 
 
-class MSGPECSerializer(JSONSerializer):
-    """A MSGPEC serializer"""
+class MsgspecSerializer(JSONSerializer):
+    """Plugs Logprep's msgspec JSON codec into the opensearch-py client."""
 
-    def __init__(self, output_connector: Output, *args, **kwargs):
+    def __init__(self, encoder, decoder, *args, **kwargs):
```
kaya-david: Missing type annotations

logprep/ng/connector/opensearch/output.py
```diff
@@ -92,9 +136,107 @@ def loads(self, s):
...
+
+    def shrink_to(self, indices: list[int]) -> None:
+        """Keeps only the events at the given indices for the next attempt."""
+        self.events = [self.events[i] for i in indices]
+        self.payloads = [self.payloads[i] for i in indices]
+
+    def fail_remaining(self, error: Exception) -> None:
+        """Marks all not yet resolved events of this chunk as failed."""
+        for event in self.events:
+            if not event.stored and not event.is_failed():
```
kaya-david, line 167: Which case is more likely at this point: that the event has failed or that it has already been stored?

Because of short-circuit evaluation, the more likely condition should be checked first:

```python
# If failed events are more common:
if not event.is_failed() and not event.stored:

# If stored events are more common:
if not event.stored and not event.is_failed():
```

logprep/ng/connector/opensearch/output.py
```diff
@@ -122,84 +264,145 @@ class Config(Output.Config):
...
+        )
+        """Item-level response statuses that are retried under
+        :code:`max_bulk_item_retries`. All other non-2xx item statuses are
+        treated as permanent failures. Default is [429], matching the
+        client's own bulk-retry policy. 503 (unavailable shards) can be
+        added: retrying it is duplicate-free, but a persistently red index
+        will stall an unlimited budget."""
+
+        retryable_transport_statuses: list[int] = field(
+            factory=lambda: [502, 503, 504],
```
kaya-david, line 378: It seems that `opensearch-py` uses the following status codes for its retry mechanism:

https://github.com/opensearch-project/opensearch-py/blob/19b6769e63cb9824e754e2f876e98d0fec81e905/test_opensearchpy/test_types/aliased_types.py#L63

We may want to extend our list as well and ensure that it includes OpenSearch’s default retry status codes as a subset.

Could you please double-check whether we actually want to retry for additional status codes or only for `[502, 503, 504]`?

logprep/ng/connector/opensearch/output.py
```diff
@@ -241,87 +477,151 @@ async def setup(self):
...
-            verify_certs=True,  # default is True, but we want to be explicit
+            verify_certs=True,
             timeout=self.config.timeout,
-            serializer=MSGPECSerializer(self),
-            max_retries=self.config.max_retries,
-            pool_maxsize=10,
-            # default for connection pooling is 10 see:
-            # https://github.com/opensearch-project/opensearch-py/blob/main/guides/connection_classes.md
+            serializer=self._serializer,
+            max_client_retries=self.config.max_client_retries,
```
kaya-david:
```suggestion
            max_retries=self.config.max_client_retries,
```
A mistake seems to have slipped in here.

The OpenSearch transport class used internally by AsyncOpenSearch still expects the keyword argument max_retries:

https://github.com/opensearch-project/opensearch-py/blob/19b6769e63cb9824e754e2f876e98d0fec81e905/opensearchpy/transport.py#L80

logprep/ng/connector/opensearch/output.py
```diff
@@ -241,87 +477,151 @@ async def setup(self):
...
+            chunk.fail_remaining(error)
+        except Exception as error:
+            logger.error("critical failure while sending bulk chunk", exc_info=True)
+            chunk.fail_remaining(error)
+
+    @Metric.measure_time_async(metric_name="processing_time_per_event")
+    @Metric.measure_time_async(metric_name="send_time_per_batch")
+    async def _store(self, events: Sequence[OutputEvent]) -> None:
+        logger.debug("Flushing %d documents to opensearch", len(events))
+        await asyncio.gather(*(self._resolve_chunk(chunk) for chunk in self._chunks(events)))
```
kaya-david, line 625: Since the generator is expanded with *, all chunks are serialized and a coroutine is created for each chunk before asyncio.gather() starts. The semaphore limits only the number of concurrent bulk requests; it does not limit the number of pending coroutines or the memory retained by the serialized payloads.

Could this become a problem for large batches by causing excessive peak memory usage? I would not call it a memory leak, since the memory should be released once the operation completes, but the memory consumption could still grow significantly with the batch size.

logprep/ng/connector/opensearch/output.py
```diff
@@ -241,87 +477,151 @@ async def setup(self):
...
+        if retry_indices:
+            self._metrics.number_of_document_rejections += len(retry_indices)
+        chunk.shrink_to(retry_indices)
+        return chunk.is_resolved
+
+    @Metric.measure_time_async(metric_name="send_time_per_bulk_request")
+    async def _send(self, chunk: _Chunk) -> dict[str, Any]:
+        """One bulk network request; counts requests, documents and bytes."""
+        payload = chunk.payload
+        self._metrics.number_of_bulk_requests += 1
```
kaya-david, line 594: According to the description of the `number_of_bulk_requests` metric:

```python
class OpensearchOutput(Output):
    """An OpenSearch output connector."""

    @define(kw_only=True)
    class Metrics(Output.Metrics):
        """Tracks statistics about the OpensearchOutput."""

        number_of_bulk_requests: CounterMetric = field(
            factory=lambda: CounterMetric(
                description="Number of bulk requests sent to Opensearch, including retries",
                name="opensearch_output_number_of_bulk_requests",
            )
        )
```

the metric is supposed to "include retries". However, it is incremented only once per `_send()` call, while `opensearch-py` may perform multiple internal transport attempts within a single `bulk()` call.

Therefore, the phrase `including retries` is not fully accurate, at least for client-level retries. Could we either clarify the metric description or account for the internal transport attempts as well?

logprep/ng/connector/opensearch/output.py
```diff
@@ -122,84 +264,145 @@ class Config(Output.Config):
...
             default=100 * 1024 * 1024,
-            validator=(validators.instance_of(int), validators.gt(1)),
+            validator=[validators.instance_of(int), validators.gt(1)],
         )
         """Max chunk size to use for bulk requests. The default is 100MB."""
 
-        max_retries: int = field(
-            default=3, validator=(validators.instance_of(int), validators.gt(0))
+        max_client_retries: int = field(
+            default=3, validator=[validators.instance_of(int), validators.gt(0)]
```
kaya-david: Do we want to allow retries to be disabled by setting this value to `0`?
```suggestion
            default=3, validator=[validators.instance_of(int), validators.ge(0)]
```

kaya-david (commented):
Thank you for your work. The first review is complete, and I’ve left a few comments.

logprep/ng/connector/opensearch/output.py
mhoff: We don't know. But this is an error scenario either way. I'd hope that errors that affect the whole batch are rather handled on the connection-level and that his rather effects single events. So most of them should be stored, which means the current implementation is fine

logprep/ng/connector/opensearch/output.py
mhoff: Yup, fine by me

logprep/ng/connector/opensearch/output.py
mhoff: Are you sure you looked up the right spot?

https://github.com/opensearch-project/opensearch-py/blob/main/opensearchpy/_async/transport.py#L78

Either way, I'd like to stick to the defaults here and the rest can be configured on deployment

logprep/ng/connector/opensearch/output.py
mhoff: That's a good catch. I actually fixed this once and it got back in.

The transport we are using also uses `max_retries`: https://github.com/opensearch-project/opensearch-py/blob/main/opensearchpy/_async/transport.py#L76

logprep/ng/connector/opensearch/output.py
mhoff: That's true. I have clarified the description

logprep/ng/connector/opensearch/output.py
mhoff: I have added a TODO to watch out for. Generally batch_size, chunk_size and the semaphore should be configured in a way, that there are no large deviations.

It might be the case that I am missing something, but doing step-wise chunking would introduce quite a bit of complexity and/or might require to punch through abstraction layers. The level of concurrency is managed at the circuit breaker level which is an instance-wide construct. Meaning it is shared between default and extra outputs and their `store` invocations. I don't see how a dynamic chunk source would fit in here. So I'd keep it out-of-scope for this PR.

logprep/ng/connector/opensearch/output.py
kaya-david: Oh, my bad. Thanks.

logprep/ng/connector/opensearch/output.py
kaya-david: Fine by me.

kaya-david (approved):
LGTM, thanks!

PR #995, opened 2026-07-17
fix: allow dotted field notation and performance upgrades

Pablu23:
## Description

- DottedTemplate now allows full dotted field notation
- Performance improvements for list comparison processor

## Assignee

### Documentation
- [ ] [Preview for docs](#readthedocs-preview) looks fine

### Code Quality

* pytest

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--995.org.readthedocs.build/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,11 +35,12 @@
 * ng: use `OutputSpec` instead of `list[dict]` to specify desired output targets for extra events
 * ng: drop "ng_" prefix completely and use a setting in the global registry to activate ng
 * ng: decouple console logging via queue/listener in separate thread
+* improve `list_comparison` performance
 
 ### Bugfix
 * fix `dissector` not dissecting multiline strings
 * fix `grokker` dropping matches on duplicate named capture groups
-* fix `list_comparison` to actually work with dotted field notation, and improve performance
+* fix `list_comparison` to actually work with dotted field notation
 
 ## 19.4.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 ### Bugfix
 * fix `dissector` not dissecting multiline strings
 * fix `grokker` dropping matches on duplicate named capture groups
+* fix `list_comparison` to actually work with dotted field notation, and improve performance
 
 ## 19.4.1
 
--- a/logprep/processor/list_comparison/processor.py
+++ b/logprep/processor/list_comparison/processor.py
@@ -129,22 +129,17 @@ class ListComparison(Processor):
         failure tags are applied instead of producing a normal ``not_in_list`` result.
         """
         list_matches: list[str] = []
-        matched_lists: set[str] = set()
         try:
             dynamic_set = rule.get_dynamic_set(event)
         except Exception as error:
             raise ProcessingWarning(str(error), rule, event) from error
 
         for value in value_list:
-            if len(matched_lists) == len(dynamic_set):
-                break
-
             for compare_list, compare_values in dynamic_set.items():
-                if compare_list in matched_lists:
+                if compare_list in list_matches:
                     continue
                 if value in compare_values:
                     list_matches.append(compare_list)
-                    matched_lists.add(compare_list)
 
         return list_matches, dynamic_set
 
--- a/logprep/processor/list_comparison/processor.py
+++ b/logprep/processor/list_comparison/processor.py
@@ -128,18 +128,23 @@ class ListComparison(Processor):
         Dynamic list loading errors are converted to ``ProcessingWarning`` so the rule's
         failure tags are applied instead of producing a normal ``not_in_list`` result.
         """
-        list_matches = []
+        list_matches: list[str] = []
+        matched_lists: set[str] = set()
         try:
             dynamic_set = rule.get_dynamic_set(event)
         except Exception as error:
             raise ProcessingWarning(str(error), rule, event) from error
 
         for value in value_list:
+            if len(matched_lists) == len(dynamic_set):
+                break
+
             for compare_list, compare_values in dynamic_set.items():
-                if compare_list in list_matches:
+                if compare_list in matched_lists:
                     continue
                 if value in compare_values:
                     list_matches.append(compare_list)
+                    matched_lists.add(compare_list)
 
         return list_matches, dynamic_set
 
--- a/logprep/processor/list_comparison/rule.py
+++ b/logprep/processor/list_comparison/rule.py
@@ -43,13 +43,13 @@ target field :code:`List_comparison.example`.
 
 import logging
 import os.path
-from string import Template
 
 from attrs import define, field, validators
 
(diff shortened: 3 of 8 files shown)


logprep/processor/list_comparison/processor.py
```diff
@@ -128,18 +128,23 @@ def _get_lists_matching_with_values(
         Dynamic list loading errors are converted to ``ProcessingWarning`` so the rule's
         failure tags are applied instead of producing a normal ``not_in_list`` result.
         """
-        list_matches = []
+        list_matches: list[str] = []
+        matched_lists: set[str] = set()
```
mhoff: From the naming it is unclear how those are semantically different

Pablu23: This is just supposed to be a little performance improvement on set lookup for if there are alot of lists matching, but not change the return value, we could probably just use a set here and then list(set) when returning

logprep/processor/list_comparison/processor.py
```diff
@@ -128,18 +128,23 @@ def _get_lists_matching_with_values(
...
         try:
             dynamic_set = rule.get_dynamic_set(event)
         except Exception as error:
             raise ProcessingWarning(str(error), rule, event) from error
 
         for value in value_list:
+            if len(matched_lists) == len(dynamic_set):
+                break
```
mhoff: Wouldn't it be smarter to transpose the loop?

Iterate through dynamic_set.items() and check per list if any value matches.
You could short-cut it like this: https://stackoverflow.com/a/2197544/6525140

logprep/processor/list_comparison/rule.py
```diff
@@ -192,11 +195,27 @@ def init_list_comparison(
...
             )
 
+            base_template = DottedTemplate(self._config.list_search_base_path)
+
             # Check if this is a static (eagerly loaded) or dynamic (lazily loaded) list_comparison
             if any(
                 identifier not in [*os.environ, "LOGPREP_LIST"]
-                for identifier in Template(self._config.list_search_base_path).get_identifiers()
+                for identifier in base_template.get_identifiers()
             ):
```
mhoff: A list can also become dynamic by having dynamic entries under list_file_paths, right?

logprep/processor/list_comparison/rule.py
```diff
@@ -192,11 +195,27 @@ def init_list_comparison(
...
+                self._is_dynamic_http = True
+                self._dynamic_templates = tuple(
+                    DottedTemplate(
+                        base_template.safe_substitute({**os.environ, "LOGPREP_LIST": list_path})
+                    )
+                    for list_path in self._config.list_file_paths
+                )
+                self._dynamic_identifiers = tuple(
```
mhoff: The name of this should reflect that *all* dynamic identifiers (across templates) are aggregated here

logprep/util/dotted_template.py
```diff
@@ -0,0 +1,5 @@
+from string import Template
+
+
+class DottedTemplate(Template):
+    braceidpattern = r"(?a:(?:\\.|[^.$\\{}])+(?:\.(?:\\.|[^.$\\{}])+)*)"
```
mhoff: This can go under helper.py or another existing utility module if there is a better fitting one

logprep/processor/list_comparison/processor.py
mhoff: I think we will rather have few lists per rule. Then the constant efforts of maintaining two data structures is greather than the benefit of having a better scaling lookup.

logprep/processor/list_comparison/processor.py
Pablu23: Sorry im unsure of what exactly you mean here :/

logprep/processor/list_comparison/rule.py
```diff
@@ -300,35 +315,32 @@ def get_dynamic_set(self, event: dict) -> dict[str, set]:
...
             loaded, so the processor can apply the rule's failure tags.
         """
+        if not self._is_dynamic_http:
+            return self._compare_sets
+
         compare_sets_result: dict[str, set] = {}
         assert self._config.list_search_base_path
 
         if not self._config.list_search_base_path.startswith("http"):
             return self._compare_sets
```
mhoff: Maybe merge both checks? There is the nuance with non-empty list_search_base_path, but is this actually relevant?

tests/unit/processor/list_comparison/test_list_comparison.py
```diff
@@ -564,6 +564,242 @@ def test_list_comparison_resolves_dynamic_http_template_from_event(self):
...
+                "target_field": "user_results",
+                "list_file_paths": [list_path],
+            },
+            "description": "",
+        }
+        config = {
+            "type": "list_comparison",
+            "rules": [],
+            "list_search_base_path": "http://localhost/${LOGPREP_LIST}",
+        }
```
mhoff, line 600: It's a bit unfortunate that we don't have the typical `TEST_CASES` list which is shared between non-ng and ng, and is used to provide the examples in the documentation. Nothing for this PR, but we should make sure to include this in the future when we touch a processor, even if the task at hand is rather unrelated. Also, we could easily inline the `rule_dict` here into `rules` and avoid the `rule_tree` shenaningans down below. I know this is a copied test but that makes the test way harder to read, less standardized and overly verbose. Feel free to resolve this comment - I just found it important to mention

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
 ### Bugfix
 * fix `dissector` not dissecting multiline strings
 * fix `grokker` dropping matches on duplicate named capture groups
+* fix `list_comparison` to actually work with dotted field notation, and improve performance
```
mhoff: The performance improvement is actually an improvement. Let's make 2 changelog entries for that

logprep/processor/list_comparison/processor.py
mhoff: That's fine.

What I meant is that instead of
```
for value in value_list:
    for compare_list, compare_values in dynamic_set.items():
        # ...
```

we could have done:
```
for compare_list, compare_values in dynamic_set.items():
    for value in value_list:
        # ...
```

Current form:
In the first version, we are taking a field value and check each set if it contains it, adding its name to the result. In the second iteration, we have to avoid re-checking a set which is already present in the result. I though that this check was meant to go somehow into this direction.

Suggestion:
Loop through the sets and then through the values. As soon as you find the first value, you can shortcut, add the set to the result and continue. It probably is also better for caching, as we compare again and again with the same set.

Feel free to neglect (and resolve) this point. I can pick it up when I rework the result structure.

tests/unit/processor/list_comparison/test_list_comparison.py
Pablu23: You are completly right and we should do this, but not in this PR. We should keep that in mind though

mhoff (approved):
LGTM, thanks!

PR #996, opened 2026-07-17
feat: add getter logging & refactor environ handling

mhoff:
## Description

* improve env var access performance by using a cached snapshot
* getter: add debug logs
* improve robustness of compose tests

## Assignee

### Documentation

### Code Quality

* pytest

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--996.org.readthedocs.build/).

### DIFF
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -13,13 +13,6 @@ from logprep.util.environ import ENV_VARS, EnvTemplate
 logger = logging.getLogger("Getter")
 yaml = YAML(typ="safe", pure=True)
 
-BLOCKLIST_VARIABLE_NAMES = [
-    "",
-    " ",
-    "LOGPREP_LIST",  # used by list_comparison processor
-]
-
-VALID_PREFIXES = ["LOGPREP_", "CI_", "GITHUB_", "PYTEST_"]
 
 ContentType: TypeAlias = str
 
@@ -142,7 +135,7 @@ class Getter(ABC):
         """Gets dict and fails otherwise"""
         result = self.get_collection()
         if not isinstance(result, dict):
-            raise ValueError("Value is not a dictionary")
+            raise ValueError(f"Expected a dict, got {type(result)}")
         return result
 
     @staticmethod
@@ -159,7 +152,7 @@ class Getter(ABC):
             content = content[content_field]
         elif content_field is not None:
             raise ValueError(
-                f"Expected mapping type when content_field is set, got {type(content)}."
+                f"Expected mapping type when content_field is set, got {type(content)}"
             )
 
         if isinstance(content, str):
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -1,16 +1,15 @@
 """Module for getter interface"""
 
+import itertools
 import json
 import logging
-import os
-import re
 from abc import ABC, abstractmethod
-from copy import deepcopy
-from string import Template
-from typing import TypeAlias
+from typing import Iterable, TypeAlias
 
 from attrs import define, field, validators
 from ruamel.yaml import YAML, YAMLError
+
+from logprep.util.environ import ENV_VARS, EnvTemplate
 
 logger = logging.getLogger("Getter")
 yaml = YAML(typ="safe", pure=True)
@@ -30,37 +29,24 @@ ContentType: TypeAlias = str
 class Getter(ABC):
     """Abstract base class describing the getter interface and providing of a factory method."""
 
-    class EnvTemplate(Template):
-        """Template class for uppercase only template variables"""
-
-        pattern = r"""
-            \$(?:
-                (?P<escaped>\$\$\$)|
-                (?P<named>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)|
-                {(?P<braced>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)}|
-                (?P<invalid>)
-            )
-        """  # type: ignore[assignment]
-
-        flags = re.VERBOSE
-
     protocol: str = field(validator=validators.instance_of(str))
     """Indicates the protocol for the factory to chose a matching getter."""
     target: str = field(validator=validators.instance_of(str))
     """The target which holds the content to return by get method."""
 
-    missing_env_vars: list = field(
+    missing_env_vars: Iterable[str] = field(
         validator=[
-            validators.instance_of(list),
+            validators.instance_of((list, set)),
             validators.deep_iterable(member_validator=validators.instance_of(str)),
         ],
-        factory=list,
+        factory=set,
         repr=False,
     )
     """used variables in content but not set in environment"""
 
     @property
     def uri(self) -> str:
+        """Full-length URI of the getter target, including the protocol"""
         return f"{self.protocol}://{self.target}"
 
     def get(self) -> str:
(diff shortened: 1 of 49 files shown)


logprep/abc/getter.py
```diff
@@ -70,30 +56,28 @@ def get(self) -> str:
...
-    def _get_used_env_vars(self, content, template):
-        found_variables = template.pattern.findall(
-            content
-        )  # returns a list of tuples in form (escaped, named, braced, invalid)
+        template = EnvTemplate(content)
+        resolved_content = template.safe_substitute(self._get_kwargs(template, content))
+        logger.debug("resolved environment placeholders in content: %s", resolved_content)
+        return resolved_content
+
+    def _get_kwargs(self, template: EnvTemplate, content: str):
```
kaya-david: Missing return annotation

logprep/abc/getter.py
```diff
@@ -174,8 +167,7 @@ def get_list(self, content_field: str | None = None) -> list:
             content = content[content_field]
         elif content_field is not None:
             raise ValueError(
```
kaya-david, line 154: All methods log before raising an exception. Why is there no `logger.debug()` or `logger.warning()` here?

logprep/util/cache.py
```diff
@@ -1,9 +1,8 @@
 """Module for caching items and checking if they need to be stored (again)."""
 
-from typing import Union
-
 import datetime
 from collections import OrderedDict
+from typing import Union
```
kaya-david: Could you remove this `Union` and use the `A | B` syntax instead?

logprep/util/environ.py
```diff
@@ -0,0 +1,43 @@
+"""Utility module for handling environment variables"""
+
+import os
+import re
+from string import Template
+from types import MappingProxyType
+
+_ENV_SNAPSHOT = dict(os.environ)
```
kaya-david, line 8: Does the snapshot need some kind of synchronization mechanism? During a configuration refresh, new environment variables could potentially be loaded at runtime, causing the snapshot to pick up changes.

logprep/util/environ.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def set_env_var(name: str, value: str):
+    """Set an env var in :code:`os.environ` and in the cache"""
+    os.environ[name] = value
+    _ENV_SNAPSHOT[name] = value
+
+
+class EnvTemplate(Template):
+    """Template class for uppercase only template variables"""
```
kaya-david:
```suggestion
    """Template for uppercase environment variables with supported prefixes."""
```

logprep/util/environ.py
```diff
@@ -0,0 +1,43 @@
...
+            {(?P<braced>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)}|
+            (?P<invalid>)
+        )
+    """  # type: ignore[assignment]
+
+    flags = re.VERBOSE
+
+    @property
+    def compiled_pattern(self) -> re.Pattern[str]:
+        """Return the compiled internal pattern used to identify vars"""
```
kaya-david:
```suggestion
        """Handle supported uppercase environment variables, excluding LOGPREP_LIST variables."""
```

logprep/util/getter.py
```diff
@@ -33,6 +30,11 @@
     ENV_NAME_LOGPREP_CREDENTIALS_FILE,
     ENV_NAME_LOGPREP_GETTER_CONFIG,
 )
+from logprep.util.environ import ENV_VARS
+
+logger = logging.getLogger("Getter")
+
+rg_logger = logging.getLogger("RefreshableGetter")
```
kaya-david, line 37: Do we really need to use two different loggers here? Could this be reduced to a single logger?

kaya-david (commented):
Thanks for the work. At first glance, it looks good. I do have a few comments, though.

logprep/util/getter.py
mhoff: This is legacy behavior, as this specific logger has existed before. In my opinion we should refactor all logger names and use the dot-based hierarchy native to python logging. But this is a (breaking) refactoring step for ops which is out-of-scope for this PR

logprep/util/environ.py
mhoff: I will keep this pydoc short and add the details about excluded vars on class level. Otherwise we have redundant info here.

Thinking about it once again. `Template` has a functiojnality to get all identifiers. I wonder why we even need to expose the compiled pattern here. I will revisit this.

logprep/util/environ.py
mhoff: Environment variables are AFAIK not affected through config refreshes. It is rather the other way around; environment variables are injected in each loaded config

logprep/util/environ.py
mhoff: Added a lengthy explanation on the class level

logprep/util/environ.py
mhoff: Done, I have removed the method and replaced it with `Template.get_identifiers()`. No need to do it ourselves if it's already implemented

logprep/abc/getter.py
mhoff: I have decided to improve our own error messages where there was a need and only introduce debug statements at the other raise-sites, not touching the internal exceptions for now

logprep/abc/getter.py
```diff
@@ -70,30 +48,21 @@ def get(self) -> str:
...
+    def _get_kwargs(self, template: EnvTemplate) -> dict[str, str]:
+        used_env_vars = set(template.get_identifiers())
+        self.missing_env_vars = used_env_vars.difference(ENV_VARS.keys())
+        return ENV_VARS | {missing_key: "" for missing_key in self.missing_env_vars}
 
     def _resolve_content_by_content_type(self) -> dict | list | str:
         """Get content with enriched environment variables parsed based on content type."""
 
         raw, content_type = self._get_raw()
+        # TODO make resolving env variables optional, especially for user-defiend pure data payloads
```
Pablu23:
```suggestion
        # TODO make resolving env variables optional, especially for user-defined pure data payloads
```

logprep/util/environ.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    pattern = r"""
+        \$(?:
+            (?P<escaped>\$\$\$)|
+            (?P<named>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)|
+            {(?P<braced>(?!LOGPREP_LIST)(?=LOGPREP_|CI_|GITHUB_|PYTEST_)[_A-Z0-9]*)}|
+            (?P<invalid>)
+        )
+    """  # type: ignore[assignment]
```
Pablu23, line 42: Would it be easier and more readable to only adjust braceditpattern instead of the full pattern?

logprep/util/getter.py
```diff
@@ -488,24 +502,28 @@ def _get_raw(self) -> tuple[bytes, ContentType | None]:
         return self.cache, self.content_type
 
     def keep_alive(self):
-        self.shared.last_called = time.monotonic()
+        """Signal a data access, rendering :code:`timeout_interval` to reset for this target"""
+        self.shared.keep_alive()
 
     def timed_out(self) -> bool:
+        """Whether the target has timed out"""
```
Pablu23:
```suggestion
        """Whether this target has timed out"""
```

logprep/util/getter.py
```diff
@@ -517,29 +535,43 @@ def remove_callbacks_for_tag(cls, tag: str) -> None:
...
                 for callback in shared_target_data.cleanup_callbacks:
                     callback["function"](*callback["args"], **callback["kwargs"])
                 continue
 
             if shared_target_data.scheduler:
                 shared_target_data.scheduler.run_pending()
 
+    @classmethod
+    def reset(cls, cleanup: bool = False):
+        """Wipe the cache and optionall run cleanup callbacks"""
```
Pablu23:
```suggestion
        """Wipe the cache and optionally run cleanup callbacks"""
```

Pablu23 (commented):
Look good, just a few smaller sidenotes

logprep/util/environ.py
mhoff: Might be the case, but I will this untouched for now, as I simply moved the class. I'd presume this is part of how we handle the env templating in general and if we even want to keep this class

logprep/abc/getter.py
mhoff: Seems like I need a spell checker, thanks :D

logprep/util/environ.py
Pablu23: > Might be the case, but I will this untouched for now, as I simply moved the class. I'd presume this is part of how we handle the env templating in general and if we even want to keep this class

Sure this seems fine to me, but I think we should look over this at some point

Pablu23 (approved):
LGTM

PR #997, opened 2026-07-21
feat: return list names in list & network comparison; fix rule failure bug

mhoff:
## Description

* fix `list_comparison` & `network_comparison` to actually work with dotted field notation
* fix `list_comparison` & `network_comparison` rule failure handling for multiple lists
* allow `list_comparison` & `network_comparison` to be supplied with explicit list names for results

## Assignee

### Documentation

### Code Quality

* pytest

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--997.org.readthedocs.build/).


logprep/ng/processor/list_comparison/processor.py
```diff
@@ -123,34 +125,50 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list[
             else [field_value_to_be_checked]
         )
 
-        list_matches, compare_sets = self._get_lists_matching_with_values(rule, value_list, event)
-
-        if len(list_matches) == 0:
-            return list(compare_sets.keys()), "not_in_list"
-        return list_matches, "in_list"
+        matching_keys, all_keys = self._get_lists_matching_with_values(rule, value_list, event)
+        if len(matching_keys) == 0:
```
kaya-david:
```suggestion
        if not matching_keys:
```
More Pythonic.

logprep/processor/list_comparison/processor.py
```diff
@@ -115,33 +120,50 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list,
...
+        if len(matching_keys) == 0:
+            return list(all_keys), "not_in_list"
+        return matching_keys, "in_list"
 
     def _get_lists_matching_with_values(
-        self, rule: ListComparisonRule, value_list: list, event: dict
-    ) -> tuple[list, dict[str, set]]:
-        """Return matching comparison-list identifiers and the evaluated compare sets.
+        self, rule: ListComparisonRule, values: Sequence[FieldValue], event: dict[str, FieldValue]
+    ) -> tuple[list[str], Iterable[str]]:
```
kaya-david: Why is the first element of the returned tuple explicitly typed as `list[str]`, while the second is typed more generally as `Iterable[str]`?

logprep/processor/list_comparison/processor.py
```diff
@@ -115,33 +120,50 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list,
...
-                if compare_list in list_matches:
-                    continue
-                if value in compare_values:
-                    list_matches.append(compare_list)
+        return matches, rule.compare_set_names
+
+    # pylint: disable=unused-argument
+    def _prepare_event_values_for_match(
+        self, values: Sequence[FieldValue], rule: ListComparisonRule, event: dict[str, FieldValue]
+    ) -> typing.Any:
```
kaya-david: Could we avoid `Any` here and use a more specific return type, such as `set[FieldValue]`?

logprep/processor/list_comparison/processor.py
```diff
@@ -115,33 +120,50 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list,
...
-            for compare_list, compare_values in dynamic_set.items():
-                if compare_list in list_matches:
-                    continue
-                if value in compare_values:
-                    list_matches.append(compare_list)
+        return matches, rule.compare_set_names
+
+    # pylint: disable=unused-argument
+    def _prepare_event_values_for_match(
+        self, values: Sequence[FieldValue], rule: ListComparisonRule, event: dict[str, FieldValue]
```
kaya-david: 1.
```suggestion
        self, values: Sequence[FieldValue], rule: _, event: __[str, FieldValue]
```

2.

```
        self, values: Sequence[FieldValue], rule: ListComparisonRule, event: dict[str, FieldValue]
```

Would `Iterable[FieldValue]` be sufficient here, or does the implementation require sequence-specific operations such as indexing, `len()`, or multiple iterations?

More generally, we seem to be introducing increasingly generic types in some signatures while still using concrete types such as `list` and `set` in others. Should we define a consistent rule and use the least specific type that still reflects the function's actual requirements?

logprep/processor/list_comparison/processor.py
```diff
@@ -115,33 +120,50 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list,
...
+    ) -> typing.Any:
+        """Convert the raw source-field values into the form used for matching.
+
+        Invalid values are reported as warnings and skipped by subclasses that need
+        to parse the values first.
+        """
+        return set(values)
+
+    # pylint: enable=unused-argument
```
kaya-david: relict?

logprep/processor/list_comparison/processor.py
```diff
@@ -115,33 +120,50 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list,
...
 
-        return list_matches, dynamic_set
+    @staticmethod
+    def _matches_compare_set(prepared_values: typing.Any, set_values: set) -> bool:
```
kaya-david:
```suggestion
    def _matches_compare_set(prepared_values: set[typing.Any], set_values: set[typing.Any]) -> bool:
```
Or differently, but please be a bit more precise.

logprep/processor/list_comparison/rule.py
```diff
@@ -184,124 +327,139 @@ def init_list_comparison(
...
+        if not isinstance(getter, RefreshableGetter):
+            raise TypeError(f"The target {uri} must be a url")
+
+        update_func = functools.partial(
+            self._update_compare_sets_via_http, getter=getter, compare_set=compare_set
         )
+
+        update_func()
         tag = self._callback_tag
+        key = (tag, uri, id(self))
```
kaya-david: I do not fully understand the underlying logic here yet, so I used AI to help me understanding the underlying logic and to analyze the changes. During that analysis, I came across the following potential issue.

I would appreciate discussing this with you so I can better assess the situation. In the meantime, please feel free to review the analysis below and help me put the findings into context:

Previously:

```python
key = (tag, uri, id(self))
```

This worked because the content was cached once per URI:

```python
_compare_sets[uri] = content
```

For example:

```python
self._compare_sets = {
    "https://example.org/shared-list": {
        "Franz",
        "Heinz",
        "Hans",
    }
}
```

One URI represented exactly one cached content set, correct?

Therefore, one refresh callback per URI was sufficient.
Now, each static compare set stores its own content in a separate object:

```python
self._static_sets = [
    _StaticCompareSet(
        name="LIST_A",
        content={"Franz", "Heinz", "Hans"},
        error=None,
    ),
    _StaticCompareSet(
        name="LIST_B",
        content={"Franz", "Heinz", "Hans"},
        error=None,
    ),
]
```

The URI is not stored in `_StaticCompareSet` itself. Instead, each compare set is separately connected to the same URI through `_load_and_refresh_uri()`:

```python
self._load_and_refresh_uri(
    self._static_sets[0],
    "https://example.org/shared-list",
)

self._load_and_refresh_uri(
    self._static_sets[1],
    "https://example.org/shared-list",
)
```

Although both compare sets use the same URI, they have separate content storage. Therefore, the callback key also needs to distinguish them:

```python
key = (tag, uri, id(self), id(compare_set))
```

Otherwise, both callbacks receive the same deduplication key:

```python
key_a = (
    tag,
    "https://example.org/shared-list",
    id(self),
)

key_b = (
    tag,
    "https://example.org/shared-list",
    id(self),
)
```

The second callback may therefore be deduplicated. After a refresh, the actual state could then look like this:

```python
self._static_sets = [
    _StaticCompareSet(
        name="LIST_A",
        content={"Franz", "Heinz"},
        error=None,
    ),
    _StaticCompareSet(
        name="LIST_B",
        content={"Franz", "Heinz", "Hans"},
        error=None,
    ),
]
```

In this example, `LIST_B` still contains stale content.

An alternative would be to keep caching content once per URI and let both compare sets reference the same cached content. A possible structure could look like this:

```python
self._contents_by_uri = {
    "https://example.org/shared-list": {
        "Franz",
        "Heinz",
        "Hans",
    }
}

self._compare_sets = [
    {
        "name": "LIST_A",
        "uri": "https://example.org/shared-list",
    },
    {
        "name": "LIST_B",
        "uri": "https://example.org/shared-list",
    },
]
```

Both compare sets would then access the same content through their URI:

```python
content = self._contents_by_uri[compare_set["uri"]]
```

This may be cleaner architecturally because the shared content would only need to be loaded and refreshed once, but it would require a larger change.

logprep/processor/network_comparison/rule.py
```diff
@@ -80,12 +86,59 @@ class Config(FieldManagerRule.Config):
...
+
+        If unset, the processor-level ``list_search_base_path`` is used. A base path must
+        be configured either on the rule or on the processor.
+
+        The value may use getter syntax and ``string.Template`` placeholders.
+        Environment variables and ``${LOGPREP_LIST}`` are resolved during setup. For
+        HTTP(S) paths, unresolved placeholders are resolved from event fields during
+        processing.
+        """
+        mapping: dict = field(default={}, init=False, repr=False, eq=False)
```
kaya-david:
```suggestion
        mapping: dict = field(factory=dict, init=False, repr=False, eq=False)
```
Why was `factory=dict` changed to `default={}` here?

With `default={}`, all config instances share the same mutable dictionary. Changes made through one instance would therefore also be visible in other instances. Should this remain `factory=dict`?

logprep/ng/processor/list_comparison/processor.py
```diff
@@ -123,34 +125,50 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list[
...
         """
-        list_matches = []
+
+        prepared_values = self._prepare_event_values_for_match(values, rule, event)
+
         try:
-            dynamic_set = rule.get_dynamic_set(event)
+            matches = [
+                set_name
+                for set_name, set_values in rule.iter_compare_sets(event)
```
kaya-david, line 147: This changes the order of matching list names.

Previously, the result order depended on the order of the source values. Now, it follows the order in which the compare sets are iterated.

Could this behavior change cause any issues for us? Do we rely on the previous ordering anywhere?

logprep/ng/processor/list_comparison/processor.py
```diff
@@ -123,34 +125,50 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list[
...
-            for compare_list, compare_values in dynamic_set.items():
-                if compare_list in list_matches:
-                    continue
-                if value in compare_values:
-                    list_matches.append(compare_list)
+        return matches, rule.compare_set_names
+
+    # pylint: disable=unused-argument
+    def _prepare_event_values_for_match(
+        self, values: Sequence[FieldValue], rule: ListComparisonRule, event: dict[str, FieldValue]
```
kaya-david: Remove the # pylint: disable=unused-argument comment above and prefix the unused parameters with an underscore instead:
```suggestion
        self, values: Sequence[FieldValue], _: ListComparisonRule, __: dict[str, FieldValue]
```

logprep/ng/processor/list_comparison/processor.py
```diff
@@ -123,34 +125,50 @@ def _list_comparison(self, rule: ListComparisonRule, event: dict) -> tuple[list[
...
+    ) -> typing.Any:
```
kaya-david: Could we avoid `Any` here and use a more specific return type, such as `set[FieldValue]`?

logprep/processor/list_comparison/rule.py
```diff
@@ -184,124 +327,139 @@ def init_list_comparison(
...
+    def _init_list_comparison_from_local_file(
+        self, base_path: str, list_paths: Sequence[str], list_names: Sequence[str] | None
+    ):
+        if not base_path.endswith("/"):
+            base_path = base_path + "/"
+
+        absolute_paths = [path if path.startswith("/") else base_path + path for path in list_paths]
+
+        if list_names is None:
+            list_names = [os.path.basename(path) for path in absolute_paths]
```
kaya-david, line 380: What happens if multiple entries in `list_file_paths` have the same basename?

For example:

```text
department_a/users.txt
department_b/users.txt
```

Both compare sets would be named `users.txt`, so the result could contain duplicate and indistinguishable names such as:

```python
["users.txt", "users.txt"]
```

Should derived compare-set names be validated for uniqueness?

logprep/processor/list_comparison/rule.py
```diff
@@ -39,38 +80,96 @@
...
+
+
+class ContentFailure(Enum):
+    """Sentinel type for indicating missing fields."""
+
+    CONTENT_FAILURE = auto()
+
+
+CONTENT_FAILURE = ContentFailure.CONTENT_FAILURE  # pylint: disable=invalid-name
+"""Sentinel value for indicating failed content retrievals"""
```
kaya-david: `ContentFailure` and `CONTENT_FAILURE` do not seem to be used anywhere. Also, the docstrings describe different purposes: one refers to missing fields, while the other refers to failed content retrievals.

Is this leftover code that can be removed, or is it intended to be used later?

logprep/processor/list_comparison/rule.py
```diff
@@ -184,124 +327,139 @@ def init_list_comparison(
...
+            content = self._get_list_contents_from_getter(getter)
+            compare_set.update_content(getter.target, content)
         except Exception as ex:
-            if mark_rule_failed:
-                self.mark_failed(error=ex)
-                return None
+            if isinstance(compare_set, _StaticCompareSet):
+                self._mark_failed(compare_set, error=ex)
+                return
+            raise ex
```
kaya-david:
```suggestion
            raise
```

Could we use a bare `raise` here instead of `raise ex` to preserve the original traceback?

logprep/processor/list_comparison/rule.py
```diff
@@ -184,124 +327,139 @@ def init_list_comparison(
...
+            content = self._get_list_contents_from_getter(GetterFactory.from_string(list_path))
+            self._static_sets.append(_StaticCompareSet(name=list_name, content=content))
+
+        self.compare_set_names = list_names
+
+    def _init_list_comparison_from_http(
+        self, base_path: str, list_paths: Sequence[str], list_names: Sequence[str] | None
+    ):
+        base_template = DottedTemplate(base_path)
+        all_dynamic_identifiers: set[str] = set()
```
kaya-david, line 404: Do we want to preserve the order of the dynamic identifiers here?

Using a `set` makes the order nondeterministic. If multiple required fields are missing or invalid, this may change which field is reported first and could make error messages or tests unstable.

The previous implementation used `dict.fromkeys(...)` to deduplicate the identifiers while preserving their order.

logprep/processor/list_comparison/rule.py
```diff
@@ -184,124 +327,139 @@ def init_list_comparison(
...
+            if isinstance(compare_set, _StaticCompareSet):
+                self._mark_failed(compare_set, error=ex)
+                return
+            raise ex
+        else:
+            # TODO ugly check in hot path, better idea?
+            if isinstance(compare_set, _StaticCompareSet):
+                self._clear_failed(compare_set)
+
+    def _recompute_failure_state(self):
```
kaya-david: 1.
```suggestion
    def _recompute_failure_state(self) -> None:
```

2.
Could we add a test covering the failure and recovery of multiple static compare sets?

```python
def test_rule_remains_failed_until_all_static_compare_sets_recover(rule):
    compare_set_a = _StaticCompareSet(name="LIST_A", content=set())
    compare_set_b = _StaticCompareSet(name="LIST_B", content=set())
    rule._static_sets = [compare_set_a, compare_set_b]

    error_a = RuntimeError("LIST_A failed")
    error_b = RuntimeError("LIST_B failed")

    rule._mark_failed(compare_set_a, error_a)
    rule._mark_failed(compare_set_b, error_b)

    assert isinstance(rule.data_error, ExceptionGroup)
    assert rule.data_error.exceptions == (error_a, error_b)

    rule._clear_failed(compare_set_a)
    assert rule.data_error is error_b

    rule._clear_failed(compare_set_b)
    assert rule.data_error is None
```

logprep/processor/list_comparison/rule.py
```diff
@@ -2,33 +2,74 @@
...
+Relative paths are transformed to absolute paths by using :code:`list_search_base_path` as a prefix.
+A forward-slash is appended to :code:`list_search_base_path` before joining (if it has none).
+The identifying name for each list is either taken from the key of :code:`list_paths` (if specified)
+or it is derived from the basename of each path.
+
+For **HTTP(s) paths**, :code:`list_search_base_path` has to carry a `${LOGPREP_LIST}` placeholder
+which is used to inject the sub-paths from :code:`list_paths` or :code:`list_file_paths` literally.
+Also, environment variables are interpolated in the process, for instance if the data source for the
+lists is an API, for which domain/host/port etc. are supplied via the environment.
+Environment variables are limited to TODO insert link to relevant docs here here.
```
kaya-david: There is still an unresolved TODO in the documentation:

```text
Environment variables are limited to TODO insert link to relevant docs here here.
```

Could we replace this with the actual documentation reference or remove the sentence for now?

logprep/processor/list_comparison/rule.py
```diff
@@ -184,124 +327,139 @@ def init_list_comparison(
...
+        self, base_path: str, list_paths: Sequence[str], list_names: Sequence[str] | None
+    ):
+        base_template = DottedTemplate(base_path)
+        all_dynamic_identifiers: set[str] = set()
+
+        if list_names is None:
+            list_names = list_paths
+
+        for name, list_path in zip(list_names, list_paths):
+            full_path = base_template.safe_substitute(LOGPREP_LIST=list_path)
```
kaya-david, line 410: The documentation states that HTTP(S) base paths must contain the `${LOGPREP_LIST}` placeholder, but this is not validated.

Without the placeholder, all configured list paths resolve to the same URI and the individual `list_paths` values are ignored.

Should we validate that `${LOGPREP_LIST}` is present for HTTP(S) base paths, or adjust the documentation if this is intentional?

logprep/processor/list_comparison/rule.py
```diff
@@ -39,38 +80,96 @@
...
+CONTENT_FAILURE = ContentFailure.CONTENT_FAILURE  # pylint: disable=invalid-name
+"""Sentinel value for indicating failed content retrievals"""
+
+
+@define(kw_only=True)
+class _CompareSet(ABC):
+    name: ListName
+
+    @abstractmethod
+    def update_content(self, uri: str, content: ListContent | None):
```
kaya-david:
```suggestion
    def update_content(self, uri: str, content: ListContent | None) -> None:
```

logprep/processor/list_comparison/rule.py
```diff
@@ -39,38 +80,96 @@
...
+    def update_content(self, _, content: ListContent | None):
+        self.content = content
+
+
+@define(kw_only=True)
+class _DynamicCompareSet(_CompareSet):
+    uri_template: DottedTemplate
+    uri_to_content: dict[str, ListContent] = field(factory=dict)
+
+    def update_content(self, uri: str, content: ListContent | None):
```
kaya-david:
```suggestion
    def update_content(self, uri: str, content: ListContent | None) -> None:
```

logprep/processor/list_comparison/rule.py
```diff
@@ -39,38 +80,96 @@
...
+        if content is None:
+            self.remove_content(uri)
+        else:
+            self.uri_to_content[uri] = content
+
+    def remove_content(self, uri: str):
```
kaya-david:
```suggestion
    def remove_content(self, uri: str) -> None:
```

logprep/processor/list_comparison/rule.py
```diff
@@ -184,124 +327,139 @@ def init_list_comparison(
...
-            # Check if this is a static (eagerly loaded) or dynamic (lazily loaded) list_comparison
-            if dynamic_identifiers:
-                self._is_dynamic_http = True
-                self._dynamic_templates = resolved_templates
-                self._all_dynamic_identifiers = dynamic_identifiers
-                return
+            self._init_list_comparison_from_http(base_path, list_paths, list_names)
 
-            self._init_static_http_list_comparison(resolved_templates)
+    def _add_static_compare_set(self, name: ListName, path: str):
```
kaya-david:
```suggestion
    def _add_static_compare_set(self, name: ListName, path: str) -> None:
```

logprep/processor/list_comparison/rule.py
kaya-david:
```suggestion
    def _recompute_failure_state(self) -> None:
```

logprep/processor/network_comparison/rule.py
```diff
@@ -126,35 +179,14 @@ class Config(FieldManagerRule.Config):
...
-            for list_path, compare_strings in self._compare_sets.items()
-            if compare_strings
-        }
+        def __attrs_post_init__(self):
+            if self.list_file_paths and self.list_paths:
+                raise ValueError("`list_file_paths` and `list_paths` must not both be specified")
+            if not self.list_file_paths and not self.list_paths:
+                raise ValueError("one of `list_file_paths` or `list_paths` needs to be specified")
+
+    def _transform_and_filter_list_element(self, elem):
```
kaya-david: Missing type annotation.

Like this?
```suggestion
    def _transform_and_filter_list_element(self, elem: str) -> str|None:
```

Please check,

logprep/processor/list_comparison/rule.py
```diff
@@ -2,33 +2,74 @@
...
-is used to identify the field which is to be checked against the provided lists.
-And the latter is used to define the parent field where the results should
-be written to. Both fields can be dotted subfields.
-
-Additionally, a list or array of lists can be provided underneath the
-required field :code:`list_file_paths`.
+Processor-specific rule configuration is done using the mandatory field :code:`list_comparison`.
+
+The fields :code:`source_fields` and :code:`target_field` are used to define which field value
+flows into the comparison and where the results flows to, respectively.
```
kaya-david:
```suggestion
flows into the comparison and where the results flow to, respectively.
```

logprep/processor/list_comparison/rule.py
```diff
@@ -2,33 +2,74 @@
...
+Also, environment variables are interpolated in the process, for instance if the data source for the
+lists is an API, for which domain/host/port etc. are supplied via the environment.
+Environment variables are limited to TODO insert link to relevant docs here here.
+Additionally, field values can be injected into the paths using the same notation and
+(potentially dotted) field references.
+A URI with field references is considered *dynamic* and has considerable performance implications,
+as every new concrete path needs to be fetched ad-hoc during event processing.
+Caching and cache timeouts are used to balance performance and memory demands.
+
+Results of the list comparison are writting to :code:`target_field`.
```
kaya-david:
```suggestion
Results of the list comparison are written to :code:`target_field`.
```

logprep/processor/list_comparison/rule.py
```diff
@@ -2,33 +2,74 @@
...
+If there was any intersection between :code:`source_fields` and any list, a sub-field called
+`"in_list"` is populated, which in turn holds a list of all matching list names.
+If there was no intersection, `"not_in_list"` is populated instead with all list names which
+were tested in the comparison.
+Do note that there is always only one of both responses.
```
kaya-david:
```suggestion
Do note that only one of these responses is present at a time.
```

logprep/processor/list_comparison/rule.py
```diff
@@ -88,6 +187,44 @@ class Config(FieldManagerRule.Config):
...
+                key_validator=validators.instance_of(ListName),
+                value_validator=validators.instance_of(str),
+                mapping_validator=validators.instance_of(dict),
+            ),
+            factory=dict,
+        )
+        """
+        Mapping for configuring list paths with representative names.
+        Keys represent the names on which results will be reported.
+        Values represent the paths which populates `${LOGPREP_LIST}`.
```
kaya-david:
```suggestion
        Values represent the paths used to populate `${LOGPREP_LIST}`.
```

kaya-david (commented):
Thank you for your work.

Since I do not yet have a deep understanding of the previous list comparison implementation and the way the dynamic paths are connected, it was difficult for me to review some parts in sufficient depth. I therefore used AI as a tool in a few places to better understand the logic and support my review. I have clearly disclosed this in the relevant comments.

A short debrief would be very helpful for me, either now or after you have gone through my comments. Wherever anything is unclear, we can of course discuss it at any time.

logprep/ng/processor/list_comparison/processor.py
mhoff: I think we are claiming nothing about the order of the result. So this should be fine

logprep/ng/processor/list_comparison/processor.py
mhoff: Good suggestion, didn't think of that in the moment

logprep/ng/processor/list_comparison/processor.py
mhoff: Unfortunately we can't. NetworkComparison reuses this code with a different result type. This would be a similar issue like with the rule type. So we have to stick to a barebone solution at the moment. At least I don't see a better way

logprep/processor/list_comparison/processor.py
mhoff: Good point. The iterable was a remainder of returning a dict_keys object. Both are Sequence now

logprep/processor/list_comparison/processor.py
mhoff: I generally like `Iterable` but it has two problems:
1) no `len` or straight-foward emptiness check
2) it can hide a generator which can only be consumed once, leading to subtle bugs

`Sequence` is most often the correct choice

I am primarily using less concrete types or read-only types with new code. But when adapting old code I don't always refactor the whole type signature as this can produce ripples in other parts of the code. I

logprep/processor/list_comparison/processor.py
mhoff: Nope, as explained above

logprep/processor/list_comparison/processor.py
mhoff: Nope, here I am activating the check again which has been disable a few lines above. I am removing this in favor of the _ and __ solution

logprep/processor/network_comparison/rule.py
mhoff: Yes, my mistake. Definitely factory=dict

logprep/processor/list_comparison/processor.py
mhoff: I have put the concrete types and will instead put a mypy ignore in it's place. We can not handle the variadic interface coupling as it is (and without the generic processor rewrite)

logprep/processor/list_comparison/rule.py
mhoff: Yes, we had a discussion about this this morning. Removed for now, no restrictions in place

logprep/processor/list_comparison/rule.py
mhoff: Good catch. Too many rewrites. Removed

logprep/processor/list_comparison/rule.py
mhoff: Sure, added to test_list_comparison_rule.py; worked like a charm :)

logprep/processor/list_comparison/rule.py
mhoff: Good catch. Added a check for this

logprep/processor/list_comparison/rule.py
mhoff: Check and test added. Thanks!

logprep/processor/list_comparison/rule.py
mhoff: Yep, this is a bug in an edge-case scenario. If the user configures two times the same path inside the same rule, then the deduplication key is the same.

Actually it should be sufficient to use `(uri, id(compare_set))`

logprep/processor/list_comparison/rule.py
mhoff: The order is not relevant. This is simply an optimization to fetch relevant event values just once for all involved dynamic lists. Each list path will only use a subset of those values in their individual order, handled by their template. Nothing to worry here I'd say

logprep/ng/processor/list_comparison/processor.py
mhoff: I have inserted the concrete types now. Probably we need a mypy excemption now

kaya-david (approved):
LGTM, thanks!

PR #998, opened 2026-07-22
feat: add dynamic url info

Pablu23:
## Description

* Briefly summarize your changes in a few bullet points (can and should correspond to [CHANGELOG.md](https://github.com/fkie-cad/Logprep/blob/main/CHANGELOG.md))
* Relates to #XXX (insert issue number here), if there is a corresponding GH issue

## Assignee

### Documentation

### Code Quality

* Ran tests

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--998.org.readthedocs.build/).

!-- readthedocs-preview logprep end -->

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--998.org.readthedocs.build/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Features
 * add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
 * introduce API-level support for asynchronous rule processing and I/O capability detection in `ng` processors
+* generic_adder: add support for templated http urls
 
 ### Improvements
 * docs: enable pydoc placeholders for facilitating component reuse through inheritance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 ### Features
 * add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
 * introduce API-level support for asynchronous rule processing and I/O capability detection in `ng` processors
-* generic_adder: add support for templated http urls
+* generic_adder: add support for templated http urls & content_field
 
 ### Improvements
 * docs: enable pydoc placeholders for facilitating component reuse through inheritance
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -114,9 +114,17 @@ class Getter(ABC):
         content = self._resolve_content(raw)
         return self._parse_json(content)
 
-    def get_collection(self) -> dict | list:
+    def get_collection(self, content_field: str | None = None) -> dict | list:
         """Gets and parses the raw content to yaml or json"""
         content = self._resolve_content_by_content_type()
+
+        if isinstance(content, dict) and content_field is not None:
+            content = content[content_field]
+        elif content_field is not None:
+            raise ValueError(
+                "Expected mapping type, like json object when content_field is set, got %s.",
+                type(content),
+            )
 
         if isinstance(content, str):
             content = self._parse_yaml_or_json(content)
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -118,13 +118,7 @@ class Getter(ABC):
         """Gets and parses the raw content to yaml or json"""
         content = self._resolve_content_by_content_type()
 
-        if isinstance(content, dict) and content_field is not None:
-            content = content[content_field]
-        elif content_field is not None:
-            raise ValueError(
-                "Expected mapping type, like json object when content_field is set, got %s.",
-                type(content),
-            )
+        content = Getter._apply_content_field(content, content_field)
 
         if isinstance(content, str):
             content = self._parse_yaml_or_json(content)
@@ -151,17 +145,25 @@ class Getter(ABC):
         """Helper which tries to convert content to list"""
         return content.splitlines()
 
-    def get_list(self, content_field: str | None = None) -> list:
-        """Gets list and fails otherwise"""
-
-        content = self._resolve_content_by_content_type()
-
+    @staticmethod
+    def _apply_content_field(
+        content: dict | list | str, content_field: str | None = None
+    ) -> dict | list | str:
         if isinstance(content, dict) and content_field is not None:
             content = content[content_field]
         elif content_field is not None:
             raise ValueError(
                 f"Expected mapping type when content_field is set, got {type(content)}"
             )
+
+        return content
+
+    def get_list(self, content_field: str | None = None) -> list:
+        """Gets list and fails otherwise"""
+
+        content = self._resolve_content_by_content_type()
+
+        content = Getter._apply_content_field(content, content_field)
 
         if isinstance(content, str):
             content = self._parse_newline_separated_list(content)
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -118,8 +118,10 @@ class Getter(ABC):
         """Gets and parses the raw content to yaml or json"""
         content = self._resolve_content_by_content_type()
 
-        content = Getter._apply_content_field(content, content_field)
+        if isinstance(content, str):
+            content = self._parse_yaml_or_json(content)
 
+        content = Getter._apply_content_field(content, content_field)
         if isinstance(content, str):
             content = self._parse_yaml_or_json(content)
 
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -133,9 +133,9 @@ class Getter(ABC):
             logger.debug("parsing yaml failed, falling back to json for content: %s", content)
             return self._parse_json(content)
 
-    def get_dict(self) -> dict:
+    def get_dict(self, content_field: str | None = None) -> dict:
         """Gets dict and fails otherwise"""
-        result = self.get_collection()
+        result = self.get_collection(content_field=content_field)
         if not isinstance(result, dict):
             raise ValueError(f"Expected a dict, got {type(result)}")
         return result
@@ -160,9 +160,7 @@ class Getter(ABC):
 
     def get_list(self, content_field: str | None = None) -> list:
         """Gets list and fails otherwise"""
-
         content = self._resolve_content_by_content_type()
-
         content = Getter._apply_content_field(content, content_field)
 
         if isinstance(content, str):
--- a/logprep/abc/getter.py
+++ b/logprep/abc/getter.py
@@ -163,6 +163,9 @@ class Getter(ABC):
     def get_list(self, content_field: str | None = None) -> list:
         """Gets list and fails otherwise"""
         content = self._resolve_content_by_content_type()
+        if content_field is not None and isinstance(content, str):
+            content = self._parse_yaml_or_json(content)
+
         content = Getter._apply_content_field(content, content_field)
 
         if isinstance(content, str):
--- a/logprep/ng/processor/generic_adder/processor.py
+++ b/logprep/ng/processor/generic_adder/processor.py
@@ -31,8 +31,8 @@ from logprep.ng.abc.processor import Processor
 from logprep.processor.base.exceptions import ProcessingWarning
 from logprep.processor.base.rule import Rule
 from logprep.processor.generic_adder.rule import GenericAdderRule
-from logprep.util.helper import FieldValue, add_fields_to
 from logprep.util.getter import RefreshableGetter
+from logprep.util.helper import FieldValue, add_fields_to
 
 
 class GenericAdder(Processor):
@@ -40,7 +40,6 @@ class GenericAdder(Processor):
 
     rule_class = GenericAdderRule
 
-    async def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
     @property
     def rules(self) -> Sequence[GenericAdderRule]:
         """Returns all rules"""
@@ -51,7 +50,7 @@ class GenericAdder(Processor):
         for rule in self.rules:
             rule.init_generic_adder(self._job_tag_for_cleanup)
 
-    def _apply_rules(self, event: dict, rule: Rule) -> None:
+    async def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
         rule = typing.cast(GenericAdderRule, rule)
 
         try:
--- a/logprep/ng/processor/generic_adder/processor.py
+++ b/logprep/ng/processor/generic_adder/processor.py
@@ -25,11 +25,13 @@ Processor Configuration
 """
 
 import typing
+from typing import Sequence
 
 from logprep.ng.abc.processor import Processor
 from logprep.processor.base.rule import Rule
 from logprep.processor.generic_adder.rule import GenericAdderRule
 from logprep.util.helper import FieldValue, add_fields_to
+from logprep.util.getter import RefreshableGetter
 
 
 class GenericAdder(Processor):
@@ -38,7 +40,22 @@ class GenericAdder(Processor):
     rule_class = GenericAdderRule
 
     async def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
+    @property
+    def _rules(self) -> Sequence[GenericAdderRule]:
+        """Returns all rules"""
+        return typing.cast(Sequence[GenericAdderRule], self.rules)
+
+    async def setup(self):
+        await super().setup()
+        for rule in self._rules:
+            rule.init_generic_adder(self._job_tag_for_cleanup)
+
+    def _apply_rules(self, event: dict, rule: Rule) -> None:
         rule = typing.cast(GenericAdderRule, rule)
         items_to_add = rule.add(event)
         if items_to_add:
             add_fields_to(event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target)
+
+    def _shut_down(self) -> None:
+        RefreshableGetter.remove_callbacks_for_tag(self._job_tag_for_cleanup)
+        return super()._shut_down()
--- a/logprep/ng/processor/generic_adder/processor.py
+++ b/logprep/ng/processor/generic_adder/processor.py
@@ -25,7 +25,7 @@ Processor Configuration
 """
 
 import typing
-from typing import Sequence
+from collections.abc import Sequence
 
 from logprep.ng.abc.processor import Processor
 from logprep.processor.base.exceptions import ProcessingWarning
@@ -42,13 +42,13 @@ class GenericAdder(Processor):
 
     async def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
     @property
-    def _rules(self) -> Sequence[GenericAdderRule]:
+    def rules(self) -> Sequence[GenericAdderRule]:
         """Returns all rules"""
-        return typing.cast(Sequence[GenericAdderRule], self.rules)
+        return typing.cast(Sequence[GenericAdderRule], super().rules)
 
     async def setup(self):
         await super().setup()
-        for rule in self._rules:
+        for rule in self.rules:
             rule.init_generic_adder(self._job_tag_for_cleanup)
 
     def _apply_rules(self, event: dict, rule: Rule) -> None:
--- a/logprep/ng/processor/generic_adder/processor.py
+++ b/logprep/ng/processor/generic_adder/processor.py
@@ -28,6 +28,7 @@ import typing
 from typing import Sequence
 
 from logprep.ng.abc.processor import Processor
+from logprep.processor.base.exceptions import ProcessingWarning
 from logprep.processor.base.rule import Rule
 from logprep.processor.generic_adder.rule import GenericAdderRule
 from logprep.util.helper import FieldValue, add_fields_to
@@ -52,7 +53,11 @@ class GenericAdder(Processor):
 
     def _apply_rules(self, event: dict, rule: Rule) -> None:
         rule = typing.cast(GenericAdderRule, rule)
-        items_to_add = rule.add(event)
+
+        try:
+            items_to_add = rule.add(event)
+        except Exception as error:
+            raise ProcessingWarning(str(error), rule, event) from error
         if items_to_add:
             add_fields_to(event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target)
 
--- a/logprep/ng/processor/generic_adder/processor.py
+++ b/logprep/ng/processor/generic_adder/processor.py
@@ -39,6 +39,6 @@ class GenericAdder(Processor):
 
     async def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
         rule = typing.cast(GenericAdderRule, rule)
-        items_to_add = rule.add
+        items_to_add = rule.add(event)
         if items_to_add:
             add_fields_to(event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target)
--- a/logprep/ng/processor/generic_adder/processor.py
+++ b/logprep/ng/processor/generic_adder/processor.py
@@ -55,11 +55,13 @@ class GenericAdder(Processor):
         rule = typing.cast(GenericAdderRule, rule)
 
         try:
-            items_to_add = rule.add(event)
+            for items_to_add in rule.additions(event):
+                if items_to_add:
+                    add_fields_to(
+                        event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target
+                    )
         except Exception as error:
             raise ProcessingWarning(str(error), rule, event) from error
-        if items_to_add:
-            add_fields_to(event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target)
 
     def _shut_down(self) -> None:
         RefreshableGetter.remove_callbacks_for_tag(self._job_tag_for_cleanup)
--- a/logprep/processor/generic_adder/processor.py
+++ b/logprep/processor/generic_adder/processor.py
@@ -41,7 +41,7 @@ class GenericAdder(Processor):
         super().setup()
         for rule in self.rules:
             rule = typing.cast(GenericAdderRule, rule)
-            rule.set_job_tag(self._job_tag_for_cleanup)
+            rule.init_generic_adder(self._job_tag_for_cleanup)
 
     def _apply_rules(self, event: dict, rule: Rule):
         rule = typing.cast(GenericAdderRule, rule)
--- a/logprep/processor/generic_adder/processor.py
+++ b/logprep/processor/generic_adder/processor.py
@@ -37,8 +37,14 @@ class GenericAdder(Processor):
 
     rule_class = GenericAdderRule
 
+    def setup(self):
+        super().setup()
+        for rule in self.rules:
+            rule = typing.cast(GenericAdderRule, rule)
+            rule.set_job_tag(self._job_tag_for_cleanup)
+
     def _apply_rules(self, event: dict, rule: Rule):
         rule = typing.cast(GenericAdderRule, rule)
-        items_to_add = rule.add
+        items_to_add = rule.add(event)
         if items_to_add:
             add_fields_to(event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target)
--- a/logprep/processor/generic_adder/processor.py
+++ b/logprep/processor/generic_adder/processor.py
@@ -25,10 +25,11 @@ Processor Configuration
 """
 
 import typing
+from typing import Sequence
 
 from logprep.abc.processor import Processor
-from logprep.processor.base.rule import Rule
 from logprep.processor.generic_adder.rule import GenericAdderRule
+from logprep.util.getter import RefreshableGetter
 from logprep.util.helper import add_fields_to
 
 
@@ -37,14 +38,21 @@ class GenericAdder(Processor):
 
     rule_class = GenericAdderRule
 
+    @property
+    def _rules(self) -> Sequence[GenericAdderRule]:
+        """Returns all rules"""
+        return typing.cast(Sequence[GenericAdderRule], self.rules)
+
     def setup(self):
         super().setup()
-        for rule in self.rules:
-            rule = typing.cast(GenericAdderRule, rule)
+        for rule in self._rules:
             rule.init_generic_adder(self._job_tag_for_cleanup)
 
-    def _apply_rules(self, event: dict, rule: Rule):
-        rule = typing.cast(GenericAdderRule, rule)
+    def _apply_rules(self, event: dict, rule: GenericAdderRule):
         items_to_add = rule.add(event)
         if items_to_add:
             add_fields_to(event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target)
+
+    def _shut_down(self) -> None:
+        RefreshableGetter.remove_callbacks_for_tag(self._job_tag_for_cleanup)
+        return super()._shut_down()
--- a/logprep/processor/generic_adder/processor.py
+++ b/logprep/processor/generic_adder/processor.py
@@ -31,7 +31,7 @@ from logprep.abc.processor import Processor
 from logprep.processor.base.exceptions import ProcessingWarning
 from logprep.processor.generic_adder.rule import GenericAdderRule
 from logprep.util.getter import RefreshableGetter
-from logprep.util.helper import add_fields_to
+from logprep.util.helper import FieldValue, add_fields_to
 
 
 class GenericAdder(Processor):
@@ -49,7 +49,7 @@ class GenericAdder(Processor):
         for rule in self.rules:
             rule.init_generic_adder(self._job_tag_for_cleanup)
 
-    def _apply_rules(self, event: dict, rule: GenericAdderRule):
+    def _apply_rules(self, event: dict[str, FieldValue], rule: GenericAdderRule):
         try:
             for items_to_add in rule.additions(event):
                 if items_to_add:
--- a/logprep/processor/generic_adder/processor.py
+++ b/logprep/processor/generic_adder/processor.py
@@ -28,6 +28,7 @@ import typing
 from typing import Sequence
 
 from logprep.abc.processor import Processor
+from logprep.processor.base.exceptions import ProcessingWarning
 from logprep.processor.generic_adder.rule import GenericAdderRule
 from logprep.util.getter import RefreshableGetter
 from logprep.util.helper import add_fields_to
@@ -49,7 +50,10 @@ class GenericAdder(Processor):
             rule.init_generic_adder(self._job_tag_for_cleanup)
 
     def _apply_rules(self, event: dict, rule: GenericAdderRule):
-        items_to_add = rule.add(event)
+        try:
+            items_to_add = rule.add(event)
+        except Exception as error:
+            raise ProcessingWarning(str(error), rule, event) from error
         if items_to_add:
             add_fields_to(event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target)
 
--- a/logprep/processor/generic_adder/processor.py
+++ b/logprep/processor/generic_adder/processor.py
@@ -25,7 +25,7 @@ Processor Configuration
 """
 
 import typing
-from typing import Sequence
+from collections.abc import Sequence
 
 from logprep.abc.processor import Processor
 from logprep.processor.base.exceptions import ProcessingWarning
@@ -40,13 +40,13 @@ class GenericAdder(Processor):
     rule_class = GenericAdderRule
 
     @property
-    def _rules(self) -> Sequence[GenericAdderRule]:
+    def rules(self) -> Sequence[GenericAdderRule]:
         """Returns all rules"""
-        return typing.cast(Sequence[GenericAdderRule], self.rules)
+        return typing.cast(Sequence[GenericAdderRule], super().rules)
 
     def setup(self):
         super().setup()
-        for rule in self._rules:
+        for rule in self.rules:
             rule.init_generic_adder(self._job_tag_for_cleanup)
 
     def _apply_rules(self, event: dict, rule: GenericAdderRule):
--- a/logprep/processor/generic_adder/processor.py
+++ b/logprep/processor/generic_adder/processor.py
@@ -51,11 +51,13 @@ class GenericAdder(Processor):
 
     def _apply_rules(self, event: dict, rule: GenericAdderRule):
         try:
-            items_to_add = rule.add(event)
+            for items_to_add in rule.additions(event):
+                if items_to_add:
+                    add_fields_to(
+                        event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target
+                    )
         except Exception as error:
             raise ProcessingWarning(str(error), rule, event) from error
-        if items_to_add:
-            add_fields_to(event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target)
 
     def _shut_down(self) -> None:
         RefreshableGetter.remove_callbacks_for_tag(self._job_tag_for_cleanup)
--- a/logprep/processor/generic_adder/rule.py
+++ b/logprep/processor/generic_adder/rule.py
@@ -87,9 +87,47 @@ import typing
 
 from attrs import define, field, validators
 
+from logprep.filter.expression.filter_expression import FilterExpression
 from logprep.processor.base.rule import InvalidRuleDefinitionError
 from logprep.processor.field_manager.rule import FieldManagerRule
-from logprep.util.getter import GetterFactory, RefreshableGetter
+from logprep.util.converters import convert_from_dict
+from logprep.util.dotted_template import DottedTemplate
+from logprep.util.getter import GetterFactory, HttpGetter, RefreshableGetter
+from logprep.util.helper import FieldValue, get_dotted_field_value
+
+
+@define(kw_only=True, frozen=True)
+class AddFromUrlConfig:
+    url: str = field(
+        validator=[validators.instance_of(str), validators.matches_re(r"^https?://.+")]
+    )
+
+    target_field: str | None = field(
+        default=None, validator=validators.optional(validators.instance_of(str))
+    )
+
+    target_field_mapping: dict[str, str] | None = field(
+        validator=validators.optional(
+            validators.deep_mapping(
+                key_validator=validators.instance_of(str),
+                value_validator=validators.instance_of(str),
+            )
+        ),
+        default=None,
+    )
+
+    def __attrs_post_init__(self) -> None:
+        if not self.target_field and not self.target_field_mapping:
+            raise ValueError("add_from_url requires target_field or target_field_mapping")
+
+
+def _convert_add_from_url(
+    value: AddFromUrlConfig | dict | None,
+) -> AddFromUrlConfig | None:
+    if value is None:
+        return None
+
+    return convert_from_dict(AddFromUrlConfig, value)
 
 
 class GenericAdderRule(FieldManagerRule):
@@ -113,13 +151,15 @@ class GenericAdderRule(FieldManagerRule):
         """Contains a dictionary of field names and values that should be added.
         If dot notation is being used, then all fields on the path are being
         automatically created."""
-        add_from_file: list = field(
-            validator=[
-                validators.instance_of(list),
-                validators.deep_iterable(member_validator=validators.instance_of(str)),
-            ],
-            converter=lambda x: x if isinstance(x, list) else [x],
-            factory=list,
+        add_from_file: list | None = field(
+            validator=validators.optional(
+                validators.deep_iterable(
+                    iterable_validator=validators.instance_of(list),
+                    member_validator=validators.instance_of(str),
+                )
+            ),
+            converter=lambda x: x if x is None or isinstance(x, list) else [x],
+            default=None,
             eq=False,
         )
         """Contains the path or url to YML file that contains a dictionary of field names
@@ -141,6 +181,19 @@ class GenericAdderRule(FieldManagerRule):
            authenticity and integrity of the loaded values.
 
         """
+
+        add_from_url: AddFromUrlConfig | None = field(
+            default=None,
+            validator=validators.optional(
+                validators.instance_of(AddFromUrlConfig)
+                # validators.deep_iterable(
+                #     iterable_validator=validators.instance_of(list),
+                #     member_validator=validators.instance_of(AddFromUrlConfig),
+                # )
+            ),
+            converter=_convert_add_from_url,
+        )
+
         only_first_existing_file: bool = field(
             validator=validators.instance_of(bool), default=False, eq=False
         )
@@ -160,6 +213,37 @@ class GenericAdderRule(FieldManagerRule):
         def __attrs_post_init__(self):
             self._base_add = copy.deepcopy(self.add)
 
+            if (
+                self.add_from_file is not None or self.add is not None
+            ) and self.add_from_url is not None:
+                raise ValueError(
+                    "only one of add_from_file + add or add_from_url is allowed per GenericAdder rule"
+                )
+
+            if not self.add and self.add_from_file is None and self.add_from_url is None:
+                raise ValueError(
+                    "one of add, add_from_file or add_from_url is neccessary per GenericAdder rule"
+                )
+
+            if (
+                self.add_from_url is not None
+                and self.add_from_url.target_field is not None
+                and self.add_from_url.target_field_mapping is not None
+            ):
+                raise ValueError(
+                    "only one of target_field or target_field_mapping is allowed per GenericAdder rule"
+                )
+
+            if (
+                self.add_from_url is not None
+                and self.add_from_url.target_field is None
+                and self.add_from_url.target_field_mapping is None
+            ):
+                raise ValueError(
+                    "one of target_field or target_field_mapping is neccessary per GenericAdder rule"
+                )
+
+            # Eagerly loaded from file
             if self.add_from_file:
                 for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
                     getter = GetterFactory.from_string(add_file)
@@ -173,6 +257,7 @@ class GenericAdderRule(FieldManagerRule):
         def _add_from_path(self):
             """Reads add fields from file"""
             missing_files = []
+            assert self.add_from_file is not None
             for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
                 try:
                     add_dict = GetterFactory.from_string(add_file).get_yaml()
@@ -195,8 +280,80 @@ class GenericAdderRule(FieldManagerRule):
                     f"The following required files do not exist: '{missing_files}'"
                 )
 
-    @property
-    def add(self) -> dict:
+    def __init__(self, filter_rule: FilterExpression, config: Config, processor_name: str):
+        super().__init__(filter_rule, config, processor_name)
+        self._dynamic_content: dict[str, FieldValue] = {}
+        self._callback_tag = ""
+
+    def set_job_tag(self, job_tag: str) -> None:
+        self._callback_tag = job_tag
+
+    def _dynamic_add_from_url(self, event: dict) -> dict[str, FieldValue]:
+        config = typing.cast(GenericAdderRule.Config, self._config)
+
+        assert config.add_from_url
+
+        dynamic_path_template = DottedTemplate(config.add_from_url.url)
+        key_val = {
+            identifier: get_dotted_field_value(event, identifier)
+            for identifier in dynamic_path_template.get_identifiers()
+        }
+        for identifier, val in key_val.items():
+            if val is None:
+                raise ValueError(
+                    f"missing event field {identifier!r} for dynamic list comparison path"
+                )
+            if not isinstance(val, (str, int)):
+                raise ValueError(
+                    f"value for list comparison field {identifier!r} is not a scalar value"
+                )
+            pass
+
+        dynamic_resolved = dynamic_path_template.substitute(key_val)
+        content: FieldValue | None = None
+        if dynamic_resolved not in self._dynamic_content:
+            http_getter = GetterFactory.from_string(dynamic_resolved)
+            assert isinstance(http_getter, HttpGetter)
+
+            http_getter.keep_alive()
+
+            content = http_getter.get_collection()
+            self._dynamic_content[dynamic_resolved] = content
+
+            # Get tag here
+            # Add update callback
+            # Add cleanup callback
+        else:
+            RefreshableGetter.keep_alive_for_target(dynamic_resolved)
+            content = self._dynamic_content[dynamic_resolved]
+
+        items_to_add: dict[str, FieldValue] = {}
+
+        if config.add_from_url.target_field:
+            items_to_add[config.add_from_url.target_field] = content
+        else:
+            assert config.add_from_url.target_field_mapping is not None
+
+            # TODO: Check this differently, what should this be?
+            assert isinstance(content, dict)
+            for (
+                mapping_source_field,
+                mapping_target_field,
+            ) in config.add_from_url.target_field_mapping.items():
+                # what should happen with missing fields? Right now None is skipped
+                # change this to get_dotted_field_value_with_missing
+                items_to_add[mapping_target_field] = get_dotted_field_value(
+                    content, mapping_source_field
+                )
+
+        return items_to_add
+
+    def add(self, event: dict) -> dict:
         """Returns the fields to add"""
         config = typing.cast(GenericAdderRule.Config, self._config)
-        return config.add
+
+        if config.add_from_file is not None:
+            return config.add
+
+        assert config.add_from_url is not None
+        return self._dynamic_add_from_url(event)
--- a/logprep/processor/generic_adder/rule.py
+++ b/logprep/processor/generic_adder/rule.py
@@ -116,7 +116,7 @@ class AddFromUrlConfig:
 
     def __attrs_post_init__(self) -> None:
         if not self.target_field and not self.target_field_mapping:
-            raise ValueError("add_from_url requires target_field or target_field_mapping")
+            raise ValueError("adding values from url requires target_field or target_field_mapping")
 
 
 def _convert_add_from_url(
@@ -149,15 +149,13 @@ class GenericAdderRule(FieldManagerRule):
         """Contains a dictionary of field names and values that should be added.
         If dot notation is being used, then all fields on the path are being
         automatically created."""
-        add_from_file: list | None = field(
-            validator=validators.optional(
-                validators.deep_iterable(
-                    iterable_validator=validators.instance_of(list),
-                    member_validator=validators.instance_of(str),
-                )
+        add_from_file: list[str] = field(
+            validator=validators.deep_iterable(
+                iterable_validator=validators.instance_of(list),
+                member_validator=validators.instance_of(str),
             ),
-            converter=lambda x: x if x is None or isinstance(x, list) else [x],
-            default=None,
+            converter=lambda x: x if isinstance(x, list) else [x],
+            factory=list,
             eq=False,
         )
         """Contains the path or url to YML file that contains a dictionary of field names
@@ -236,21 +234,18 @@ class GenericAdderRule(FieldManagerRule):
                 )
 
             # Eagerly loaded from file
-            if self.add_from_file:
-                for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
-                    getter = GetterFactory.from_string(add_file)
-                    if isinstance(getter, RefreshableGetter):
-                        # TODO: This never gets cleaned up, Memory leak on a lot of new generic adders / generic resolvers
-                        getter.add_callback(
-                            f"generic_adder:{self.id}:{add_file}", self._refresh_add
-                        )
-                    self._add_from_path()
+            for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
+                getter = GetterFactory.from_string(add_file)
+                if isinstance(getter, RefreshableGetter):
+                    # TODO: This never gets cleaned up, Memory leak on a lot of new generic adders / generic resolvers
+                    getter.add_callback(f"generic_adder:{self.id}:{add_file}", self._refresh_add)
+                self._add_from_path()
 
         def _add_from_path(self):
             """Reads add fields from file"""
             missing_files = []
-            assert self.add_from_file is not None
-            for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
+
+            for add_file in self.add_from_file:
                 try:
                     add_dict = GetterFactory.from_string(add_file).get_yaml()
                 except FileNotFoundError:
@@ -343,6 +338,7 @@ class GenericAdderRule(FieldManagerRule):
                 deduplication_key=(tag, dynamic_resolved, id(self)),
                 fnc_args=[dynamic_resolved],
             )
+
         else:
             RefreshableGetter.keep_alive_for_target(dynamic_resolved)
             content = self._dynamic_content[dynamic_resolved]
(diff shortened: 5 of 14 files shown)


logprep/abc/getter.py
```diff
@@ -137,10 +137,18 @@ def get_json(self) -> dict | list:
...
+    def get_collection(self, content_field: str | None = None) -> dict | list:
         """Gets and parses the raw content to yaml or json"""
         content = self._resolve_content_by_content_type()
 
+        if isinstance(content, dict) and content_field is not None:
+            content = content[content_field]
+        elif content_field is not None:
+            raise ValueError(
+                "Expected mapping type, like json object when content_field is set, got %s.",
+                type(content),
```
mhoff: use f-template for errors

logprep/processor/generic_adder/processor.py
```diff
@@ -37,8 +37,14 @@ class GenericAdder(Processor):
 
     rule_class = GenericAdderRule
 
+    def setup(self):
+        super().setup()
+        for rule in self.rules:
+            rule = typing.cast(GenericAdderRule, rule)
```
mhoff: Maybe this instead? At some point we will have generics here and can then simply delete this casting property

```
    @property
    def rules(self) -> Sequence[PseudonymizerRule]:
        """Returns all rules"""
        return typing.cast(Sequence[PseudonymizerRule], super().rules)
```

logprep/abc/getter.py
```diff
@@ -137,10 +137,18 @@ def get_json(self) -> dict | list:
...
+            )
```
mhoff: Let's introduce a utility method for this, like `_apply_content_field`. We will use this method in several top level functions

logprep/processor/generic_adder/processor.py
```diff
@@ -37,8 +37,14 @@ class GenericAdder(Processor):
...
+            rule.init_generic_adder(self._job_tag_for_cleanup)
```
mhoff, line 50: I wonder if we really need this external init and would like to avoid it. The `list_comparison` has it, as the `list_search_base_path` comes from the processor and is relevant for the function of the rule. Here rule could generate it's own job tag, couldn't it?

logprep/processor/generic_adder/rule.py
```diff
@@ -83,13 +83,49 @@
...
+        validator=validators.deep_mapping(
+            key_validator=validators.instance_of(str),
+            value_validator=validators.instance_of(str),
+        ),
+        factory=dict,
+    )
+
+    def __attrs_post_init__(self) -> None:
+        if not self.target_field and not self.target_field_mapping:
+            raise ValueError("add_from_url requires target_field or target_field_mapping")
```
mhoff: Uncertain about this point, but typically you'd want to avoid putting "add_from_url" here, because the class theoretically does not know about this name. Your choice. I understand why putting it there is helpful to users

```suggestion
            raise ValueError("target_field or target_field_mapping required")
```

logprep/processor/generic_adder/rule.py
```diff
@@ -113,13 +149,15 @@ class Config(FieldManagerRule.Config):
...
+        add_from_file: list | None = field(
+            validator=validators.optional(
+                validators.deep_iterable(
+                    iterable_validator=validators.instance_of(list),
+                    member_validator=validators.instance_of(str),
+                )
+            ),
+            converter=lambda x: x if x is None or isinstance(x, list) else [x],
+            default=None,
             eq=False,
```
mhoff, line 190: Not sure about the precise use case for eq=False... but we would theoretically also need it for add_from_url as well

logprep/processor/generic_adder/rule.py
```diff
@@ -160,6 +205,37 @@ def _refresh_add(self):
...
+                and self.add_from_url.target_field is None
+                and len(self.add_from_url.target_field_mapping) <= 0
+            ):
+                raise ValueError(
+                    "one of target_field or target_field_mapping is neccessary per GenericAdder rule"
+                )
+
+            # Eagerly loaded from file
             if self.add_from_file:
                 for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
```
mhoff:
```suggestion
                for add_file in self.add_from_file:
```

Should work, right?

logprep/processor/generic_adder/rule.py
```diff
@@ -173,6 +249,7 @@ def __attrs_post_init__(self):
         def _add_from_path(self):
             """Reads add fields from file"""
             missing_files = []
+            assert self.add_from_file is not None
             for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
```
mhoff:
```suggestion
            for add_file in self.add_from_file:
```

Should work, right?

tests/unit/processor/generic_adder/test_generic_adder.py
```diff
@@ -303,8 +303,8 @@
             "comp\\lex.field": "value",
             "comp\\lex.nested": {"field": 42},
             "nested": {"comp\\lex.field": 1337},
-            "\\u\\0\\1\\x\y": 1338,  # pylint: disable=anomalous-backslash-in-string
-            "\\u\\0\\1\\x\z": "whatever",  # pylint: disable=anomalous-backslash-in-string
+            "\\u\\0\\1\\x\\y": 1338,  # pylint: disable=anomalous-backslash-in-string
+            "\\u\\0\\1\\x\\z": "whatever",  # pylint: disable=anomalous-backslash-in-string
```
mhoff:
```suggestion
            "\\u\\0\\1\\x\\y": 1338,
            "\\u\\0\\1\\x\\z": "whatever",
```

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +272,115 @@ def _add_from_path(self):
...
+        config = typing.cast(GenericAdderRule.Config, self._config)
+        if config.add_from_file or config.add:
+            return
+
+        assert config.add_from_url is not None
+
+        base_template = DottedTemplate(config.add_from_url.url)
+        resolved_template = DottedTemplate(base_template.safe_substitute({**os.environ}))
+        self._dynamic_template = resolved_template
+        self._dynamic_identifiers = tuple(resolved_template.get_identifiers())
```
mhoff: Here is the obvious question: do we want to base the `GenericAdder` on the latest changes we did on `ListComparison`? I know this is unfortunate as some parts of the efforts are redundant. One aspect we need to integrate at some point is the rule failure handling for static lists (which we do not have here yet). Happy to discuss this.

For the moment I will not look deeper into the implementation, as this might become partially obsolete

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +272,115 @@ def _add_from_path(self):
...
+            assert config.add_from_url.target_field_mapping is not None
+
+            # TODO: Check this differently, what should this be?
+            assert isinstance(content, dict)
+            for (
+                mapping_source_field,
+                mapping_target_field,
+            ) in config.add_from_url.target_field_mapping.items():
+                # what should happen with missing fields? Right now None is skipped
+                # change this to get_dotted_field_value_with_missing
```
mhoff: I would like to go towards having `None` a valid value. At least I don't like simply ignoring field without a value. So, MISSING handling would feel like the right choice here

tests/unit/processor/generic_adder/test_generic_adder_rule.py
mhoff, line 16: `test_cases` and `failure_test_cases` would be nice here as well. I figure that this is a scope creep; so probably we drop it for this PR

logprep/processor/generic_adder/processor.py
Pablu23: > I wonder if we really need this external init and would like to avoid it. The `list_comparison` has it, as the `list_search_base_path` comes from the processor and is relevant for the function of the rule. Here rule could generate it's own job tag, couldn't it?

Yes we do, I forgot to add the `_shut_down` function here, this makes it clearer why we need to init here

logprep/processor/generic_adder/rule.py
Pablu23: > Not sure about the precise use case for eq=False... but we would theoretically also need it for add_from_url as well

Yeah im unsure about this aswell, would we just want to remove the eq here then?

tests/unit/processor/generic_adder/test_generic_adder_rule.py
Pablu23: > `test_cases` and `failure_test_cases` would be nice here as well. I figure that this is a scope creep; so probably we drop it for this PR

Yes thats right, but I would rather have this be in another PR / task of itself

logprep/processor/generic_adder/rule.py
mhoff: Maybe remove it and see what happens. I'd figure rule equality might have something to do with the rule tree or the auto rule tester. Might also be a question we could place Monday. I suspect we can remove it, tbh. Either way, remove it completely or add it to both fields - no in between

Pablu23: > Maybe remove it and see what happens. I'd figure rule equality might have something to do with the rule tree or the auto rule tester. Might also be a question we could place Monday. I suspect we can remove it, tbh. Either way, remove it completely or add it to both fields - no in between

Okay I asked ai, and it seems like eq=False says dont check this field when comparing two rules against each other. Why we add that here and nowhere else is because, eager loading of a rule file adds the content to add which DOES get checked for equality. Meaning two rules, one with a static add, and one pointing at a file with the same content, are both evaulating to be the same rule. In that case I would actually leave eq=False for `add_from_file` only so that logic stays, but that doesnt make sense for `add_from_url` as we cannot eagerly load that.

logprep/processor/generic_adder/processor.py
mhoff: Mh, I see where you are coming from. Alright, we can keep this for consistency with other processors (namely `list_comparison`).

But I still think that the `job_tag` and managing callbacks  is rather an implementation detail of the rule, which it should handle itself. I think it would be consequential, that processor.setup() calls rule.setup() and processor.shut_down() calls rule.shut_down(), whereas managing callbacks is completely handled by the rule.

Currently we are calling an init (which has setup-like semantics) on the rule but basically handle the shutdown on the processor-level, which is inconsistent.

logprep/processor/generic_adder/rule.py
mhoff: I understand. Please add a comment on the `eq=False` to explain this and _maybe_ also add `eq=True` for `add_from_url` and explain why this is `True`

logprep/abc/getter.py
```diff
@@ -143,18 +145,26 @@ def _parse_newline_separated_list(content: str) -> list:
...
             )
 
+        return content
+
+    def get_list(self, content_field: str | None = None) -> list:
+        """Gets list and fails otherwise"""
+
+        content = self._resolve_content_by_content_type()
+
+        content = Getter._apply_content_field(content, content_field)
```
mhoff, line 169: Can you add this to `get_dict` as well? Not related to your PR, but it would make sense to add support to all three high-level interface methods in the same way.

Also, my personal preference would be:

```suggestion
    def get_list(self, content_field: str | None = None) -> list:
        """Gets list and fails otherwise"""

        content = self._resolve_content_by_content_type()
        content = Getter._apply_content_field(content, content_field)
```

(so without the line break)

logprep/ng/processor/generic_adder/processor.py
```diff
@@ -25,10 +25,13 @@
 """
 
 import typing
+from typing import Sequence
```
mhoff:
```suggestion
from collections.abc import Sequence
```

logprep/processor/generic_adder/processor.py
```diff
@@ -25,10 +25,12 @@
 """
 
 import typing
+from typing import Sequence
```
mhoff:
```suggestion
from collections.abc import Sequence
```

logprep/ng/processor/generic_adder/processor.py
```diff
@@ -37,8 +40,26 @@ class GenericAdder(Processor):
 
     rule_class = GenericAdderRule
 
+    @property
+    def _rules(self) -> Sequence[GenericAdderRule]:
+        """Returns all rules"""
+        return typing.cast(Sequence[GenericAdderRule], self.rules)
```
mhoff: Why not "rules"?

```suggestion
    @property
    def rules(self) -> Sequence[GenericAdderRule]:
        """Returns all rules"""
        return typing.cast(Sequence[GenericAdderRule], super().rules)
```

like in

```
    @property
    def rules(self) -> Sequence[PseudonymizerRule]:
        """Returns all rules"""
        return typing.cast(Sequence[PseudonymizerRule], super().rules)
```

and other sites

logprep/processor/generic_adder/rule.py
```diff
@@ -83,13 +83,59 @@
...
+        default=None, validator=validators.optional(validators.instance_of(str))
+    )
+
+    target_field_mapping: dict[str, str] = field(
+        validator=validators.deep_mapping(
+            key_validator=validators.instance_of(str),
+            value_validator=validators.instance_of(str),
+        ),
+        factory=dict,
+    )
```
mhoff: Missing pydoc for all three

logprep/abc/getter.py
Pablu23: Sure ill fix the line breaks, but adding this to get_dict somehow feels wrong to me, because we just have to do this here and for collection because we want to return a list, but in the dict, we can just get the content_field value like you would normally in a dict. Whats your opinion on this?

logprep/processor/generic_adder/rule.py
```diff
@@ -83,13 +83,59 @@
...
+
+    def __attrs_post_init__(self) -> None:
+        if not self.target_field and not self.target_field_mapping:
+            raise ValueError("adding values from url requires target_field or target_field_mapping")
```
mhoff: This is does not have a super class, but we need to ensure all derived classes actually call super __attrs_post_init__. I wonder if we can just always have it (even for non-derived classes), but I have not tried it and out this might throw an error. Try it out if you feel like it, otherwise resolve as a no-op.

```suggestion
    def __attrs_post_init__(self) -> None:
        super().__attrs_post_init__()
        if not self.target_field and not self.target_field_mapping:
            raise ValueError("adding values from url requires target_field or target_field_mapping")
```

logprep/processor/generic_adder/rule.py
```diff
@@ -141,6 +187,13 @@ class Config(FieldManagerRule.Config):
            authenticity and integrity of the loaded values.
 
         """
+
+        add_from_url: AddFromUrlConfig | None = field(
+            default=None,
+            validator=validators.optional(validators.instance_of(AddFromUrlConfig)),
+            converter=_convert_add_from_url,
+        )
```
mhoff, line 218: pydoc missing

logprep/processor/generic_adder/rule.py
```diff
@@ -160,20 +213,47 @@ def _refresh_add(self):
         def __attrs_post_init__(self):
```
mhoff, line 270: super needs to be called here

logprep/processor/generic_adder/rule.py
```diff
@@ -160,20 +213,47 @@ def _refresh_add(self):
...
+                raise ValueError(
+                    "one of target_field or target_field_mapping is neccessary per GenericAdder rule"
+                )
+
+            # Eagerly loaded from file
+            for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
+                getter = GetterFactory.from_string(add_file)
+                if isinstance(getter, RefreshableGetter):
+                    # TODO: This never gets cleaned up, Memory leak on a lot of new generic adders / generic resolvers
+                    getter.add_callback(f"generic_adder:{self.id}:{add_file}", self._refresh_add)
```
mhoff: We have the job_tag now and can resolve this properly

logprep/processor/generic_adder/rule.py
```diff
@@ -160,20 +213,47 @@ def _refresh_add(self):
...
+                self.add_from_url is not None
+                and not self.add_from_url.target_field
+                and not self.add_from_url.target_field_mapping
+            ):
+                raise ValueError(
+                    "one of target_field or target_field_mapping is neccessary per GenericAdder rule"
+                )
+
+            # Eagerly loaded from file
+            for add_file in self.add_from_file:  # pylint: disable=not-an-iterable
```
mhoff:
```suggestion
            for add_file in self.add_from_file:
```

It's an iterable, right?

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +275,157 @@ def _add_from_path(self):
...
+            return
+
+        assert config.add_from_url is not None
+
+        base_template = DottedTemplate(config.add_from_url.url)
+        resolved_template = DottedTemplate(base_template.safe_substitute({**ENV_VARS}))
+        self._dynamic_template = resolved_template
+        self._dynamic_identifiers = tuple(resolved_template.get_identifiers())
+
+        if len(self._dynamic_identifiers) > 0:
```
mhoff:
```suggestion
        if self._dynamic_identifiers:
```

logprep/processor/generic_adder/rule.py
Pablu23: Not here we dont, because that is the rule config and gets loaded on config time, we dont pass in the job tag, but I guess I could just change that with a private _callback_tag on the config, pass that into the rule from the processor and then also remove the init because now the rule knows the callback tag?

logprep/processor/generic_adder/rule.py
mhoff: You could even go all fancy with

```suggestion
        self._is_dynamic = bool(self._dynamic_identifiers)
```

but I think the if makes sense.

However, you can simply merge this if with the check afterwards

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +275,157 @@ def _add_from_path(self):
...
+            http_getter.add_callback(
+                self._callback_tag,
+                self._update_static_content,
+                deduplication_key=(self._callback_tag, static_uri, id(self)),
+                fnc_args=[http_getter, static_uri],
+            )
+
+            self._static_uri = static_uri
+
+    def _dynamic_add_from_url(self, event: dict) -> dict[str, FieldValue]:
```
mhoff: Typing missing on dict

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +275,157 @@ def _add_from_path(self):
...
+        config = typing.cast(GenericAdderRule.Config, self._config)
```
mhoff: Use the standard casted config property for this

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +275,157 @@ def _add_from_path(self):
...
+            self.mark_failed(error)
+        else:
+            self.clear_failed()
+
+    def _cleanup(self, resolved_uri: str):
+        self._dynamic_content.pop(resolved_uri, None)
+
+    def add(self, event: dict) -> dict:
         """Returns the fields to add"""
         config = typing.cast(GenericAdderRule.Config, self._config)
```
mhoff: Same here, casted property

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +275,157 @@ def _add_from_path(self):
...
-        return config.add
+
+        if config.add_from_file or config.add:
+            return config.add
```
mhoff: Is this the logic? 
- `if config.add_from_file` => `config.add` should already be populated through __attrs_post_init__
- `if config.add` => `config.add` already preloaded through __attrs_post_init__ or static http

Shouldn't the check then be simply `if config.add`?

mhoff: Nevermind, I confused `config.add` with `self.add`. But you want to return `self.add` here, right?

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +275,157 @@ def _add_from_path(self):
...
+
+        if not self._is_dynamic:
+            assert self._static_uri
+            return self._content_to_items_to_add(self._dynamic_content[self._static_uri])
```
mhoff: This reads weird.

`if not dynamic: ... get dynamic content ...`

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +275,157 @@ def _add_from_path(self):
...
+        else:
+            RefreshableGetter.keep_alive_for_target(dynamic_resolved)
+            content = self._dynamic_content[dynamic_resolved]
+
+        return self._content_to_items_to_add(content)
+
+    def _content_to_items_to_add(self, content: FieldValue):
+        items_to_add: dict[str, FieldValue] = {}
+
+        config = typing.cast(GenericAdderRule.Config, self._config)
```
mhoff: property

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +275,157 @@ def _add_from_path(self):
...
+        assert config.add_from_url
+
+        if config.add_from_url.target_field:
+            items_to_add[config.add_from_url.target_field] = content
+        else:
+            assert config.add_from_url.target_field_mapping is not None
+
+            if not isinstance(content, dict):
+                raise ValueError("add_from_url.target_field_mapping requires a mapping response")
```
mhoff:
```suggestion
                raise ValueError(f"add_from_url.target_field_mapping requires a mapping response, got: {content}")
```

logprep/processor/generic_adder/rule.py
Pablu23: You are correct, I initially didnt understand that add_from_file populates add automatically

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +275,157 @@ def _add_from_path(self):
...
+
+            for (
+                mapping_source_field,
+                mapping_target_field,
+            ) in config.add_from_url.target_field_mapping.items():
+                item = get_dotted_field_value_with_missing(content, mapping_source_field)
+                if item is MISSING:
+                    logger.warning(
+                        "could not add source_field: %s for target_field: %s because was missing",
```
mhoff:
```suggestion
                        "could not add source_field %s for target_field %s because it is missing",
```

logprep/processor/generic_adder/rule.py
Pablu23: Yes you are right, I might have to change the name of _dynamic_content, as that is just a cache

logprep/processor/generic_adder/rule.py
mhoff: I think it reads weird with the colons. Rather "source_field=%s" or "source_field %s" or "source_field '%s'" I'd say

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +275,157 @@ def _add_from_path(self):
...
+                        mapping_source_field,
+                        mapping_target_field,
+                    )
+                    continue
```
mhoff: If we stick to this, this might be a case for a processor warning

logprep/abc/getter.py
mhoff: Theroretically you could always get the field yourself, as supplying a `content_field` enforces the top-level structure to be a `dict` and thus the calling code could do it itself. That the top-level structure and the content are both a `dict` does not really change anything about this. It is rather a coincidence than enything else.

Does this make sense? Maybe I am misunderstanding what you were trying to say/ask.

logprep/processor/generic_adder/rule.py
mhoff: Why not simply move this code to the init and use one job_tag (the one from outside) for every callback?

logprep/processor/generic_adder/rule.py
Pablu23: Do you suggest doing this in this PR?

logprep/processor/generic_adder/processor.py
Pablu23: You are absolutely right in that, but I would rather not change the list_comparison in this PR again. For this we should keep it in mind and do a seperate PR like with the test cases

logprep/processor/generic_adder/rule.py
Pablu23: Does not work like this by itself, ill make this a noop and resolve

logprep/processor/generic_adder/rule.py
mhoff: I think we talked about this in person. Feel free to resolve

Pablu23: I removed the target_mapping like we talked about. If that sufficient for you please resolve this thread

logprep/processor/generic_adder/processor.py
mhoff: Fair enough

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +245,139 @@ def _add_from_path(self):
...
+                        self.config._refresh_add,
+                        deduplication_key=(self._callback_tag, add_file, id(self)),
+                    )
+                self.config._add_from_path()
+            return
+
+        assert self.config.add_from_url is not None
+
+        base_template = DottedTemplate(self.config.add_from_url.url)
+        resolved_template = DottedTemplate(base_template.safe_substitute({**ENV_VARS}))
```
mhoff:
```suggestion
        resolved_template = DottedTemplate(base_template.safe_substitute(ENV_VARS))
```

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +245,139 @@ def _add_from_path(self):
...
+        self._dynamic_template = resolved_template
+        self._dynamic_identifiers = tuple(resolved_template.get_identifiers())
+
+        if not self._dynamic_identifiers:
+            static_uri = resolved_template.substitute()
+            http_getter = GetterFactory.from_string(static_uri)
+
+            assert isinstance(http_getter, HttpGetter)
```
mhoff:
```suggestion
            assert isinstance(http_getter, RefreshableGetter)
```

I would like to get rid of the `isinstance` checks generally, but we lack the design for that at the moment. Having said that, _if_ we check for a class type it should be `RefreshableGetter` and not one of the concrete implementations. For most cases I don't see a reason to differentiate between `FileGetter` and `HttpGetter`.

Hence I would also argue that we should remove this differentiation from the variable naming as well:

```suggestion
            assert isinstance(getter, RefreshableGetter)
```

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +245,139 @@ def _add_from_path(self):
                     f"The following required files do not exist: '{missing_files}'"
                 )
 
+    def __init__(self, filter_rule: FilterExpression, config: Config, processor_name: str):
+        super().__init__(filter_rule, config, processor_name)
+        self._chached_add_content: dict[str, FieldValue] = {}
```
mhoff:
```suggestion
        self._cached_add_content: dict[str, FieldValue] = {}
```

logprep/processor/generic_adder/rule.py
```diff
@@ -195,8 +245,139 @@ def _add_from_path(self):
...
+
+        return {self.config.add_from_url.target_field: content}
+
+    def _update_dynamic_content(self, http_getter: HttpGetter, resolved_uri: str):
+        content = http_getter.get_collection()
+        self._chached_add_content[resolved_uri] = content
+
+    def _update_static_content(self, getter: HttpGetter, uri: str) -> None:
+        try:
+            self._update_dynamic_content(getter, uri)
```
mhoff: These three lines are surprising to me. `_update_static_content` calls `_update_dynamic_content` but does nothing more than error handling. 

It seems to me those methods are not properly named. 
Might `_update_dynamic_content` rather be `_get_and_cache_uri`?

Also be careful with the error handling. Right now you are registering a callback which does not handle errors. So the errors escalate towards the refreshable getter path, which is unexpected.

Pablu23: This should be fixed now I believe, but im not 100% sure so please do look over this again

tests/unit/ng/processor/generic_adder/test_generic_adder.py
```diff
@@ -95,10 +101,123 @@ async def test_add_only_copies(self):
...
+        }
+        assert responses.calls[0].request.url == resolved_url
+
+    @responses.activate
+    async def test_dynamic_url_failure_is_event_scoped(self):
+        failed_url = "https://values.example/acme"
+        successful_url = "https://values.example/beta"
+        responses.add(responses.GET, failed_url, status=500)
+        responses.add(responses.GET, successful_url, json={"risk": {"score": 7}})
+        RefreshableGetter.reset()
```
mhoff: As discussed

logprep/processor/generic_adder/rule.py
mhoff: Do we still need this?

With the removal of `# pylint: enable=anomalous-backslash-in-string` this now spans the whole file

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +80,104 @@
...
+from logprep.util.helper import (
+    DottedTemplate,
+    FieldValue,
+    get_dotted_field_value,
+)
+
+
+@define(kw_only=True, frozen=True)
+class UriTargetConfig:
+    """Configuration for adding values loaded from an HTTP(S) URL."""
```
mhoff:
```suggestion
    """Configuration for adding values loaded from an URIs, which can be HTTP(S) URLs or file paths"""
```

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +80,104 @@
...
+
+    uri: str = field(validator=validators.and_(validators.instance_of(str), validators.min_len(1)))
+    """The URL to load values from.
```
mhoff:
```suggestion
    """The URI to load values from.
```

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +80,104 @@
...
+
+    Environment variables and dotted event fields can be inserted with placeholders such as
+    :code:`${TENANT}` and :code:`${tenant.id}`.
```
mhoff:
```suggestion
    :code:`${LOGPREP_DATA_API}` and :code:`${tenant.id}`.
```

Rather use something LOGPREP_ prefixed in the env example here, as we will most likely reintroduce this restriction at some point in the future

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +80,104 @@
...
+    )
+    """The dotted event field into which the complete response is written."""
+
+
+UriSpec = str | UriTargetConfig
+
+
+def _convert_uri_specs(
+    value: UriSpec | dict | Sequence[UriSpec | dict],
+) -> tuple[UriSpec, ...]:
```
mhoff:
```suggestion
) -> Sequence[UriSpec]:
```

I would recommend to use `Sequence` as return type as it's basically the same as tuple but not exposing the concrete data structure. And the "..." makes the tuple type harder to read imo

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +80,104 @@
...
+    if isinstance(value, (str, UriTargetConfig, dict)):
+        values: tuple[UriSpec | dict, ...] = (value,)
```
mhoff: I would also recommend to use `Sequence` as a type here just to get around the "...". Using a `Sequence` further diminishes the else path, as it is already a `Sequence` then

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +80,104 @@
...
+    else:
+        values = tuple(value)
 
+    return tuple(
+        convert_from_dict(UriTargetConfig, item) if isinstance(item, dict) else item
```
mhoff:
```suggestion
        convert_from_dict(UriTargetConfig, item)
```

`convert_from_dict` already does this

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +80,104 @@
...
+                iterable_validator=validators.instance_of(list),
+                member_validator=validators.instance_of(str),
+            ),
             converter=lambda x: x if isinstance(x, list) else [x],
             factory=list,
+            # Eq false in this case means that this is not taken into account when comparing two Configs,
+            # that is neccessary in this case because on init this will be loaded into add,
+            # and when comparing two rules, we dont care if whats to add came from a file or is written inline
+            # but we do care if whatever gets added is the same
             eq=False,
```
mhoff: This needs to be revisited as we now refrain from eager loading, right?

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
...
+
+        add_from_uri: tuple[UriSpec, ...] = field(
+            factory=tuple,
+            converter=_convert_uri_specs,
+            validator=validators.deep_iterable(
+                iterable_validator=validators.instance_of(tuple),
+                member_validator=validators.instance_of((str, UriTargetConfig)),
+            ),
+            # Explicitly set it to true here to show the difference between this and add_from_file
+            eq=True,
```
mhoff: Same here

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
            authenticity and integrity of the loaded values.
 
         """
+
+        add_from_uri: tuple[UriSpec, ...] = field(
+            factory=tuple,
```
mhoff:
```suggestion
        add_from_uri: Sequence[UriSpec] = field(
            factory=list,
```

I get that using a read-only type is useful to strictly forbid writing to a once created config. The only thing I dislike about this is that the tuple as a type is not nice to use and exposes and implementation detail. The type should only be `Sequence`. Maybe we can keep the factory to tuple

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
...
+            converter=_convert_uri_specs,
+            validator=validators.deep_iterable(
+                iterable_validator=validators.instance_of(tuple),
+                member_validator=validators.instance_of((str, UriTargetConfig)),
+            ),
+            # Explicitly set it to true here to show the difference between this and add_from_file
+            eq=True,
+        )
+        """Configuration for loading values from an HTTP(S) URL and adding them to the event.
```
mhoff: URIs

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
...
+            if self.add_from_file:
+                self.add_from_uri = tuple(self.add_from_file)
 
-        def _refresh_add(self):
-            self.add = copy.deepcopy(self._base_add)
-            self._add_from_path()
+            if self.only_first_existing_file and not self.add_from_file:
+                raise ValueError(
+                    "only_first_existing_file is only supported with deprecated add_from_file"
+                )
```
mhoff, line 282: Good. But let's talk about this on Monday. Hopefully we don't this in the future

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
...
-        """Stores original add fields (as provided in the config) for future refreshes of getters"""
+        def __attrs_post_init__(self):
+            if self.add_from_file and self.add_from_uri:
+                raise ValueError(
+                    "Deprecated add_from_file and new add_from_uri cannot both be configured"
+                )
 
-        # pylint: enable=anomalous-backslash-in-string
+            if self.add_from_file:
+                self.add_from_uri = tuple(self.add_from_file)
```
mhoff: I am surprised this works. I would have thought post-init runs after validation

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
...
-                    raise InvalidRuleDefinitionError(error_msg)
-                if self.only_first_existing_file:
-                    break
-            if missing_files and not self.only_first_existing_file:
+    def __init__(self, filter_rule: FilterExpression, config: Config, processor_name: str):
+        super().__init__(filter_rule, config, processor_name)
+        self._callback_tag = ""
+        self._uri_sources: list[_UriSource] = []
+
+    def init_generic_adder(self, job_tag: str) -> None:
```
mhoff, line 307: pydoc for public method

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
...
+        self._callback_tag = job_tag
+        self._uri_sources.clear()
```
mhoff: I'd say let's rather fail on double init or simply assume it doesn't happen

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
            authenticity and integrity of the loaded values.
 
         """
+
+        add_from_uri: tuple[UriSpec, ...] = field(
```
mhoff: I just realized that we are not unifying the type to always be `UriTargetConfig` but `UriSpec` which is `str | UriTargetConfig`. One downside of this approach is that we have to differentiate using `isinstance` checks. The alternative would be to always have a `UriTargetConfig` which has `target_field` of type `str | None` which allows for `is None` checks and `uri` access is checkless. Did you consider this approach?

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
...
     @property
-    def add(self) -> dict:
+    def merge_with_target(self) -> bool:
+        """Returns the nested config merge_with_target"""
+        return self.config.merge_with_target
+
+    @property
+    def config(self) -> Config:
+        """Return typed config"""
+        return typing.cast(GenericAdderRule.Config, self._config)
```
mhoff: Rather group those at the top of the class, jsut before init

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
...
+
+        resolved_uri = source.template.substitute(values)
+
+        if resolved_uri in source.content_by_uri:
+            RefreshableGetter.keep_alive_for_target(resolved_uri)
+            return source.content_by_uri[resolved_uri]
+
+        getter = GetterFactory.from_string(resolved_uri)
+        if not isinstance(getter, RefreshableGetter):
+            return getter.get_collection()
```
mhoff: Actually, we should at least produce a warning here or maybe even fail here. This is non-cached I/O in a hot path

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
...
+        getter.add_cleanup_callback(
+            self._callback_tag,
+            self._cleanup,
+            deduplication_key=key,
+            fnc_args=[source, resolved_uri],
+        )
+
+        return content
+
+    def _content_for_source(self, source: _UriSource, event: dict[str, FieldValue]) -> FieldValue:
```
mhoff, line 427:
```suggestion
    def _get_cached_or_dynamic_content(self, source: _UriSource, event: dict[str, FieldValue]) -> FieldValue:
```

Maybe something like this?

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,240 @@ class Config(FieldManagerRule.Config):
...
+
+        if isinstance(content, dict):
+            return dict(content)
+
+        raise ValueError(
+            f"URI source {uri!r} without target_field must contain a mapping, got {type(content).__name__}"
+        )
+
+    def _fetch_and_cache_uri(self, source: _UriSource, getter: Getter, resolved_uri: str):
+        content = getter.get_collection()
```
mhoff: content_field?

logprep/processor/generic_adder/rule.py
Pablu23: This doesnt work correctly, what would you suggestion be? Also the if check needs to stay as doing tuple(str) returns not [str] but ["s", "t", "r"]

mhoff: `[value]`

logprep/processor/generic_adder/rule.py
mhoff: Probably Error, but feel free to make up your mind

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +79,104 @@
...
+    FieldValue,
+    get_dotted_field_value,
+)
+
+
+@define(kw_only=True, frozen=True)
+class UriConfig:
+    """Configuration for adding values loaded from an URI, which can be HTTP(S) URLs or file paths"""
+
+    uri: str = field(validator=validators.and_(validators.instance_of(str), validators.min_len(1)))
```
mhoff, line 113: AFAIK this is the same:

```suggestion
    uri: str = field(validator=[validators.instance_of(str), validators.min_len(1)])
```

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +79,104 @@
...
+    """The URI to load values from.
+
+    Environment variables and dotted event fields can be inserted with placeholders such as
+    :code:`${LOGPREP_DATA_API}` and :code:`${tenant.id}`.
+    """
+
+    target_field: str | None = field(
+        default=None,
+        validator=validators.optional(
+            validators.and_(validators.instance_of(str), validators.min_len(1))
```
mhoff, line 123: Maybe here `[]` work as well, but I don't know

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +79,104 @@
...
+        UriConfig(uri=item) if isinstance(item, str) else convert_from_dict(UriConfig, item)
+        for item in values
+    ]
+
+
+@define(kw_only=True)
+class _UriSource:
+    config: UriConfig
+    template: DottedTemplate
+    identifiers: tuple[str, ...]
```
mhoff:
```suggestion
    identifiers: Sequence[str]
```

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +198,285 @@ class Config(FieldManagerRule.Config):
...
 
-    @property
-    def add(self) -> dict:
+            self._uri_sources.append(source)
+
+            try:
+                self._init_static_source(source)
+                if source.error is not None:
+                    raise InvalidRuleDefinitionError(
+                        f"Could not load generic-adder URI {config.uri!r}: {source.error}"
```
mhoff:
```suggestion
                        f"Could not load generic_adder URI {config.uri!r}: {source.error}"
```

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +198,285 @@ class Config(FieldManagerRule.Config):
...
+                deduplication_key=(self._callback_tag, resolved_uri, id(source)),
+                fnc_args=[source, getter, resolved_uri],
+            )
+            return
+
+        try:
+            self._fetch_and_cache_uri(source, getter, resolved_uri)
+        except (FileNotFoundError, ValueError) as error:
+            raise InvalidRuleDefinitionError(
+                f"Could not load generic-adder URI {resolved_uri!r}: {error}"
```
mhoff:
```suggestion
                f"Could not load generic_adder URI {resolved_uri!r}: {error}"
```

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +198,285 @@ class Config(FieldManagerRule.Config):
...
+        return source.content_by_uri[resolved_uri]
+
+    def _content_to_items_to_add(
+        self, config: UriConfig, content: FieldValue
+    ) -> dict[str, FieldValue]:
+        if config.target_field is not None:
+            return {config.target_field: content}
+
+        if isinstance(content, dict):
+            return dict(content)
```
mhoff: Do we need to return a copy here? Otherwise `return content` should be fine

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +198,285 @@ class Config(FieldManagerRule.Config):
...
+
+        raise ValueError(
+            f"URI source {config.uri!r} without target_field must contain a mapping, got {type(content).__name__}"
+        )
+
+    def _fetch_and_cache_uri(self, source: _UriSource, getter: Getter, resolved_uri: str):
+        content = getter.get_collection(content_field=self.config.content_field)
+
+        self._content_to_items_to_add(source.config, content)
```
mhoff: Why do we do this? Validation?

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +198,285 @@ class Config(FieldManagerRule.Config):
...
+            fnc_args=[source, resolved_uri],
+        )
+
+        return content
+
+    def _content_for_source(self, source: _UriSource, event: dict[str, FieldValue]) -> FieldValue:
+        if source.identifiers:
+            return self._get_cached_or_dynamic_content(source, event)
+
+        resolved_uri = source.template.substitute()
```
mhoff: Template substitution in the hot path... we should avoid this

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +198,285 @@ class Config(FieldManagerRule.Config):
...
         """Returns the fields to add"""
-        config = typing.cast(GenericAdderRule.Config, self._config)
-        return config.add
+        items_to_add: dict[str, FieldValue] = dict(self.config.add)
+
+        for source in self._uri_sources:
+            content = self._content_for_source(source, event)
+            items_to_add.update(self._content_to_items_to_add(source.config, content))
+
+        return items_to_add
```
mhoff: Now we are always creating a fresh dict, even if we only return static entries given by `config.add`.

I see multiple optimization paths:
1) Have an early return with `return self.config.add` if `not self._uri_sources`. This would make the most basic case straight-forward.
2) Have a dedicated precomputed attribute on the rule `_static_add` which contains `config.add` and the fetched entries of all static lists. Each static list update callback recomputes `_static_add` if their contents changed. This would make the most basic case *and* the static case more straight-forward.
3) Return a Generator of dicts instead of a single unified dict. This way we can properly handle key collisions (currently two sources declaring the same key would simply overwrite each other instead of merging; the old logic did the same) and we can circumvent unnecessary object creation & dict updates which will be thrown away either way.

I normally would have recommended to go with at least (1) and also do (2) in this or another PR. But actually the generator solutions seems to solve all problems at once, it is just heavier on the tests and would need a wrapper call to unpack & wrap in a dict like before.

Happy to discuss

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +198,285 @@ class Config(FieldManagerRule.Config):
...
+                fnc_args=[source, getter, resolved_uri],
+            )
+            return
+
+        try:
+            self._fetch_and_cache_uri(source, getter, resolved_uri)
+        except (FileNotFoundError, ValueError) as error:
+            raise InvalidRuleDefinitionError(
+                f"Could not load generic-adder URI {resolved_uri!r}: {error}"
+            ) from error
```
mhoff: Why not consolidate paths here and simply control error handling through a flag? This would also make `_init_first_existing_file` easier. 

```suggestion
    def _init_static_source(self, source: _UriSource, raise_on_error: bool):
        resolved_uri = source.template.substitute()
        getter = GetterFactory.from_string(resolved_uri)

        assert self._callback_tag

        self._update_static_content(source, getter, resolved_uri)
        if isinstance(getter, RefreshableGetter):
           getter.add_callback(
                self._callback_tag,
                self._update_static_content,
                deduplication_key=(self._callback_tag, resolved_uri, id(source)),
                fnc_args=[source, getter, resolved_uri],
            )

        if source.error and raise_on_error:
            raise InvalidRuleDefinitionError(
                f"Could not load generic-adder URI {resolved_uri!r}: {error}"
            ) from source.error
```

logprep/processor/generic_adder/rule.py
Pablu23: For me this and the following comment result in mypy / lsp errors, attrs documentation says this should work, but the typing is unhappy. I would rather leave it like this and not have the errors here but im also open to changing it

logprep/processor/generic_adder/rule.py
Pablu23: Sorry I forgot / am unsure, this is a noop now right?

logprep/processor/generic_adder/rule.py
Pablu23: Is this implemented in the way you thought of now? Please look over it and let me know

logprep/processor/generic_adder/rule.py
mhoff: I see. I tried it quickly with some tuple and got a mypy issue when using `and_`. Could not infer argument type 1. What is your exact error message?

logprep/processor/generic_adder/rule.py
mhoff: We keep it as it is for now. If @ppcad confirms that it's not required anymore we can remove it, but this can also happen later and outside of this PR. Keep this conversation open until we are ready to merge -- if there is no clear decision until then, we keep it and resolve this conversation.

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,290 @@ class Config(FieldManagerRule.Config):
...
+
+    def _recompute_failure_state(self) -> None:
+        errors = [source.error for source in self._uri_sources if source.error is not None]
+
+        if not errors:
+            self.clear_failed()
+        elif len(errors) == 1:
+            self.mark_failed(errors[0])
+        else:
+            self.mark_failed(ExceptionGroup("generic-adder URI loading failed", errors))
```
mhoff:
```suggestion
            self.mark_failed(ExceptionGroup("generic_adder URI loading failed", errors))
```

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,290 @@ class Config(FieldManagerRule.Config):
...
+        if isinstance(content, dict):
+            return content
+
+        raise ValueError(
+            f"URI source {config.uri!r} without target_field must contain a mapping, got {type(content).__name__}"
+        )
+
+    def _fetch_and_cache_uri(self, source: _UriSource, getter: Getter, resolved_uri: str):
+        content = getter.get_collection(content_field=self.config.content_field)
+
```
mhoff:
```suggestion
```

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,290 @@ class Config(FieldManagerRule.Config):
...
-    @property
-    def add(self) -> dict:
+            self._uri_sources.append(source)
+
+            try:
+                self._init_static_source(source, raise_on_error=True)
+                if source.error is not None:
+                    raise InvalidRuleDefinitionError(
+                        f"Could not load generic_adder URI {config.uri!r}: {source.error}"
+                    ) from source.error
```
mhoff:
```suggestion
```

Can now be removed, right?

---

Or you go the other road and set `raise_on_error=False` like:

```
if source.error is not None:
    error = source.error
    if isinstance(error, InvalidRuleDefinitionError) and isinstance(error.__cause__, FileNotFoundError):
        missing_files.append(config.uri)
        continue
    # ...
```

logprep/processor/generic_adder/rule.py
Pablu23: ```logprep/processor/generic_adder/rule.py:104: error: Argument "validator" to "field" has incompatible type "list[function]"; expected "Callable[[Any, Attribute[Never], Never], Any] | Sequence[Callable[[Any, Attribute[Never], Never], Any]] | None"  [arg-type]```

This one

logprep/ng/processor/generic_adder/processor.py
```diff
@@ -37,8 +40,28 @@ class GenericAdder(Processor):
...
+    def rules(self) -> Sequence[GenericAdderRule]:
+        """Returns all rules"""
+        return typing.cast(Sequence[GenericAdderRule], super().rules)
+
+    async def setup(self):
+        await super().setup()
+        for rule in self.rules:
+            rule.init_generic_adder(self._job_tag_for_cleanup)
+
     def _apply_rules(self, event: dict, rule: Rule) -> None:
```
mhoff:
```suggestion
    def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
```

logprep/processor/generic_adder/processor.py
```diff
@@ -37,8 +39,26 @@ class GenericAdder(Processor):
...
+    def rules(self) -> Sequence[GenericAdderRule]:
+        """Returns all rules"""
+        return typing.cast(Sequence[GenericAdderRule], super().rules)
+
+    def setup(self):
+        super().setup()
+        for rule in self.rules:
+            rule.init_generic_adder(self._job_tag_for_cleanup)
+
+    def _apply_rules(self, event: dict, rule: GenericAdderRule):
```
mhoff:
```suggestion
    def _apply_rules(self, event: dict[str, FieldValue], rule: GenericAdderRule):
```

logprep/processor/generic_adder/rule.py
mhoff: But this line actually works fine with my IDE. Resolved

logprep/processor/generic_adder/rule.py
```diff
@@ -80,46 +79,105 @@
...
+from logprep.util.helper import (
+    DottedTemplate,
+    FieldValue,
+    get_dotted_field_value,
+)
+
+
+@define(kw_only=True, frozen=True)
+class UriConfig:
+    """Configuration for adding values loaded from an URI, which can be HTTP(S) URLs or file paths"""
```
mhoff:
```suggestion
    """
    Configuration for adding values loaded from an URI, which can be HTTP(S) URLs or file paths
    """
```

(line too long)

logprep/processor/generic_adder/rule.py
Pablu23: That is weird.. which version of mypy are you using? For me it can be either then I guess

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,285 @@ class Config(FieldManagerRule.Config):
...
+            source = self._create_uri_source(config)
+            if source.identifiers:
                 raise InvalidRuleDefinitionError(
-                    f"The following required files do not exist: '{missing_files}'"
+                    f"only_first_existing_file does not support event-dependent paths: {config!r}"
                 )
 
-    @property
-    def add(self) -> dict:
+            self._uri_sources.append(source)
```
mhoff: What do you think about doing the append only after the init was successful? I know that `_uri_sources` is used in computing failure state, but this should not matter for this case and might simplify the code

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,285 @@ class Config(FieldManagerRule.Config):
...
+
+            try:
+                self._init_static_source(source, raise_on_error=True)
+            except InvalidRuleDefinitionError as error:
+                self._uri_sources.pop()
+
+                if isinstance(error.__cause__, FileNotFoundError):
+                    missing_files.append(config.uri)
+                    continue
```
mhoff, line 337:
```suggestion
                    continue
                raise
            return
        raise InvalidRuleDefinitionError(f"None of the configured files exist: {missing_files!r}")
```

My preference would be less lines, but if you like it better we can keep it as it is - your choice.

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,285 @@ class Config(FieldManagerRule.Config):
...
+        return _UriSource(
+            config=config,
+            template=template,
+            identifiers=tuple(identifiers),
+            static_uri=None if identifiers else template.substitute(),
+        )
+
+    def _init_static_source(self, source: _UriSource, raise_on_error: bool):
+        resolved_uri = source.template.substitute()
+        getter = GetterFactory.from_string(resolved_uri)
```
mhoff:
```suggestion
        assert source.static_url is not None
        getter = GetterFactory.from_string(source.static_url)
```

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,285 @@ class Config(FieldManagerRule.Config):
...
+
+    def _update_static_content(self, source: _UriSource, getter: Getter, uri: str) -> None:
+        try:
+            self._fetch_and_cache_uri(source, getter, uri)
+        except Exception as error:
+            source.error = error
+        else:
+            source.error = None
+
+        self._recompute_failure_state()
```
mhoff, line 459: We are recomputing the list and aggregate error always. But should not be too bad as we shouldn't fetch too often either way... Only meant as a hint and feel free to resolve this comment

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +199,285 @@ class Config(FieldManagerRule.Config):
...
+
+    def _fetch_and_cache_uri(self, source: _UriSource, getter: Getter, resolved_uri: str):
+        content = getter.get_collection(content_field=self.config.content_field)
+        source.content_by_uri[resolved_uri] = content
+        return content
+
+    def _update_static_content(self, source: _UriSource, getter: Getter, uri: str) -> None:
+        try:
+            self._fetch_and_cache_uri(source, getter, uri)
+        except Exception as error:
```
mhoff: My pylint complains that we are catching a too general exception here. Feel free to disable it for this line:

```suggestion
        except Exception as error: # pylint: disable=broad-except
```

Not sure why it does not complain for the list_comparison. Maybe because we re-raise conditionally.

logprep/processor/generic_adder/rule.py
Pablu23: Haha same thing I commented on, on your PR, it should be relativly close to that one, would close this as we know about it, except if my implementation deviates from yours, and is through that less performant

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +202,279 @@ class Config(FieldManagerRule.Config):
...
+        self, config: UriConfig, content: FieldValue
+    ) -> dict[str, FieldValue]:
+        if config.target_field is not None:
+            return {config.target_field: content}
+
+        if isinstance(content, dict):
+            return content
+
+        raise ValueError(
+            f"URI source {config.uri!r} without target_field must contain a mapping, got {type(content).__name__}"
```
mhoff: Line too long, 100 is max

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +202,279 @@ class Config(FieldManagerRule.Config):
...
+            self.clear_failed()
+        elif len(errors) == 1:
+            self.mark_failed(errors[0])
+        else:
+            self.mark_failed(ExceptionGroup("generic_adder URI loading failed", errors))
+
+    def _cleanup(self, source: _UriSource, resolved_uri: str) -> None:
+        source.content_by_uri.pop(resolved_uri, None)
+
+    def additions(self, event: dict) -> Iterator[dict[str, FieldValue]]:
```
mhoff:
```suggestion
    def additions(self, event: dict[str, FieldValue]) -> Iterator[dict[str, FieldValue]]:
```

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +202,279 @@ class Config(FieldManagerRule.Config):
...
+        if self.config.add:
+            yield self.config.add
+
+        for source in self._uri_sources:
+            content = self._content_for_source(source, event)
+            yield self._content_to_items_to_add(source.config, content)
+
+    def add(self, event: dict) -> dict:
```
mhoff:
```suggestion
    def add(self, event: dict[str, FieldValue]) -> dict[str, FieldValue]:
```

logprep/processor/generic_adder/rule.py
```diff
@@ -141,62 +202,279 @@ class Config(FieldManagerRule.Config):
...
+        for identifier, val in values.items():
+            if val is None:
+                raise ValueError(
+                    f"missing event field {identifier!r} for dynamic generic adder URI"
+                )
+            if not isinstance(val, (str, int)):
+                raise ValueError(
+                    f"value for generic adder field {identifier!r} is not a scalar value"
+                )
+            pass
```
mhoff:
```suggestion
```

logprep/processor/generic_adder/rule.py
mhoff: Fine by me, resolved

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 * add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
+* generic_adder: add support for templated http urls
```
```suggestion
* generic_adder: add support for templated http urls & content_field
```

With the `generic_adder: ` pattern we are slightly deviating from older changelogs. But I think its fine and we can adopt it

logprep/processor/generic_adder/rule.py
mhoff: 2.1.0

Pablu23: Me too, that we get different behavior here is very weird, then ill just go with the list one, as it does look better

tests/unit/processor/generic_adder/test_generic_adder.py
mhoff, line 160: - there is no example test case for `add_from_uri`,  that would be important for the docs
- we need a test case for demonstrating the changed behavior through the generator: if any 2 sources (add, add_from_file, add_from_uri) define the same field names, they will now be merged instead of replaced (if merge_with_target=True)

logprep/processor/generic_adder/rule.py
ppcad: @mhoff and @Pablu23 Please do not remove the functionality of `only_first_existing_file`, as it is still required.

mhoff: Alright, thanks!

tests/unit/processor/generic_adder/test_generic_adder.py
Pablu23: Implemented but im unsure if thats what you meant exactly, please look over this once. Resolve if your satisfied

tests/unit/processor/generic_adder/test_generic_adder.py
```diff
@@ -24,6 +24,103 @@
...
+                }
+            },
+        },
+        {"event_id": 123},
+        {
+            "event_id": 123,
+            "uri_content": {
+                "some_added_field": "some value",
+                "another_added_field": "another_value",
+                "dotted.added.field": "yet_another_value",
```
mhoff, line 43: I needed a moment to understand why this dotted field notation did not get resolved to individual keys. But that's because of the target_field being on top and the dotted field name is hidden inside the payload dict. Might be suprising to some, but actually correct this way. Just an FYI and nothing to do.

tests/unit/processor/generic_adder/test_generic_adder.py
mhoff: Satisfied :)

mhoff (approved):
The code looks good, many thanks.

The docs for `add_from_uri` are unfortunately a bit short. The thing is, that the docs for the nested class is not wired automatically and the type alias (`UriConfig`) not at all.Also the security best practice is missing. It's good that we added the examples for people to know how to use it. If we don't extend the docs now, we should do it in the near future.

Image

PR #1003, opened 2026-07-31
feat: processors async apis

kaya-david:
## Description

* made some abstract Processor methods and processor-specific methods asynchronous, especially the processing chain through __apply_rules(...)
* introduced has_async_io() with a default return value of False
* updated and fixed the tests

## Assignee

### Documentation

### Code Quality

* pytest

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--1003.org.readthedocs.build/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Upcoming Changes
 ### Breaking
 * restrict UNIX timestamp normalization to seconds, milliseconds, microseconds, and nanoseconds.
+* handle non-string `concatenator` source fields with `ProcessingWarning` instead of `ProcessingCriticalError`
 
 ### Features
 * add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 * add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
-* add support for asynchronous rule processing and I/O capability detection to `ng` processors
+* introduce API-level support for asynchronous rule processing and I/O capability detection in `ng` processors
 
 ### Improvements
 * docs: enable pydoc placeholders for facilitating component reuse through inheritance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features
 * add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
+* add support for asynchronous rule processing and I/O capability detection to `ng` processors
 
 ### Improvements
 * docs: enable pydoc placeholders for facilitating component reuse through inheritance
--- a/logprep/ng/processor/concatenator/processor.py
+++ b/logprep/ng/processor/concatenator/processor.py
@@ -62,9 +62,10 @@ class Concatenator(FieldManager):
         self._handle_missing_fields(event, rule, rule.source_fields, source_field_values)
 
         source_field_values = [field for field in source_field_values if field is not None]
-        source_field_values_str: list[str] = [
-            str(source_field_value) for source_field_value in source_field_values
-        ]
-        target_value = f"{rule.separator}".join(source_field_values_str)
+
+        try:
+            target_value = f"{rule.separator}".join(source_field_values)  # type: ignore[arg-type]
+        except:
+            raise
 
         self._write_target_field(event, rule, target_value)
--- a/logprep/ng/processor/concatenator/processor.py
+++ b/logprep/ng/processor/concatenator/processor.py
@@ -25,9 +25,16 @@ Processor Configuration
 .. automodule:: logprep.processor.concatenator.rule
 """
 
+import typing
+from collections.abc import Iterable
+
 from logprep.ng.processor.field_manager.processor import FieldManager
+from logprep.processor.base.rule import Rule
 from logprep.processor.concatenator.rule import ConcatenatorRule
-from logprep.util.helper import get_dotted_field_value
+from logprep.util.helper import (
+    FieldValue,
+    get_dotted_field_value,
+)
 
 
 class Concatenator(FieldManager):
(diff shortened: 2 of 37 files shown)


logprep/ng/processor/concatenator/processor.py
```diff
@@ -48,13 +55,22 @@ def _apply_rules(self, event: dict, rule: ConcatenatorRule) -> None:
...
+        source_field_values = [
+            get_dotted_field_value(event, source_field) for source_field in rule.source_fields
+        ]
+
+        self._handle_missing_fields(
+            event,
+            rule,
+            rule.source_fields,
+            source_field_values,
+        )
```
mhoff:
```suggestion
        source_field_values = get_dotted_field_values(event, source_fields)
        self._handle_missing_fields(event, rule, rule.source_fields, source_field_values)
```

The helper `get_dotted_field_values` makes it a bit shorter and I am not sure if stretching the lines here is helpful, as exactly this sequence of calls (`get_dotted_field_value`, `self._handle_missing_fields`) happens in all processors

logprep/ng/processor/ip_informer/processor.py
```diff
@@ -58,10 +61,16 @@ def _get_results(self, ip_address_list: Iterable, rule: IpInformerRule) -> dict:
...
-    def _get_flat_ip_address_list(self, event: dict, rule: IpInformerRule) -> Iterable:
-        source_field_values = list(map(partial(get_dotted_field_value, event), rule.source_fields))
-        list_elements = filter(lambda x: isinstance(x, list), source_field_values)
-        str_elements = filter(lambda x: isinstance(x, str), source_field_values)
+    def _get_flat_ip_address_list(
+        self, event: dict[str, FieldValue], rule: IpInformerRule
+    ) -> Iterable:
+        source_field_values = [
+            get_dotted_field_value(event, source_field) for source_field in rule.source_fields
+        ]
```
mhoff: get_dotted_field_values

logprep/ng/processor/concatenator/processor.py
```diff
@@ -48,13 +55,22 @@ def _apply_rules(self, event: dict, rule: ConcatenatorRule) -> None:
...
 
-        source_field_values = [field for field in source_field_values if field is not None]
-        target_value = f"{rule.separator}".join(source_field_values)
+        string_values = [
+            field_value for field_value in source_field_values if isinstance(field_value, str)
+        ]
 
+        target_value = rule.separator.join(string_values)
```
mhoff: I am always a bit weary about this kind of change. We are effectively replacing an error with a silent ignore. Rather raise an error if there is a non-string type in the value list.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 * add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
+* add support for asynchronous rule processing and I/O capability detection to `ng` processors
```
mhoff: From a pure changelog reader perspective I'd think that we are done here and rules are truly processed asynchronously. Maybe emphasize that we did only API-level changes

mhoff (commented):
Many thanks, just a few minor remarks

logprep/ng/processor/concatenator/processor.py
kaya-david: Thanks for the suggestion. This task is mainly about moving the code to async, not refactoring the existing implementation. I’m happy to make the changes, but I’d like to avoid losing focus on the main goal.

logprep/ng/processor/concatenator/processor.py
kaya-david: This task is mainly about moving the code to async, not refactoring the existing implementation. I’d like to avoid losing focus on the main goal.

mhoff: I fully agree. But you are currently changing the logic from raising an error (which `str.join` does for non-string as parameters) to silently ignoring non-str values.

The problem mypy raised on `rule.separator.join(source_field_values)` is valid and leads to runtime errors if there are non-str values present. But silently ignoring these values will not work as we are changing our application behavior, which would be a breaking change.

kaya-david: Good point, thanks. I’ll solve this part differently.

tests/unit/ng/processor/ip_informer/test_ip_informer.py
```diff
@@ -10,14 +10,14 @@
...
 )
 from tests.unit.processor.ip_informer.test_ip_informer import (
-    test_cases as non_ng_test_cases,
+    test_cases as ng_test_cases,
 )
 
-test_cases = deepcopy(non_ng_test_cases)
-failure_test_cases = deepcopy(non_ng_failure_test_cases)
+test_cases = deepcopy(ng_test_cases)
+failure_test_cases = deepcopy(ng_failure_test_cases)
```
mhoff: The exact same pattern (import non_ng test_cases in ng world, make a copy and call it test_cases) is present in many test files and changing the naming convention for only one intrduces inconcistencies.

We can can certainly discuss the naming (while I personally still prefer it the way it was), but I figure you would not want to do the complete refactoring in this PR. So, please simply revert this change.

tests/unit/processor/concatenator/test_concatenator.py
```diff
@@ -182,3 +185,23 @@ def test_process_raises_field_exists_warning_if_target_field_exists_and_should_n
...
+                "delete_source_fields": False,
+            },
+        }
+        self._load_rule(rule)
+
+        document = {"field": {"a": "first", "b": True}, "other_field": {"c": "third"}}
+        result = self.object.process(document)
+
+        assert len(result.errors) == 1
+        assert isinstance(result.errors[0], ProcessingCriticalError)
```
mhoff: Expected errors (which this is) should be handled by producing a `ProcessingWarningError` instead of a `ProcessingCriticalError`. The latter is problematic in the old architecture and leads to the deletion of the event. Normally this should either be fixed immediately (by doing `self._handle_warning_error`) or delayed by adding a corresponding TODO with high priority.

kaya-david: Implemented as discussed. Please have a look and check whether I understood and implemented it correctly.

mhoff (approved):
Looks good, thanks!

PR #1006, opened 2026-08-05
feat: processors async

kaya-david:
## Description

* `ng` async infrastructure
  * add asynchronous getter interfaces and NG getter implementations
  * use asynchronous HTTP getters based on `aiohttp`
  * add async scheduling for refreshable getters and asynchronous refresh callbacks
  * isolate refresh callback failures so a failing callback does not stop subsequent refresh processing
  * remove orphaned refreshable getter targets and schedulers after callback cleanup
  * await refreshable getter scheduling in the NG runner
  * move rule tree configuration and rule loading into processor setup
  * initialize rule trees asynchronously for NG processors while preserving synchronous behavior for legacy processors
  * rebuild and replace rule trees atomically during processor setup to support safe repeated initialization
  * add asynchronous NG rule loading for local and HTTP-backed rule sources
  * load NG configuration sources and rule definitions asynchronously
  * reuse refreshed getter cache data in callbacks to avoid duplicate refresh requests
  * share concurrent cache updates per getter target to avoid duplicate asynchronous requests
  * classify processors that perform asynchronous I/O during event processing via `has_asyncio()`
  * clean up shared getter update tasks after cancelled waiters complete
  * prevent running getter updates from recreating removed shared target state
  * share getter initialization across concurrent callers and cancel it during target cleanup
  * roll back initialized pipeline components when manager setup fails or is cancelled
  * offload credentials file validation from the asynchronous event loop
  * refresh independent getter targets concurrently to avoid cross-target refresh delays
 
* `geoip_enricher`
  * make database loading and processing async
  * use async file locking and offload blocking filesystem operations
  * adapt tests to the async setup lifecycle
  * close the GeoIP database reader during processor shutdown

* `grokker`
  * load custom grok patterns asynchronously during setup
  * offload blocking downloads, ZIP extraction, and pattern initialization
  * clean up temporary pattern directories asynchronously
  * adapt tests to asynchronous getter usage and setup

* `requester`
  * use `aiohttp` for asynchronous HTTP requests
  * manage the HTTP client session during setup and shutdown
  * preserve existing request configuration through `aiohttp` parameter conversion
  * convert basic authentication to `Authorization` headers instead of deprecated `aiohttp` auth parameters
  * adapt tests to use an async HTTP test server
  * add coverage for session lifecycle, timeouts, authentication, and async I/O detection
  * harden setup lifecycle handling to avoid leaking existing HTTP sessions on repeated initialization

* `amides`
  * load models asynchronously during setup
  * offload blocking filesystem operations, downloads, ZIP access, and model deserialization
  * use asynchronous NG getters for model downloads

* `domain_resolver`
  * offload blocking DNS resolution waits to avoid blocking the async event loop

* `generic_adder`
  * load URI-based additions asynchronously
  * support static and event-templated URI sources with refresh and cleanup callbacks
  * use asynchronous NG getters for local and HTTP content
  * adapt tests to the async processor lifecycle and getter cache handling

* `generic_resolver`
  * introduce an NG-specific rule implementation using asynchronous getters
  * move file-dependent resolve mapping initialization into async setup
  * load local and HTTP-backed resolve mappings asynchronously
  * refresh HTTP-backed resolve mappings through asynchronous callbacks
  * adapt tests to the async setup and refresh lifecycle
  * clean up and deduplicate refresh callbacks across repeated setup and shutdown
  * avoid sharing mutable additions between rule configuration instances (non-ng + ng)

* `list_comparison` / `network_comparison`
  * introduce NG-specific rule implementations using asynchronous getters
  * load static and dynamic comparison lists asynchronously
  * refresh HTTP-backed comparison lists asynchronously
  * adapt comparison processing and tests to asynchronous compare-set iteration

* `selective_extractor`
  * load `extract_from_file` asynchronously during setup
  * move file-dependent rule initialization out of rule construction

* `pseudonymizer`
  * load regex mappings asynchronously during setup
  * reset asynchronously loaded mapping state during shutdown
  * initialize public key loading asynchronously during processor setup

* `template_replacer`
  * load replacement templates asynchronously during setup

* `requester`
  * preserve timeout, TLS certificate, and environment proxy behavior when using aiohttp

* misc
  * disable unused WebSocket support in the asynchronous HTTP server to avoid deprecation warnings
 
* tests
  * adapt NG processor tests to explicit async setup and shutdown lifecycle handling
  * add reusable setup/shutdown handling for processor tests
  * add async scheduler coverage for synchronous and asynchronous jobs and pending-job execution
  * add dedicated NG getter coverage including async HTTP, retries, caching, credentials, and TLS handling
  * update getter, rule tree, rule loading, refresh, dynamic URI, and resolver tests for asynchronous APIs
  * use async HTTP test servers for NG HTTP-dependent processor tests
  * wait for finite acceptance pipelines to finish processing before collecting their output
  * remove the temporary NG rule-tree getter monkeypatch

## Assignee

### Documentation

### Code Quality

* pytest

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--1006.org.readthedocs.build/).


logprep/framework/rule_tree/rule_tree.py
```diff
@@ -103,29 +102,87 @@ class Config:
...
+    def _initialize(self, config_data: dict) -> None:
+        """Initialize the rule tree with loaded configuration data.
+
+        Parameters
+        ----------
+        config_data: dict
+            Parsed rule tree configuration.
+        """
+        self.tree_config = RuleTree.Config(**config_data)
+        self.rule_parser = RuleParser(self.tree_config.tag_map)
```
mhoff: I'd suggest a different route which is more in line with how the other components behave: Don't make the component (`RuleTree`) instantiate it's own config, but inject it as a constructor argument (same as `Processor`, `Input` and `Output`). Let the caller (here `Processor.setup`) handle the conditional check and produce a `RuleTree.Config` object. This can then simply be passed. This is also the ideal branching point. We have sync and async `Processor.setup`, so we can nicely implement the `Getter` call in their respective methodology.

logprep/ng/abc/getter.py
```diff
@@ -0,0 +1,201 @@
...
+
+logger = logging.getLogger("Getter")
+yaml = YAML(typ="safe", pure=True)
+
+
+ContentType: TypeAlias = str
+
+
+@define(kw_only=True)
+class Getter(ABC):
```
mhoff, line 21: Did you consider inheriting the sync `Getter` here? This comes with the stylistic downsides of having overrides which violate the method, but it would make for less redundant code

logprep/framework/rule_tree/rule_tree.py
```diff
@@ -103,29 +102,87 @@ class Config:
...
+
+        if config:
+            config_data = getter_factory.from_string(config).get_dict()
+
+        self._initialize(config_data)
+
+    async def init_async(
+        self,
+        config: str | None,
+        getter_factory,
```
mhoff: If we were to keep this, we'd need a type here. But see the more important comment below first.

logprep/ng/abc/processor.py
```diff
@@ -275,9 +286,22 @@ def _write_target_field(self, event: dict, rule: "Rule", result: Any) -> None:
             )
 
     async def setup(self) -> None:
-        """Set up the processor."""
+        """Set up the processor"""
 
         await super().setup()
 
+        rule_tree = RuleTree()
```
mhoff, line 293: Why don't we use `self._rule_tree` here which we constructed in __init__?

logprep/ng/abc/processor.py
```diff
@@ -210,17 +210,28 @@ def test_rules(self) -> dict | None:
 
         """
 
-    def load_rules(self, rules_targets: Sequence[str | dict]) -> None:
+    async def load_rules(
+        self,
+        rules_targets: Sequence[str | dict],
+        rule_tree: RuleTree | None = None,
+    ) -> None:
```
mhoff, line 217: What is the exact use case for passing `rule_tree=None`? I'd assume we typically want to operate on `self._rule_tree` either way

logprep/ng/processor/amides/processor.py
```diff
@@ -205,29 +206,33 @@ def _evaluate_cmdline_cached(self):
...
             models["multi"],
             self._config.num_rule_attributions,
         )
 
-    def _load_and_unpack_models(self):
+    async def _load_and_unpack_models(self):
         models_path = self._config.models_path
-        if not Path(models_path).exists():
+
+        if not await asyncio.to_thread(Path(models_path).exists):
```
mhoff, line 220: 1) Is there not proper async Path library to use here?
2) As I understand it `to_thread` simply calls `loop.run_in_executor` using the default `ThreadPoolExecutor`. Do we define how many threads this executor has? Seems like an important performance tuning parameter to me

logprep/ng/processor/amides/processor.py
```diff
@@ -205,29 +206,33 @@ def _evaluate_cmdline_cached(self):
...
+        return await asyncio.to_thread(self._load_models_from_archive, models_path)
+
+    @staticmethod
+    def _load_models_from_archive(models_path: str):
         with ZipFile(models_path, mode="r") as zip_file:
             with zip_file.open("model", mode="r") as models_file:
-                models = joblib.load(models_file)
-
-        return models
+                return joblib.load(models_file)
```
mhoff, line 235: This is sync, isn't it?

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -220,14 +221,19 @@ def _add_resolve_infos_to_event(
...
-    def _resolve_ip(self, domain: str) -> tuple[str | None, int]:
+    async def _resolve_ip(self, domain: str) -> tuple[str | None, int]:
         """Resolve domain with timeout.
 
         Assumes socket default timeout is None and relies on threading to create a timeout.
+        The blocking wait for the thread result is executed in a worker thread to avoid
+        blocking the event loop.
         """
         try:
             result = self._thread_pool.apply_async(socket.gethostbyname, (domain,))
```
mhoff, line 232: Did you consider https://github.com/aio-libs/aiodns?

logprep/ng/processor/generic_adder/rule.py
```diff
@@ -0,0 +1,610 @@
...
+    identifiers: Sequence[str]
+    content_by_uri: dict[str, FieldValue] = field(factory=dict)
+    error: Exception | None = None
+
+
+class GenericAdderRule(Rule):
+    """Check if documents match a filter and initialize the fields and values can be added."""
+
+    @define(kw_only=True)
+    class Config(Rule.Config):
```
mhoff, line 168: Data classes can be inherited/imported from the non-ng world. This applies to all copied rules (and files)

logprep/ng/processor/generic_resolver/processor.py
```diff
@@ -214,5 +215,17 @@ def _update_cache_metrics(self) -> None:
         self.metrics.cache_load += cache_info.currsize / self.max_cache_entries
 
     async def setup(self) -> None:
+        """Set up the processor and initialize its rules."""
         await super().setup()
+
+        RefreshableGetter.remove_callbacks_for_tag(self._job_tag_for_cleanup)
```
mhoff, line 221: I think we should only clean up afterwards and not before

logprep/ng/processor/generic_resolver/processor.py
```diff
@@ -214,5 +215,17 @@ def _update_cache_metrics(self) -> None:
...
+
+        for rule in self.rules:
+            rule = typing.cast(GenericResolverRule, rule)
```
mhoff, line 224: Rather use the same approach as the last processor:

```
    @property
    def rules(self) -> Sequence[GenericAdderRule]:
        """Returns all rules"""
        return typing.cast(Sequence[GenericAdderRule], super().rules)
```

logprep/ng/processor/generic_resolver/processor.py
```diff
@@ -214,5 +215,17 @@ def _update_cache_metrics(self) -> None:
...
+            await rule.setup(self._job_tag_for_cleanup)
```
mhoff, line 225: Even though I like the wording, the established pattern for rules is to not have a `setup` but rather an `init`. At least as far as I know

logprep/ng/processor/geoip_enricher/processor.py
```diff
@@ -60,49 +60,61 @@ class Config(FieldManager.Config):
...
 
-    def _try_getting_geoip_data(self, ip_string: str) -> dict:
+    async def _try_getting_geoip_data(self, ip_string: str) -> dict:
         try:
             ip_addr = str(ip_address(ip_string))
-            ip_data = self._city_db.city(ip_addr)
+            ip_data = await asyncio.to_thread(
+                self._city_db.city,
+                ip_addr,
+            )
```
mhoff, line 117: This is probably outdated

logprep/ng/processor/geoip_enricher/processor.py
```diff
@@ -157,3 +169,17 @@ async def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
...
+
+        if city_db is not None:
+            await asyncio.to_thread(city_db.close)
+            del self._city_db
+
+        await super().shut_down()
+
+    async def has_asyncio(self) -> bool:
+        """Return whether the processor performs asynchronous I/O operations."""
+        return True
```
mhoff, line 185: No, we don't have asyncio during event processing

logprep/ng/processor/grokker/processor.py
```diff
@@ -29,26 +29,42 @@
...
 from logprep.util.helper import FieldValue, add_fields_to, get_dotted_field_value
 
 logger = logging.getLogger("Grokker")
 
 
+@asynccontextmanager
+async def _temporary_directory(*, suffix: str = "") -> AsyncIterator[Path]:
+    """Create and clean up a temporary directory without blocking the event loop."""
+    temporary_directory = await asyncio.to_thread(
+        tempfile.TemporaryDirectory,
```
mhoff, line 58: No async version available?

logprep/ng/processor/grokker/processor.py
```diff
@@ -105,34 +135,73 @@ async def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
...
                 overwrite_target=rule.overwrite_target,
             )
-        if self._handle_missing_fields(event, rule, rule.actions.keys(), source_values):
+
+        if self._handle_missing_fields(
+            event,
+            rule,
+            rule.actions.keys(),
+            source_values,
+        ):
```
mhoff: Line too long? This is an extremely standard method call for processors always with nearly the same arguments. Laying it out like this takes up much space with redundant info. I would rather keep it as a one-liner for readability

logprep/ng/processor/grokker/processor.py
```diff
@@ -105,34 +135,73 @@ async def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
...
+                    target_dir=patterns_tmp_path,
                 )
+
                 for rule in self.rules:
-                    rule.set_mapping_actions(patterns_tmp_path)
-                return
+                    await asyncio.to_thread(
+                        rule.set_mapping_actions,
+                        str(patterns_tmp_path),
+                    )
```
mhoff, line 149: Why is this threaded? There is no I/O happening, is there?

logprep/ng/processor/grokker/processor.py
```diff
@@ -105,34 +135,73 @@ async def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
...
+
         if re.search(r"http(s)?:\/\/.*?\.zip", custom_patterns_dir):
-            with tempfile.TemporaryDirectory("grok") as patterns_tmp_path:
-                self._download_zip_file(
-                    source_file=custom_patterns_dir, target_dir=Path(patterns_tmp_path)
+            async with _temporary_directory(suffix="grok") as patterns_tmp_path:
+                await self._download_zip_file(
+                    source_file=custom_patterns_dir,
+                    target_dir=patterns_tmp_path,
                 )
```
mhoff, line 143: We use 3 threads shortly after another. Why not:

```
await raw = getter.get_raw()
to_thread(extract_zip_file, raw)
# ...
rule.set_mapping_actions(...)

def extract_zip_file(raw):
    with tempfile.TemporaryDirectory("grok") as patterns_tmp_path:
        with tempfile.TemporaryFile("wb+") as archive:
            # ...
```

logprep/ng/util/configuration.py
```diff
@@ -1058,18 +1079,25 @@ async def _verify(self) -> None:
...
+
         if ENV_NAME_LOGPREP_CREDENTIALS_FILE in ENV_VARS:
             try:
                 credentials_file_path = ENV_VARS.get(ENV_NAME_LOGPREP_CREDENTIALS_FILE)
                 if credentials_file_path is None:
                     raise ValueError("credentials file path was None but expected it to be set")
-                _ = CredentialsFactory.get_content(Path(credentials_file_path))
+
+                _ = await asyncio.to_thread(
+                    CredentialsFactory.get_content,
```
mhoff, line 1095: CredentialsFactory.get_content should rather employ a Getter call. It is loading data from file, then deserializing and then constructing an object from it. We are only reading a file in there. We could also change the interface such that we only pass and parse the already loaded data. Can be a follow-up ticket

logprep/ng/util/getter.py
```diff
@@ -0,0 +1,1031 @@
...
+        headers = credential_headers | self._headers
+
+        if self.hash:
+            headers["If-None-Match"] = self.hash
+
+        try:
+            async with aiohttp.ClientSession(
+                trust_env=True,
+            ) as session:
+                for attempt in range(self._MAX_RETRIES + 1):
```
mhoff, line 957: Are we diverging here? I don't see the retry loop in non-ng

logprep/ng/util/rule_loader.py
```diff
@@ -0,0 +1,302 @@
...
+    async def load_rules(self) -> list[Rule]:
+        """Generate and return a list of rules from the specified
+        dictionary.
+
+        Returns
+        -------
+        list of Rule
+          A list with one Rule object created from the dictionary.
+        """
+        from logprep.registry import Registry  # pylint: disable=import-outside-toplevel
```
mhoff, line 290: Any idea why we do this? I see the same for the non-ng version, but it feels unnecessary to me

logprep/ng/util/rule_loader.py
```diff
@@ -0,0 +1,302 @@
...
+    ) -> list[str]:
+        """return a list of rule files in the source directory"""
+        if extensions is None:
+            extensions = RULE_FILE_EXTENSIONS
+
+        extensions = tuple(extensions)
+
+        return await asyncio.to_thread(
+            lambda: [str(path) for path in Path(source).glob("**/*") if path.suffix in extensions]
+        )
```
mhoff, line 131: Why not

```suggestion
        child_paths = await asyncio.to_thread(Path(source).glob("**/*"))
        return [str(path) for path in child_paths if path.suffix in extensions]
```

logprep/ng/util/rule_loader.py
```diff
@@ -0,0 +1,302 @@
...
+from typing import Iterable, Sequence
+
+from logprep.ng.util.getter import GetterFactory
+from logprep.processor.base.rule import Rule
+from logprep.util.defaults import RULE_FILE_EXTENSIONS
+
+logger = getLogger("RuleLoader")
+
+
+class RuleLoader:
```
mhoff, line 16: This module should be refactored completely. The logic is way simpler than the class structure suggests and spreads I/O around the place.

Let's consider the (most complex) use case of loading a source path which is a directory:
- Find all files in the directory (glob)
- Load content for each file in parallel (asynchronous gather on multiple Getters)
- Create rule from each loaded dict/content

logprep/ng/manager.py
```diff
@@ -66,18 +65,49 @@ async def setup(self) -> None:
         if error_output is None:
             logger.warning("No error output configured")
 
-        await asyncio.gather(*(component.setup() for component in self._components))
```
mhoff, line 69: I assume you rather want something like this: https://github.com/fkie-cad/Logprep/blob/main/logprep/ng/util/workflow/workflow.py#L135

`return_exceptions=True` is the right fit and automatically handles cancellation

logprep/util/async_scheduler.py
```diff
@@ -0,0 +1,102 @@
+"""Async wrappers for schedule.Job and schedule.Scheduler."""
+
+import asyncio
+import datetime
+import inspect
+from collections.abc import Hashable
+from typing import Any
+
+from schedule import CancelJob, Job, Scheduler
```
mhoff, line 9: Seems like the schedule project is rather outdated and the modifications we do in this file add a lot of complexity. Maybe we should fork the project and add async support. Or maybe there is a project which has more native async support

logprep/util/async_scheduler.py
```diff
@@ -0,0 +1,102 @@
...
+            return CancelJob
+
+        return ret
+
+
+class AsyncScheduler:
+    """Scheduler supporting synchronous and asynchronous jobs."""
+
+    def __init__(self) -> None:
+        self._scheduler = Scheduler()
```
mhoff, line 43: I think inheritance would have been the better approach here. Yes, overwriting sync with async methods is not nice, but the code would be a lot clearer and less non-performant wrapping. One could also add alternative run methods like here: https://github.com/dbader/schedule/pull/647

tests/acceptance/util.py
```diff
@@ -180,11 +178,16 @@ def get_patched_runner(config_path):
...
     def keep_iterating():
-        """generator that stops on first iteration"""
-        return
-        yield
+        """Wait for all pipelines to finish and return an empty iterator"""
+        for pipeline in runner._manager._pipelines:
+            pipeline.join(timeout=10)
+
+            if pipeline.is_alive():
+                raise TimeoutError(f"Pipeline {pipeline.name} did not finish within 10 seconds")
```
mhoff, line 188: For which use cases is this required?

tests/unit/ng/processor/amides/test_amides.py
```diff
@@ -203,25 +203,42 @@ async def test_setup_get_model_via_file_getter(self, tmp_path, monkeypatch):
...
-        responses.add(responses.GET, model_uri, model_original_content)
+
+        async def handler(_: web.Request) -> web.Response:
+            return web.Response(
+                body=model_original_content,
+                content_type="application/zip",
+            )
+
+        app = web.Application()
+        app.router.add_get("/model.zip", handler)
```
mhoff, line 223: This is very verbose. I'd hope for a less verbose way to mock async https requests, even if we have to develop it ourselves

tests/unit/ng/processor/domain_resolver/test_domain_resolver.py
```diff
@@ -286,3 +289,24 @@ async def test_delete_source_field(self, _):
         log_event = LogEvent(document, original=b"", input_meta=InputMeta())
         await self.object.process(log_event)
         assert log_event.data == expected
+
+    async def test_resolve_ip_waits_asynchronously(self):
+        result = mock.Mock()
+
+        self.object._thread_pool.apply_async = mock.Mock(return_value=result)
```
mhoff, line 296: Why do we need this?

tests/unit/ng/processor/generic_adder/test_generic_adder.py
```diff
@@ -48,12 +47,23 @@ class TestGenericAdder(BaseProcessorTestCase[GenericAdder]):
         "rules": ["tests/testdata/unit/generic_adder/rules"],
     }
 
+    @pytest.fixture(autouse=True)
+    def reset_refreshable_getters(self):
+        RefreshableGetter.reset()
+        yield
+        RefreshableGetter.reset()
```
mhoff, line 54: conftest already does that: https://github.com/fkie-cad/Logprep/blob/main/tests/conftest.py#L47

Why did you need that?

tests/unit/ng/processor/generic_adder/test_generic_adder.py
```diff
@@ -69,10 +79,13 @@ async def test_generic_adder_testcases_failure_handling(
...
     async def test_dynamic_uri_failure_handling(self, rule, event, error_message):
-        await self._load_rule(rule)
-        await self.object.setup()
+        config = deepcopy(self.CONFIG)
+        config["rules"] = [rule]
+        self.object = Factory.create({"test instance": config})
+
         log_event = LogEvent(event, original=b"", input_meta=InputMeta())
 
+        await self.object.setup()
```
mhoff, line 88: What was the problem with the old code?

tests/unit/ng/processor/generic_resolver/test_generic_resolver.py
```diff
@@ -45,193 +47,424 @@ class TestGenericResolver(BaseProcessorTestCase[GenericResolver]):
...
+            app.router.add_get("/resolve-mapping", handler)
+            server = await aiohttp_server(app)
+
+            rule["generic_resolver"]["resolve_from_file"]["path"] = str(
+                server.make_url("/resolve-mapping")
+            )
+
+        config["rules"] = [rule]
+
+        processor = Factory.create({"test instance": config})
```
mhoff, line 94: All the mocking boilerplate needs to be handled by `provision_context`. That is what the method is for.

tests/unit/ng/processor/generic_resolver/test_generic_resolver.py
```diff
@@ -45,193 +47,424 @@ class TestGenericResolver(BaseProcessorTestCase[GenericResolver]):
...
+
+            app = web.Application()
+            app.router.add_get("/resolve-mapping", handler)
+            server = await aiohttp_server(app)
+
+            rule["generic_resolver"]["resolve_from_file"]["path"] = str(
+                server.make_url("/resolve-mapping")
+            )
+
+        processor = self._create_test_instance(
```
mhoff, line 141: Sometimes you use `self.object`, sometimes `Factory.create` and sometimes `_create_test_instance` (even in the same file). I don't understand the direction where we are going with this PR

mhoff (changes requested):
Hello @kaya-david, thanks for the work. I have added a few remarks and have a few questions. See below

PR #1010, opened 2026-08-07
fix: allow None as valid input via add in `generic_adder`

Pablu23:
## Description

generic_adder: allow None as valid input via `add`

## Assignee

### Documentation

### Code Quality

* Tests

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--1010.org.readthedocs.build/).

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--1010.org.readthedocs.build/).

### DIFF
--- a/logprep/processor/generic_adder/processor.py
+++ b/logprep/processor/generic_adder/processor.py
@@ -54,7 +54,12 @@ class GenericAdder(Processor):
             for items_to_add in rule.additions(event):
                 if items_to_add:
                     add_fields_to(
-                        event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target
+                        event,
+                        items_to_add,
+                        rule,
+                        rule.merge_with_target,
+                        rule.overwrite_target,
+                        skip_none=False,
                     )
         except Exception as error:
             raise ProcessingWarning(str(error), rule, event) from error
--- a/logprep/processor/generic_adder/rule.py
+++ b/logprep/processor/generic_adder/rule.py
@@ -93,6 +93,7 @@ from attrs import define, field, validators
 from logprep.abc.getter import Getter
 from logprep.filter.expression.filter_expression import FilterExpression
 from logprep.processor.base.rule import InvalidRuleDefinitionError, Rule
+from logprep.util import helper
 from logprep.util.converters import convert_from_dict
 from logprep.util.environ import ENV_VARS
 from logprep.util.getter import GetterFactory, RefreshableGetter
@@ -167,10 +168,10 @@ class GenericAdderRule(Rule):
         of the source fields.
         """
 
-        add: dict = field(
+        add: dict[str, FieldValue] = field(
             validator=validators.deep_mapping(
                 key_validator=validators.instance_of(str),
-                value_validator=validators.instance_of((str, bool, list, dict, int, float)),
+                value_validator=helper.field_value_validator,
             ),
             factory=dict,
         )
--- a/logprep/util/helper.py
+++ b/logprep/util/helper.py
@@ -23,6 +23,8 @@ from typing import (
     cast,
 )
 
+from attrs import Attribute
+
 from logprep.processor.base.exceptions import FieldExistsWarning
 from logprep.util.defaults import DEFAULT_CONFIG_LOCATION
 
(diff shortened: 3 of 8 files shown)


logprep/processor/generic_adder/rule.py
```diff
@@ -167,10 +168,10 @@ class Config(Rule.Config):
         of the source fields.
         """
 
-        add: dict = field(
+        add: dict[str, FieldValue] = field(
```
mhoff, line 171: Do we already have a test for the `None` case? If not, that would be useful. Oftentimes `None` is treated like a missing value which leads to it being skipped.

logprep/util/helper.py
```diff
@@ -939,3 +941,19 @@ def transform_field_value(
             return transform_value(data)
         case _:
             raise ValueError(f"unexpected type encountered: {type(data)}")
+
+
+def field_value_validator(_: object, attribute: Attribute, value: object) -> None:
```
mhoff, line 946: pydoc

logprep/processor/generic_adder/processor.py
```diff
@@ -54,7 +54,12 @@ def _apply_rules(self, event: dict[str, FieldValue], rule: GenericAdderRule):
             for items_to_add in rule.additions(event):
                 if items_to_add:
                     add_fields_to(
-                        event, items_to_add, rule, rule.merge_with_target, rule.overwrite_target
+                        event,
+                        items_to_add,
+                        rule,
+                        rule.merge_with_target,
+                        rule.overwrite_target,
+                        skip_none=False,
```
mhoff, line 62: We need a CHANGELOG entry for the `None` support (which also changes the interface slightly, but it's not a breaking change I'd say)

mhoff:
And please update the PR description. Then we are ready to go

tests/unit/util/test_helper.py
```diff
@@ -2,6 +2,7 @@
 # pylint: disable=no-self-use
 import re
 import typing
+from types import SimpleNamespace
```
mhoff, line 5: double check pls

tests/unit/processor/generic_adder/test_generic_adder.py
```diff
@@ -399,6 +399,12 @@
         {"float_value_test": "Test", "float_field": 12.3},
         id="Add float value",
     ),
+    pytest.param(
+        {"filter": "*", "generic_adder": {"add": {"null_field": None}}},
+        {"null_value_test": "Test"},
+        {"null_value_test": "Test", "null_field": None},
+        id="Add null value",
```
mhoff, line 406: https://github.com/fkie-cad/Logprep/blob/main/tests/conftest.py#L176

Pablu23: Please look at the current implementation is that how you envisioned it?

tests/unit/util/test_helper.py
Pablu23: This is needed because we expect a certain error message, that uses attribute.name, if thats not set this would fail, because of multiple reasons. We could also use a mock here but SimpleNamespace from the std lib is just the easiest way to create an arbitrary object that fits attribute, with a name field that has the value "add"

tests/unit/processor/generic_adder/test_generic_adder.py
```diff
@@ -399,6 +399,12 @@
...
+    ),
```
mhoff:
```suggestion
    *[pytest.param(
        {"filter": "*", "generic_adder": {"add": {"some_field": test_case.values[0]}}},
        {"message": "Test"},
        {"message": "Test", "some_field": test_case.values[0]},
        id=f"Add single value: {test_case.id}",
    ) for test_case in FIELD_VALUE_TEST_CASES],
```

tests/unit/processor/generic_adder/test_generic_adder.py
mhoff: Actually I rather meant something like seen below. There are certainly pros and cons to it. What do you think? I think the main question is if these are good examples or not. I actually tend towards a no, but we should keep it consistent across all tests

tests/unit/util/test_helper.py
mhoff: Yes, but it's also surprising which makes it harder to maintain. In a test I would either expect a mock or a true `attrs.Attribute` and not some object which is completely unrelated and simply shares the single used attribute `name`.

I would propose to simply replace it with a mock:
```
mock = Mock()
mock.name = "add"

...
```

Pablu23: Yes you are right I agree, I changed this

tests/unit/processor/generic_adder/test_generic_adder.py
Pablu23: I see, I changed this now, I also think those arent actually good examples, but they are good / necissary tests

tests/unit/util/test_helper.py
```diff
@@ -496,3 +497,25 @@ def test_error_on_illegal_field_value(self):
...
+        "value, expected_error",
+        [
+            ({"nested": ["value", 42, 13.37, True, None, {"child": False}]}, None),
+            ({1: "value"}, "add must have string keys"),
+            ({"nested": [object()]}, "add must be a FieldValue, got object"),
+        ],
+    )
+    def test_validates_recursive_field_values(self, value, expected_error):
+        # attribute = SimpleNamespace(name="add")
+
```
mhoff:
```suggestion
```

mhoff (approved):
LGTM. I have checked the preview box as it looks fine and your patch coverage is > 95% ("All modified and coverable lines are covered by tests.")

PR #1012, opened 2026-08-13
feat: add FieldNameReplacer processor

Pablu23:
## Description

add new `field_name_replacer` processor to replace occurences of strings in key names

## Assignee

### Documentation

### Code Quality

- running test
- running benchmark.py

## Reviewer
- Documentation](#documentation) is up-to-date

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--1012.org.readthedocs.build/).


logprep/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class FieldNameReplacer(Processor):
+    """A processor that renames fields in log events."""
+
+    rule_class = FieldNameReplacerRule
+
+    def _apply_rules(self, event: dict, rule: FieldNameReplacerRule) -> None:
+        """Apply one field-name replacement rule to an event."""
+        rule.replace(event)
```
mhoff: Typically rules should provide the data for processor operations. Here we delegate the functionality completely to the rule. I might be wrong and there might be a precedence for the pattern you are following. If not, we should maintain the convention that logic lives on the processor

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,132 @@
...
+                ),
+                iterable_validator=validators.and_(
+                    validators.instance_of(list), validators.min_len(1)
+                ),
+            ),
+            converter=lambda x: x if isinstance(x, list) else [x],
+        )
+        target_separator: str = field(
+            validator=validators.and_(validators.instance_of(str), validators.min_len(1))
+        )
```
mhoff: This wording is in conflict with the processor name. Nothing about "FieldNameReplacer" implies that we are focusing on replacing "separators".

I think I personally like `source_chars` --> `target_chars` better. `collapse_repetitions` or `collapse_repeated_chars` would then be the other field

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,132 @@
...
+                    validators.instance_of(str), validators.min_len(1)
+                ),
+                iterable_validator=validators.and_(
+                    validators.instance_of(list), validators.min_len(1)
+                ),
+            ),
+            converter=lambda x: x if isinstance(x, list) else [x],
+        )
+        target_separator: str = field(
+            validator=validators.and_(validators.instance_of(str), validators.min_len(1))
```
mhoff: Should we allow `target_seperator` to be an empty string? I think this might be a use case at some point

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,132 @@
...
+    def _replace_children(self, root: dict[str, FieldValue]):
+        for key in list(root):
+            new_key = self._replace_key(key)
+            entry = root.pop(key)
+
+            if self.config.recursive:
+                self._replace_value(entry)
+
+            if new_key in root and root[new_key] is not None:
+                if self.config.merge_with_target:
```
mhoff: Use `transform_field_value` here (or rather in the processor and the rule simply provides the key transformation function). However, `transform_field_value` can not handly merging as of now... we'd have to think about this

logprep/registry.py
```diff
@@ -58,6 +58,7 @@ class Registry:
         "template_replacer": "logprep.processor.template_replacer.processor.TemplateReplacer",
         "timestamp_differ": "logprep.processor.timestamp_differ.processor.TimestampDiffer",
         "timestamper": "logprep.processor.timestamper.processor.Timestamper",
+        "field_name_replacer": "logprep.processor.field_name_replacer.processor.FieldNameReplacer",
```
mhoff: Don't forget about the ng-variant later on

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,132 @@
...
+        elif self.config.strip_suffix:
+            key = key.rstrip("".join(self.config.source_separators))
+
+        for sep in self.config.source_separators:
+            if self.config.collapse_repeated_separators:
+                key = re.sub(rf"(?:{re.escape(sep)})+", lambda _: self.config.target_separator, key)
+            else:
+                key = key.replace(sep, self.config.target_separator)
+
+        return key
```
mhoff: All of this needs to be pre-computed and provided as an optimized function for the processor to apply. Regular expressions should be pre-compiled, strings already joined, and maybe even the if...else stuff (`stripper = (lambda s: s.strip(chars)) if ... else (lambda s: s.lstrip(...))`)

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,132 @@
...
+        if self.config.strip_prefix and self.config.strip_suffix:
+            key = key.strip("".join(self.config.source_separators))
+        elif self.config.strip_prefix:
+            key = key.lstrip("".join(self.config.source_separators))
+        elif self.config.strip_suffix:
+            key = key.rstrip("".join(self.config.source_separators))
+
+        for sep in self.config.source_separators:
+            if self.config.collapse_repeated_separators:
+                key = re.sub(rf"(?:{re.escape(sep)})+", lambda _: self.config.target_separator, key)
```
mhoff:
```suggestion
                key = re.sub(rf"(?:{re.escape(sep)})+", self.config.target_separator, key)
```

tests/unit/processor/field_name_replacer/test_field_name_replacer.py
```diff
@@ -0,0 +1,194 @@
+# pylint: disable=missing-docstring
+import pytest
+
+from logprep.processor.base.exceptions import FieldExistsWarning
+from tests.unit.processor.base import BaseProcessorTestCase
+
+test_cases = [
```
mhoff: Have the most user-relevant test-cases emphasized as `example_test_cases`. Don't forget `normalize_test_cases`. See `generic_resolver` tests for inspiration

tests/unit/processor/field_name_replacer/test_field_name_replacer.py
```diff
@@ -0,0 +1,194 @@
...
+    def test_replaces_configured_separators_in_field_names(self, testcase, config, event, expected):
+        rule = {
+            "filter": "test",
+            "field_name_replacer": {
+                "source_fields": ["test"],
+                "source_separators": ".",
+                "target_separator": "__",
+                **config,
+            },
+        }
```
mhoff: This should be part of `pytest.param`. Otherwise the person reading our docs does not know what the rule configuration entails. Feel free to have a helper function to not have this duplicate in all the test_cases

logprep/processor/field_name_replacer/rule.py
Pablu23: No the lambda was intentional here as putting a string into this function or func / lambda does different things. If it is a string, backslashes in the replacement string are processed, for a lambda it is taken verbatim, because I didnt see a case why we should dissallow backslashes as target_separators (gets renamed) I put a lambda there

tests/testdata/unit/field_name_replacer/rules/rule.json
```diff
@@ -0,0 +1,11 @@
...
+    {
+        "filter": "message",
+        "field_name_replacer": {
+            "source_fields": ["message"],
+            "source_separators": ".",
+            "target_separator": "__"
+        },
+        "description": "Replace dots in field names below message."
+    }
+]
```
mhoff: We don't need this test file. The test_cases context can handle this. See `generic_resolver` as an inspiration

Pablu23: No this is still needed to make base rule tree tests pass

logprep/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,30 @@
+"""
```
mhoff, line 1: Also add the processor to the benchmark pipeline configurations and make sure it is being used for at least a subset of the events

logprep/processor/field_name_replacer/rule.py
Pablu23: Yes I was about to comment that aswell. I can extend the transform_field_value helper so that it accepts a collision_handler, but this gets rather large by doing it. Which is fine I guess. Also how would you propose we do the actual path resolution then?

logprep/processor/field_name_replacer/processor.py
Pablu23: I dont know why but I relativly consistently get ~3.350 docs/s on this branch with the field_name_replacer and around ~2.850 docs/s on the main branch? Please check this on your machine aswell. But so far to me it seems, performance is not a concern anymore here

Pablu23: Only talking about the non-ng variant here, havent run the ng version yet

logprep/processor/field_name_replacer/rule.py
mhoff: Interesting. Lambda seems expensive though. Can we activate the same behavior with a re-flag maybe?

Pablu23: No not direclty, but we can also "cache" the lambda, so we dont have to recreate the inner lambda on every replacement. This would also have the benefit of not having to access self.config on every replacement call

logprep/processor/field_name_replacer/processor.py
mhoff: Weird. What we do by adding an event in the input data is changing the proportion of other events in the overall event stream. So an event type which was an 8th before is now an 9th of all events

examples/exampledata/kafka_generate_input_logdata/logclass/test_input.jsonl
```diff
@@ -5,4 +5,5 @@
 {"test_pseudonymizer": "test", "something": "pseudonymize_me"}
 {"test_pseudonymizer": "test", "something_special": "pseudonymize_me"}
 {"test_pre_detector": "test", "something": "test"}
+{"field_name_replacer": {"k8s.application/kind": "value"}}
```
mhoff: Maybe stay consistent here:

```suggestion
{"test_pre_detector": "test", "something": "test"}
{"test_field_name_replacer": {"k8s.application/kind": "value"}}
```

(same for the configs then)

logprep/ng/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,92 @@
+"""
+|PROCESSOR_NAME|
+================
+
+The `field_name_replacer` processor replaces characters in dictionary keys.
+This can make field names safe to store in systems that do not allow dots in keys.
```
mhoff:
```suggestion
This can make field names safe to store in systems that, for instance, do not allow dots in keys.
```

logprep/ng/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,92 @@
...
+        rule = typing.cast(FieldNameReplacerRule, rule)
+
+        for source_field in rule.config.source_fields:
+            path = get_dotted_field_list(source_field)
+
+            curr: dict[str, FieldValue] | FieldValue = event
+            for p in path:
+                if not isinstance(curr, dict):
+                    break
+                curr = curr.get(p)
```
mhoff: This seems to be `get_dotted_field_value` (see helpers)

logprep/ng/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,92 @@
...
+
+            self._replace_value(curr, event, rule)
```
mhoff: Use `transform_field_value` (extended by collision handling) and then `add_fields_to` instead. `merge_with_target` and `overwrite_target` don't have a meaning as long as we have no `target_field`. So use differently named collision handler parameters for that

tests/testdata/unit/field_name_replacer/rules/rule.json
mhoff: I think you checked if we can remove it by refactoring base.py, did that work?

tests/unit/processor/field_name_replacer/test_field_name_replacer.py
```diff
@@ -0,0 +1,203 @@
...
+        id="collapses consecutive characters when configured",
+    ),
+    pytest.param(
+        rule({"characters_to_replace": "*", "collapse_repeated_characters": True}),
+        {"test": {"k8s***application": "value"}},
+        {"test": {"k8s__application": "value"}},
+        id="treats regex metacharacters as literal characters when collapsing",
+    ),
+    pytest.param(
+        rule({"characters_to_replace": "ab", "collapse_repeated_characters": True}),
```
mhoff: Okay, that reads weird. Maybe then `to_replace` or `text_to_replace` is better. From reading the rule definition I would have suspected that we replace all "a"s and all "b"s.

Then `collapse_repeated_characters` should also not refer to `characters` and rather something like `collapse_sequences` or `collapse_repetitions`. Isn't sequences what we actually do? I mean if I replace ["ab", "cd"] with "_", would "abcd" then become "_" or "__"? I think I'd prefer the former

Pablu23: Sorry I dont really understand the question? If you replace "ab" and "cd" with "" then abcd, obviously becomes "" where would the "__" come from? Atleast thats the only thing that makes sense, If I accidently implemented this wrong ill fix it. Ill have another look

mhoff: Sorry for that. Markdown got in the way. Here is how the question should actually read:

Then collapse_repeated_characters should also not refer to characters and rather something like collapse_sequences or collapse_repetitions. Isn't sequences what we actually do? I mean if I replace ["ab", "cd"] with "*", would "abcd" then become "*" or "**"? I think I'd prefer the former

mhoff: Okay, something is really messing with GitHub. Can't even edit my comments right now.

Next attempt:
If I do characters_to_replace=["ab", "cd"] & replacement="#" & collapse_repeated_characters=True, would "abcd" then become "#" or "##"? I think I'd prefer "#" because these are all character sequences I want to remove.

So, all I want to get at is that "characters" is not the best wording as we rather treat the characters as text. So either simply drop "characters" (`to_replace`, `replacement`, `collapse_sequences`) or use "text" (`text_to_replace`, ...). I personally favor the former variant

logprep/ng/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,82 @@
...
+
+    async def _apply_rules(self, event: dict, rule: Rule) -> None:
+        """Apply one rule to the selected parts of an event."""
+        rule = typing.cast(FieldNameReplacerRule, rule)
+
+        for source_field in rule.config.source_fields:
+            value = get_dotted_field_value(event, source_field)
+
+            if not isinstance(value, (dict, list)):
+                continue
```
mhoff, line 53: Do we need this? I would suspect `transform_field_value` to cover this case

Pablu23: Not entirely sure, but this is a short circuit when we already know we cant do anything, the normal `transform_field_value` could also operate on field_values, so we just save some time here, can probably remove, but Id keep it

tests/unit/ng/processor/base.py
mhoff: Does this work?

```suggestion
import yaml

def test_load_rules_calls_getter_factory(self, tmp_path):
        first_rule_ref = self.rules_dirs[0]
        if isinstance(first_rule_ref, str):
            rules_dir = [first_rule_ref]
        elif isinstance(first_rule_ref, dict):
            rules_dir = tmp_path
            rule_file = rules_dir / "rule.yaml"
            rule_file.write_text(yaml.dump(first_rule_ref)
        else:
            raise AssertionError("unexpected type ...")
        
        with mock.patch("logprep.util.getter.GetterFactory.from_string") as getter_factory:
            with pytest.raises(TypeError):
                self.object.load_rules(rules_targets=rules_dir)
            getter_factory.assert_called()
```

Pablu23: I suppose, but why would we want to do that? I find this test to be unneccessary even now, more so with the new change that we do not have to write rule files. Why would we now create ephemeral files just to satisfy this in harsh words useless test? The getter and getterfactory are extensivly tested in their respective files

logprep/ng/processor/field_name_replacer/processor.py
mhoff: Independent of the performance discussion (I suspect isinstance for native types like dict and list is fine performance-wise), I wonder if this check makes sense from a use case perspective. Typically you'd only use the processor for complex data structures which need actual replacing. Using it on a basic field does not make sense, even though I get that this might happen in the wild.

But anyways, if the condition holds true we exit early and if it doesn't hold true the operation is expensive and thus the check does not matter too much. Feel free to keep it and resolve or change it. I am fine either way.

logprep/ng/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,82 @@
...
+            case "merge", dict(), dict():
+                return {**later, **existing}
+            case "merge", list(), list():
+                return [*existing, *later]
+            case "merge", list(), str() | int() | float() | bool():
+                return [*existing, later]
+            case "merge", str() | int() | float() | bool(), list():
+                return [existing, *later]
+            case _:
+                raise FieldExistsWarning(rule, event, [key])
```
mhoff: This should be two different helper methods (so without the self). I'd prefer to raise only a `ValueError` and wrap that in a `FieldExistsWarning` in the calling code, to keep the function pure. But as other helpers also handle `rule` the way you do here, we can also keep it.

logprep/ng/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,82 @@
...
+    ) -> FieldValue:
+
+        match rule.config.collision_strategy, existing, later:
+            case "keep_first", _, _:
+                return existing
+            case "merge", dict(), dict():
+                return {**later, **existing}
+            case "merge", list(), list():
+                return [*existing, *later]
+            case "merge", list(), str() | int() | float() | bool():
```
mhoff: `None` is missing

logprep/ng/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,82 @@
...
+                recursive=rule.config.recursive,
+            )
+
+            add_fields_to(event, {source_field: transformed}, rule=rule, overwrite_target=True)
+
+    def _handle_key_collision(
+        self,
+        key: str,
+        existing: FieldValue,
+        later: FieldValue,
```
mhoff: I think `later` is rather unconventional wording. Something like `old_value` / `new_value` might be an alternative

tests/unit/ng/processor/base.py
mhoff: Fair enough. I just think we shouldn't skip. I'd be fine with removing it, if it does not reduce the coverage.

logprep/util/helper.py
```diff
@@ -926,17 +928,36 @@ def transform_field_value(
     ValueError
         If an unexpected type is encountered
     """
-    match (data):
+
+    def transform_child(value: FieldValue) -> FieldValue:
+        if not recursive:
+            return value
```
mhoff: I would honestly remove the `recursive` flag altogether. I don't think we need it at the moment and this way we have a double recursion.

Rather do something like this:

```
def transform_field_value(
  data: FieldValue,
  /,
  transform_value: Callable[[str | int | float | bool | None], FieldValue],
  transform_key: Callable[[str], str],
  collision_handler: Callable[[str, FieldValue, FieldValue], FieldValue] | None = None
):

  def _recurse(value: FieldValue) -> FieldValue:
    match value:
       case dict:
         # ... transform_key(key) ...
         # ... _recurse(value) ...

return _recurse(data)

```

logprep/ng/processor/field_name_replacer/processor.py
Pablu23: Honestly, I would like to keep this for now, when the time arises, that we copy this exact portion everywhere / we use this 2 or 3 times, we should make this a helper, before I would not bother creating a more generic helper method, just to never use it in the future. Also this would require every using place to adapt, collision_strategy in some way and the developer to know about this

logprep/ng/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,81 @@
...
+
+            if not isinstance(value, (dict, list)):
+                continue
+
+            transformed = transform_field_value(
+                value,
+                transform_key=rule.replace_key,
+                transform_value=lambda leaf: leaf,
+                collision_handler=lambda key, existing, later: self._handle_key_collision(
+                    key, existing, later, event, rule
```
mhoff:
```suggestion
                collision_handler=lambda key, existing, incoming: self._handle_key_collision(
                    key, existing, incoming, event, rule
```

logprep/ng/processor/field_name_replacer/processor.py
mhoff: I don't see the issue tbh. Firstly, my main point is that we should have two methods instead of one, taking the strategy out of each and every match...case. Secondly, "keep_first" does not need a match...case at all and instead becomes a simple parameter returning function.

```
collision_handler = merge_handler if rule.config.collision_strategy == "merge" else keep_first_handler
```

Yes, we should typically only generalize if there are multiple uses. But this already _is_ generic functionality cramped into a specific processor. Except for the strategy part, nothing in here is specific to `field_name_replacer` and I deem it rather likely that we will use the helpers in the future.

And you know how it is. No one will look by themselves in specific processors to see if there is some generic functionality they could re-use. But the helpers.py is a central file which every logprep developer can be expected to know and to respect.

(By the way, why do we use "keep_first" and not "keep_last"? "keep_last" seems to be the default (think about `{**dict1, **dict2}` and "keep_first" is rather an exception, isn't it?)

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,183 @@
...
+class FieldNameReplacerRule(Rule):
+    """Configure how field names are changed in selected event fields."""
+
+    @define(kw_only=True)
+    class Config(Rule.Config):
+        """Configuration for a FieldNameReplacer rule."""
+
+        collision_strategy: str = field(
+            validator=validators.and_(
+                validators.instance_of(str), validators.in_(["keep_first", "merge", "raise"])
```
mhoff: Should be an `enum`

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,183 @@
...
+                    validators.instance_of(str), validators.min_len(1)
+                ),
+                iterable_validator=validators.instance_of(list),
+            )
+        )
+        """Fields whose contained dictionary keys should be changed."""
+
+        strip_prefix: bool = field(validator=validators.instance_of(bool), default=False)
+        """Remove configured text at the beginning of a key. Defaults to :code:`False`."""
+        strip_suffix: bool = field(validator=validators.instance_of(bool), default=False)
```
mhoff, line 120:
```suggestion
        """Remove configured text at the beginning of a key. Defaults to :code:`False`."""

        strip_suffix: bool = field(validator=validators.instance_of(bool), default=False)
```

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,183 @@
...
+    def config(self) -> Config:
+        """Return the rule configuration with its specific type."""
+        return typing.cast(FieldNameReplacerRule.Config, self._config)
+
+    def __init__(self, filter_rule: FilterExpression, config: Config, processor_name: str):
+        super().__init__(filter_rule, config, processor_name)
+
+        # Pre compute strip function
+        alternatives = "|".join(
+            re.escape(text) for text in sorted(self.config.to_replace, key=len, reverse=True)
```
mhoff: That longer strings match first should be documented behavior

logprep/ng/processor/field_name_replacer/processor.py
mhoff: Also, a strategy which merges in-place (`existing.extend(incoming)`) would be important as `[*existing, *incoming]` allocates extra objects and performs very badly in comparision to `extend`

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,183 @@
...
+            prefix_pattern = re.compile(rf"^(?:{alternatives})+")
+        if self.config.strip_suffix:
+            suffix_pattern = re.compile(rf"(?:{alternatives})+$")
+
+        def _strip_func(value: str) -> str:
+            if prefix_pattern is not None:
+                value = prefix_pattern.sub("", value)
+            if suffix_pattern is not None:
+                value = suffix_pattern.sub("", value)
+            return value
```
mhoff: Maybe put this in a method `_create_strip_func`?

We are currently re-using the `alternatives` variable two disjoint contexts

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,183 @@
...
+
+            self._replace_func = lambda value: pattern.sub(lambda _: self.config.replacement, value)
+        else:
+
+            def _replace_func(key: str) -> str:
+                for ctr in self.config.to_replace:
+                    key = key.replace(ctr, self.config.replacement)
+                return key
+
+            self._replace_func = _replace_func
```
mhoff: Same here, `self._replace_func = self._create_replace_func()` and have the rest in a method

tests/unit/processor/field_name_replacer/test_field_name_replacer.py
```diff
@@ -0,0 +1,232 @@
...
+        "field_name_replacer": {
+            "source_fields": ["test"],
+            "to_replace": ".",
+            "replacement": "__",
+            **config,
+        },
+    }
+
+
+INLINE_RULE = {
```
mhoff:
```suggestion
DEFAULT_RULE = {
```

INLINE_RULE tells us how the variable is being used on a technical level. But DEFAULT_RULE describes its semantical usage

tests/unit/processor/field_name_replacer/test_field_name_replacer.py
```diff
@@ -0,0 +1,232 @@
...
+        rule({}),
+        {"test": {"nested": {"k8s.application/kind": "value"}}},
+        {"test": {"nested": {"k8s__application/kind": "value"}}},
+        id="recursively replaces characters by default",
+    ),
+    pytest.param(
+        rule({"collision_strategy": "merge"}),
+        {"test": {"k8s.application": ["source"], "k8s__application": ["target"]}},
+        {"test": {"k8s__application": ["source", "target"]}},
+        id="merges colliding lists in earlier then later order",
```
mhoff:
```suggestion
        id="merges colliding lists by appending",
```

doc/source/configuration/processor.rst
```diff
@@ -35,3 +35,4 @@ Processors
 .. automodule:: logprep.processor.template_replacer.processor
 .. automodule:: logprep.processor.timestamper.processor
 .. automodule:: logprep.processor.timestamp_differ.processor
+.. automodule:: logprep.processor.field_name_replacer.processor
```
mhoff: I can't see it in the doc preview. Can you?

Edit: Nevermind. It's at the end and breaks alphabetic ordering

logprep/registry.py
```diff
@@ -109,6 +110,7 @@ class Registry:
         "template_replacer": "logprep.ng.processor.template_replacer.processor.TemplateReplacer",
         "timestamp_differ": "logprep.ng.processor.timestamp_differ.processor.TimestampDiffer",
         "timestamper": "logprep.ng.processor.timestamper.processor.Timestamper",
+        "field_name_replacer": "logprep.ng.processor.field_name_replacer.processor.FieldNameReplacer",
         # Connectors
```
mhoff: Here as well, maintain the alphabetic order pls

logprep/processor/field_name_replacer/rule.py
mhoff: Than we can remove the comment "# Pre compute strip function" and the other below as well

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def replace_key(self, key: str) -> str:
+        """Return a key after applying the configured character replacements."""
+        return self._replace_func(self._strip_func(key))
```
mhoff, line 205: FYI - if only single characters were involved we could employ `str.translate` + `str.strip` which should be way quicker then what we are doing... too bad

Pablu23: Well we could make to_replace be only one char, but that feels a bit arbitrary to me

logprep/ng/processor/field_name_replacer/processor.py
mhoff: Thanks for extracting the helpers. But we still choose right handler for each call of `transform_key`. As with `replace_key`, this decision should be made once per rule, as the merge strategy does not change

logprep/util/helper.py
```diff
@@ -926,23 +927,85 @@ def transform_field_value(
...
-                transform_field_value, transform_value=transform_value, transform_key=transform_key
-            )
             return [transform_recursive(item) for item in data]
         case str() | int() | float() | bool() | None:
             return transform_value(data)
         case _:
             raise ValueError(f"unexpected type encountered: {type(data)}")
 
 
+def keep_incoming_collision_handler(
```
mhoff, line 974: pydoc missing for all three helpers

logprep/util/helper.py
```diff
@@ -926,23 +927,85 @@ def transform_field_value(
...
+    _key: str,
+    _existing: FieldValue,
+    incoming: FieldValue,
+) -> FieldValue:
+    return incoming
+
+
+def merge_collision_handler(
+    _: str,
```
mhoff:
```suggestion
    _key: str,
```
Better to keep the name as with `keep_incoming_collision_handler`. Maybe even rather `key` instead of `_key` as calling with key-value args makes sense

logprep/util/helper.py
```diff
@@ -926,23 +927,85 @@ def transform_field_value(
...
+            existing.extend(incoming)
+            return existing
+        case list(), str() | int() | float() | bool() | None:
+            existing.append(incoming)
+            return existing
+        case str() | int() | float() | bool() | None, list():
+            incoming.insert(0, existing)
+            return incoming
+        case _:
+            raise ValueError(f"cannot merge collision of {type(incoming)} with {type(existing)}")
```
mhoff: Maybe rather raise a specialized exception which can be parsed easier by calling sites to produce a proper `FieldExistsWarning`. Or keep producing a `FieldExistsWarning` here, which would require passing optional rule objects etc.

logprep/ng/processor/field_name_replacer/processor.py
Pablu23: Ahh yes now I see what you wanted. Be sure to check this again, it should now be correctly implemented

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,216 @@
...
+        self._replace_func: Callable[[str], str] = self._create_replace_func()
+
+    def _create_replace_func(self) -> Callable[[str], str]:
+        if self.config.collapse_sequences:
+            alternatives = "|".join(
+                re.escape(text)
+                # Sorting is required because otherwise overlaps in replacement text,
+                # like ["a", "ab"] would result in the replacement of "ab",
+                # becoming "__b", because "a" matched before "ab" does
+                for text in sorted(self.config.to_replace, key=len, reverse=True)
```
mhoff: Ah, sorry. Earlier I mentioned that the sorting stuff should be documented. But by that I meant in the pydoc of the field, such that the user knows.

I'd suggest to remove the comments here, put the `lambda s: list(sorted(s, key=len, reverse=True))` as a converter on the field and add to the pydoc of the field that longest items take precedence. Same for the copied block below.

logprep/processor/field_name_replacer/rule.py
```diff
@@ -0,0 +1,216 @@
...
+            )
+            pattern = re.compile(rf"(?:{alternatives})+")
+
+            return lambda value: pattern.sub(lambda _: self.config.replacement, value)
```
mhoff, line 174: Might be a bit faster than a lambda but could mess with typing and the extra parameters from sub would be accepted which is unexpected. Feel free to drop this suggestion

```suggestion
            return functools.partial(pattern.sub, lambda _: self.config.replacement)
```

logprep/processor/field_name_replacer/rule.py
mhoff: I agree. Resolved

logprep/util/helper.py
```diff
@@ -926,23 +935,96 @@ def transform_field_value(
...
+
+
+def merge_collision_handler(
+    key: str,
+    existing: FieldValue,
+    incoming: FieldValue,
+) -> FieldValue:
+    """Resolve compatible key collisions by returning a merged value.
+
+    The inputs are not mutated. Raises :class:`ValueError` for incompatible
```
mhoff:
```suggestion
    The inputs are not mutated. Raises :class:`FieldCollisionError` for incompatible
```

logprep/util/helper.py
```diff
@@ -926,23 +935,96 @@ def transform_field_value(
...
+
+def merge_mutating_collision_handler(
+    key: str,
+    existing: FieldValue,
+    incoming: FieldValue,
+) -> FieldValue:
+    """Resolve compatible key collisions by mutating and returning a value.
+
+    This avoids allocating a replacement container where possible. Raises
+    :class:`ValueError` for incompatible value types.
```
mhoff:
```suggestion
    :class:`FieldCollisionError` for incompatible value types.
```

logprep/util/helper.py
```diff
@@ -926,23 +935,96 @@ def transform_field_value(
...
+        case dict(), dict():
+            return {**existing, **incoming}
+        case list(), list():
+            return [*existing, *incoming]
+        case list(), str() | int() | float() | bool() | None:
+            return [*existing, incoming]
+        case str() | int() | float() | bool() | None, list():
+            return [existing, *incoming]
+        case _:
+            raise FieldCollisionError(key)
```
mhoff: Maybe add the involved types in the error message?

logprep/ng/processor/field_name_replacer/processor.py
```diff
@@ -0,0 +1,77 @@
...
+
+            if not isinstance(value, (dict, list)):
+                continue
+
+            transformed = transform_field_value(
+                value,
+                transform_key=rule.replace_key,
+                transform_value=leaf_handler,
+                collision_handler=collision_handler,
+            )
```
mhoff: Why not here?

```suggestion
            try:
                transformed = transform_field_value(
                    value,
                    transform_key=rule.replace_key,
                    transform_value=leaf_handler,
                    collision_handler=rule.collision_handler,
                )
            except FieldCollisionError as error:
                raise FieldExistsWarning(rule, event, [key]) from error
```

mhoff (approved):
LGTM, many  thanks!

PR #1016, opened 2026-08-19
perf: improve metrics performance

mhoff:
## Description

* metrics: cache labeled child metric collectors
* metrics: avoid heavy time context manager and hardwire passtrough metric collector methods on wrapper class
* metrics: remove `__add__` interface from metrics and use the passthrough instead
* metrics: consolidate `measure_time` and `measure_time_async` in single decorator
* tests: add mock_env decorator support for async functions

## Assignee

### Documentation

### Code Quality

- pytest

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--1016.org.readthedocs.build/).

### DIFF
--- a/logprep/metrics/metrics.py
+++ b/logprep/metrics/metrics.py
@@ -171,8 +171,11 @@ class Metric(ABC, Generic[M]):
 
     _collector: M = field(default=None, init=False)
 
-    _collector_methods: ClassVar[set[str]] = set()
+    _collector_methods: ClassVar[set[str]]
     """Methods marked as tracked method collected on subclass init"""
+
+    _value_series_suffix: ClassVar[str]
+    """Suffix implemented by subclasses to select the right series carrying the metric value"""
 
     def __init_subclass__(cls, **kwargs) -> None:
         super().__init_subclass__(**kwargs)
@@ -238,14 +241,14 @@ class Metric(ABC, Generic[M]):
         """Return the labelled child of the collector for the given labels"""
         return self._collector.labels(**(self.labels | labels))
 
-    def _bind(self, child: M) -> M:
-        """Bind the child's exposed methods directly onto this instance."""
+    def _bind(self, child: M) -> Self:
+        """Bind the child's exposed methods directly onto this instance"""
         for name in self._collector_methods:
             setattr(self, name, getattr(child, name))
-        return child
+        return self
 
-    def _lazy_bind_default_child(self) -> M:
-        """Bind the default child on first use and return it."""
+    def _lazy_bind_default_child(self) -> Self:
+        """Bind the default child on first use and return the metric."""
         return self._bind(self._labeled_child(self.labels))
 
     def child_collector(self, labels: dict[str, str], inject_label_values: bool = True) -> Self:
@@ -260,9 +263,11 @@ class Metric(ABC, Generic[M]):
     def _init_collector(self) -> M:
         """Create the concrete prometheus metric object"""
 
-    @abstractmethod
-    def __add__(self, other):
-        """Increment the metric by the given value"""
+    @property
+    def value(self) -> float:
+        """The value this metric currently exports, summed over its labelled children."""
+        series_name = f"{self.fullname}{self._value_series_suffix}"
+        return sum(sample.value for sample in self.collect_samples() if sample.name == series_name)
 
     def collect_samples(self) -> list[Sample]:
         """Return the samples of the whole collector, including every labelled child."""
@@ -360,6 +365,9 @@ class Metric(ABC, Generic[M]):
 class CounterMetric(Metric[Counter]):
     """Wrapper for prometheus Counter metric"""
 
+    _value_series_suffix: ClassVar[str] = "_total"
+    """A counter exports its value under the `_total` series"""
+
     @property
     def collector_type(self) -> type[Counter]:
         return Counter
@@ -375,11 +383,7 @@ class CounterMetric(Metric[Counter]):
     @_collector_method
     def inc(self, amount: float = 1, exemplar: dict[str, str] | None = None) -> None:
         """Increment the counter. Rebinds to the collector's method on first call."""
-        self._lazy_bind_default_child().inc(amount, exemplar)
-
-    def __add__(self, other: Any) -> Self:
-        self.inc(other)
-        return self
+        return self._lazy_bind_default_child().inc(amount, exemplar)
 
     def add_with_labels(self, other: Any, labels: dict) -> None:
         """Deprecated method. Always creates a metric with labels set and adds/sets the value"""
@@ -389,6 +393,15 @@ class CounterMetric(Metric[Counter]):
 @define(kw_only=True)
 class HistogramMetric(Metric[Histogram]):
     """Wrapper for prometheus Histogram metric"""
+
+    _value_series_suffix: ClassVar[str] = "_sum"
+    """The histogram value is the sum of the observations"""
+
+    @property
+    def count(self) -> float:
+        """How many observations were made, summed over the labelled children"""
+        series = f"{self.fullname}_count"
+        return sum(sample.value for sample in self.collect_samples() if sample.name == series)
 
     @property
     def collector_type(self) -> type[Histogram]:
@@ -406,16 +419,7 @@ class HistogramMetric(Metric[Histogram]):
     @_collector_method
     def observe(self, amount: float, exemplar: dict[str, str] | None = None) -> None:
         """Observe a value. Rebinds to the collector's method on first call."""
-        self._lazy_bind_default_child().observe(amount, exemplar)
-
-    @_collector_method
-    def time(self) -> Any:
-        """Time a block or function. Rebinds on first call."""
-        return self._lazy_bind_default_child().time()
-
-    def __add__(self, other) -> Self:
-        self.observe(other)
-        return self
+        return self._lazy_bind_default_child().observe(amount, exemplar)
 
     def add_with_labels(self, other: Any, labels: dict) -> None:
         """Deprecated method. Always creates a metric with labels set and adds/sets the value"""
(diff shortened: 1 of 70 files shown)


mhoff:
```
ref=perf-metrics/ng=0:
=== FINAL BENCHMARK SUMMARY ===
runs:                  3
total runtime:         135.001 s
total processed:       396_439

throughput (weighted): 2,936.57 docs/s
throughput (median):   2,893.38 docs/s
throughput (average):  2,936.57 docs/s
throughput (min/max):  2,818.17 / 3,098.17 docs/s
throughput (std dev):  144.91 docs/s
================================

ref=main/ng=0:
=== FINAL BENCHMARK SUMMARY ===
runs:                  3
total runtime:         135.002 s
total processed:       381_822

throughput (weighted): 2,828.28 docs/s
throughput (median):   2,909.08 docs/s
throughput (average):  2,828.28 docs/s
throughput (min/max):  2,609.40 / 2,966.35 docs/s
throughput (std dev):  191.70 docs/s
================================

ref=perf-metrics/ng=1:
=== FINAL BENCHMARK SUMMARY ===
runs:                  3
total runtime:         135.000 s
total processed:       1_000_121

throughput (weighted): 7,408.28 docs/s
throughput (median):   7,444.49 docs/s
throughput (average):  7,408.28 docs/s
throughput (min/max):  7,335.22 / 7,445.13 docs/s
throughput (std dev):  63.27 docs/s
================================

ref=main/ng=1:
=== FINAL BENCHMARK SUMMARY ===
runs:                  3
total runtime:         135.000 s
total processed:       884_902

throughput (weighted): 6,554.81 docs/s
throughput (median):   6,556.98 docs/s
throughput (average):  6,554.81 docs/s
throughput (min/max):  6,550.20 / 6,557.25 docs/s
throughput (std dev):  3.99 docs/s
================================
```

logprep/metrics/metrics.py
```diff
@@ -166,198 +203,211 @@ def __attrs_post_init__(self):
...
+    def value(self) -> float:
+        """The value this metric currently exports, summed over its labelled children."""
+        series_name = f"{self.fullname}{self._value_series_suffix}"
+        return sum(sample.value for sample in self.collect_samples() if sample.name == series_name)
+
+    def collect_samples(self) -> list[Sample]:
+        """Return the samples of the whole collector, including every labelled child."""
+        metrics = self._collector.collect()
+        assert isinstance(metrics, list) and len(metrics) == 1, ".collect() implementation changed"
+        return metrics[0].samples
```
mhoff, line 274:
```suggestion
        return next(iter(metrics)).samples
```

```suggestion
        return itertools.chain.from_iterable(m.samples for m in metrics)
```

```suggestion
        metrics_list = list(metrics)
        if len(metrics_list) != 1:
            # raise error
        return metrics_list[0].samples
```

logprep/metrics/metrics.py
```diff
@@ -146,18 +160,41 @@ class Metric(ABC):
...
+
+    _value_series_suffix: ClassVar[str]
+    """Suffix implemented by subclasses to select the right series carrying the metric value"""
+
+    def __init_subclass__(cls, **kwargs) -> None:
+        super().__init_subclass__(**kwargs)
+        cls._collector_methods = {
+            name
+            for klass in cls.__mro__
+            for name, value in vars(klass).items()
```
Pablu23: klass?

logprep/metrics/metrics.py
```diff
@@ -166,198 +203,211 @@ def __attrs_post_init__(self):
...
+
+    def _bind(self, child: M) -> Self:
+        """Bind the child's exposed methods directly onto this instance"""
+        for name in self._collector_methods:
+            setattr(self, name, getattr(child, name))
+        return self
+
+    def _lazy_bind_default_child(self) -> Self:
+        """Bind the default child on first use and return the metric."""
+        return self._bind(self._labeled_child(self.labels))
```
Pablu23: How is this inherently lazy? When I see a function clearly labeled as lazy, I think calling this function will not do anything yet, but setup things in a way that the next call to something will evaluate this

logprep/metrics/metrics.py
```diff
@@ -166,198 +203,211 @@ def __attrs_post_init__(self):
...
+                try:
+                    result = func(*args, **kwargs)
+                finally:
+                    duration = perf_counter() - begin
+                    getattr(self.metrics, metric_name).observe(duration)
+                append_measurement(self, args, duration)
                 return result
 
-            return inner
+            return timed_and_appended_async if is_async else timed_and_appended
```
Pablu23, line 357: Would like to see this in the else block honestly

logprep/metrics/metrics.py
```diff
@@ -366,25 +416,28 @@ def _init_tracker(self):
...
         )
 
-    @property
-    def tracker(self) -> Histogram:
-        """Returns the prometheus metric object"""
-        return typing.cast(Histogram, self._tracker)
+    @_collector_method
+    def observe(self, amount: float, exemplar: dict[str, str] | None = None) -> None:
+        """Observe a value. Rebinds to the collector's method on first call."""
+        return self._lazy_bind_default_child().observe(amount, exemplar)
```
Pablu23: I find this name confusing, for me observe would, either take a callback that gets called when the observed value changes, or return the value, not mutate it?

tests/unit/ng/component/base.py
```diff
@@ -138,9 +137,7 @@ def test_expected_metrics_attributes_are_initialized(self):
             )
             metric_name = metric_name.replace("logprep_", "")
             metric_attribute = getattr(self.object.metrics, metric_name)
-            assert metric_attribute.tracker is not None
-            possibile_tracker_types = (Counter, Gauge, Histogram)
-            assert isinstance(metric_attribute.tracker, possibile_tracker_types)
```
Pablu23, line 143: possible_tracker_types

logprep/metrics/metrics.py
Pablu23: I find the current implementation to be the best / most valid one

logprep/metrics/metrics.py
mhoff: "class" is a keyword in python and `cls` is already used. I can call it `super_cls`

logprep/metrics/metrics.py
mhoff: Fair point. I will call it `bind_default_child`. The lazy part comes from the call-site

logprep/metrics/metrics.py
mhoff: Do you mean this?:

```
if is_async:
   def foo():
      # ...
   return foo

else:
   def bar():
      # ...
   return bar
```

What I like would like to keep is both inner functions on the same indentation level. I find that easier to read. In the above structure pylint will complain that the else can dropped, as it's unnecessary (due to the return in the then-block). Or did you think about something else?

logprep/metrics/metrics.py
mhoff: Yea, but this is how prometheus client object works. I just pass it through. People familiar with prometheus would probably find this easier to read and understand. We _could_ rename it, but I'd like to keep it for consistency tbh

tests/unit/ng/component/base.py
mhoff: Yea, that's what I removed ;)

logprep/metrics/metrics.py
Pablu23: I actually meant, 
```
if not append:

else: 
```
exactly like you said, so both append and not append are on the same indentation level

tests/unit/ng/component/base.py
Pablu23: Sorry thats on me then :P

logprep/metrics/metrics.py
Pablu23: Yea im fine with that then, lets not rename it to stay consistent with prometheus

logprep/metrics/metrics.py
mhoff: I see and I actually like that better as well, but then we'd need 

```
if not append_to_event:  # pylint: disable=no-else-return
```

(or disable the rule somewhere else like whole file or globally).

Assuming we stick with the pylint rule, I see two paths:

A:

```
def append_measurement(...): ...

if not append_to_event:  # pylint: disable=no-else-return
  def foo(): ...

  def bar(): ...

  return ...
else:
  def foo2(): ...
    append_measurement(...)

  def bar2(): ...
    append_measurement(...)

  return ...
```

because only this way there is a chance to have enough vertical display space to actually see `bar` and `foo2` in alignment, which is the core benefit of having one alignment level imho.
Downside is the rule disable and that `append_measurement` becomes logically more disconnected.

B: keep it as it is

logprep/metrics/metrics.py
```diff
@@ -166,198 +203,209 @@ def __attrs_post_init__(self):
...
+        for name in self._collector_methods:
+            child_collector_method = getattr(child_collector, name)
+            setattr(self, name, child_collector_method)
+        return self
+
+    def child_collector(self, labels: dict[str, str], inject_label_values: bool = True) -> Self:
+        """Return a child metric configured with the given labels"""
+        child = attrs.evolve(
+            self, labels=self.labels | labels, inject_label_values=inject_label_values
+        )
```
Pablu23, line 256: What does evolve do?

logprep/metrics/metrics.py
Pablu23: Sorry for the late answer I think we should just keep it as it is then

Pablu23 (commented):
Looks good to me, please quickly update the doc/source/development/processor_how_to.rst:190 to use .set instead of +=

logprep/metrics/metrics.py
mhoff: Create a copy of an attrs-based objects with some attributes deviating. The child receives the combined labels and by default immediately registers the child metric with prometheus with a 0 value (`inject_label_values`). The main use case for a child collector is having an individual metric per dynamic object. So for instance a counter metric per partition of a topic. The parent metric on the component itself should typically not register, but as soon as we evolve a child it does make sense

Pablu23 (approved):
Sorry that this took so long, I thought I had already approved this, looks good to me

PR #1018, opened 2026-08-25
refactor: optimize calculator

krempel-4711:
## Description

* calculator: extended expression functionality (add modulo operator and the `min`, `max`, `any` and `not` functions)
* calculator: optimized runtime expression evaluation (utilizing an AST setup at at init)

## Assignee

### Documentation

### Code Quality

* run `pytest`

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--1018.org.readthedocs.build/).


logprep/processor/calculator/ast/compile.py
```diff
@@ -0,0 +1,256 @@
...
+    assert operator_symbol in COMPARISON_OPERATORS
+    operator_type = COMPARISON_OPERATORS[operator_symbol]
+    return operator_type(lhs, rhs)
+
+
+def _setup_bnf() -> ParserElement:
+    bnf = Forward()
+
+    e = CaselessKeyword("E")
+    e.set_parse_action(_build_constant)
```
mhoff: Probably up to personal preference. But I think for some straigh-forward declaration this might be more readable as it takes up less vertical screen and gives you a better overview of what is happening

```suggestion
    e = CaselessKeyword("E").set_parse_action(_build_constant)
```

logprep/processor/calculator/ast/parse.py
```diff
@@ -0,0 +1,256 @@
...
+
+    variable_as_hex = (
+        Suppress("from_hex")
+        + Suppress("(")
+        + Optional(Suppress("0x"))
+        + Suppress("${")
+        + Regex(DottedTemplate.braceidpattern)
+        + Suppress("}")
+        + Suppress(")")
+    )
```
mhoff, line 192: I like it that you reuse the `DottedTemplate.braceidpattern`. It feels like injecting an implementation detail of an unrelated component though. Maybe introduce a global constant `logprep.util.helper.VARIABLE_PATTERN` which the `DottedTemplate` and the `calculator` can simply use.

If you don't want to make changes to helper.py, I'd at least like a constant like the following:
```
# somewhere at the top
VARIABLE_PATTERN = DottedTemplate.braceidpattern
```

Would this work? Or do you feel like it's less clear?

```suggestion
    variable_as_hex = (
        Suppress("from_hex(")
        + Optional(Suppress("0x"))
        + Suppress("${")
        + Regex(VARIABLE_PATTERN)
        + Suppress("})")
    )
```

logprep/processor/calculator/ast/compile.py
```diff
@@ -0,0 +1,256 @@
...
+    number.set_parse_action(_build_constant)
+
+    variable = Suppress("${") + Regex(DottedTemplate.braceidpattern) + Suppress("}")
+    variable.set_parse_action(_build_variable)
+
+    hex_number = Regex(r"0x[0-9a-f]")
+    number_from_hex = (
+        Suppress("from_hex")
+        + Suppress("(")
+        + Regex(r"(0x)?[a-f0-9]+", flags=RegexFlag.IGNORECASE)
```
mhoff:
```
# somewhere at the top
HEX_PATTERN = r"[0-9a-f]" # maybe even [0-9a-fA-F]
```

```suggestion
    hex_number = Regex(rf"0x{HEX_PATTERN}") # also add a + here?
    number_from_hex = (
        Suppress("from_hex")
        + Suppress("(")
        + Regex(rf"(0x)?{HEX_PATTERN}+", flags=RegexFlag.IGNORECASE) # why ignore case here and not above?
```

logprep/processor/calculator/ast/compile.py
```diff
@@ -0,0 +1,256 @@
...
+        + Suppress("}")
+        + Suppress(")")
+    )
+    variable_as_hex.set_parse_action(_build_hex_variable)
+
+    hex_number_handling = hex_number | number_from_hex | variable_as_hex
+
+    ident = Word(alphas, alphanums + "_$")
+    plus, minus, mult, div, mod = map(Literal, "+-*/%")
+    lpar, rpar = map(Suppress, "()")
```
mhoff: Now we switch from explicitly creating `Suppress("(")` each time to producing it once and re-using it. One can clearly argue one way or another, but I'd like to have it consistent

logprep/processor/calculator/ast/compile.py
```diff
@@ -0,0 +1,256 @@
...
+    addop = plus | minus
+    multop = mult | div | mod
+    expop = Literal("^")
+    comparisonop = one_of(">= <= == != > <")
```
mhoff: The following seems clearer to me. Apparently `one_of` does support `Iterable` as well.

```suggestion
    comparisonop = one_of([">=", "<=", "==", "!=", ">", "<"])
```

logprep/processor/calculator/processor.py
```diff
@@ -23,66 +23,51 @@
...
+        for rule in self.rules:
+            rule.init_calculator()
+
     def _apply_rules(self, event, rule):
-        source_field_dict = get_source_fields_dict(event, rule)
-        if self._handle_missing_fields(event, rule, rule.source_fields, source_field_dict.values()):
-            return
-        if self._has_missing_values(event, rule, source_field_dict):
-            return
+        assert isinstance(rule, CalculatorRule)
```
mhoff: Even though I like this use of `isinstance` better, it's currently the established pattern to do `rule = typing.cast(...)`. We can certainly decide to use `isinstance` here, but then we should plan to do this in all implementations of `_apply_rules`

logprep/processor/calculator/processor.py
```diff
@@ -23,66 +23,51 @@
...
+
+        @timeout(seconds=rule.timeout)
+        def calculate():
+            return rule.program.evaluate(event)
```
mhoff: This is an ad-hoc decoration in each call of `_apply_rules`. But I see the difficulty as `rule.timeout` is dynamic and thus the decoration can not easily be applied statically. Maybe we can find a better solution to this, like using a context manager instead of an decorator

logprep/processor/calculator/rule.py
```diff
@@ -174,6 +177,28 @@ def __attrs_post_init__(self):
             self.source_fields = re.findall(FIELD_PATTERN, self.calc)
             super().__attrs_post_init__()
 
+    def __init__(self, filter_rule, config, processor_name):
+        super().__init__(filter_rule, config, processor_name)
+        # TODO: do i really need init call or could je directly
+        # compile (and maybe fail at) init
```
mhoff: True. I think doing it in `__init__` is fine here. There is not a big difference and I'd always favor `__init__` for static fail fast scenarios. Obviously I/O etc. should never (and in async _can_ never) be handled in `__init__`, but this is irrelevant here.

logprep/processor/calculator/rule.py
```diff
@@ -174,6 +177,28 @@ def __attrs_post_init__(self):
...
+        def compile():
+            program = compile_expression(self.calc)
+            self.__program = program.optimize()
+
+        compile()
+
+    @property
+    def program(self) -> ASTNode:
+        assert self.__program
+        return self.__program
```
mhoff: I think "expression" (or even "parsed_expression") would be a good name as well. "program" feels big to me

logprep/processor/calculator/rule.py
```diff
@@ -174,6 +177,28 @@ def __attrs_post_init__(self):
...
+        # compile (and maybe fail at) init
+        self.__program: ASTNode | None = None
+
+    def init_calculator(self) -> None:
+
+        # NOTE: Timeouts might occur especially on optimizing
+        # when doing certain calculations (i.e. 9^9^9)
+        @timeout(seconds=self.timeout)
+        def compile():
+            program = compile_expression(self.calc)
```
mhoff: I'd suggest to remove the `calc` property from the rule and simply do it inline. I think these properties were used earlier when mypy typing was not too advanced. Nowadays we ensure there is one config property with the proper typing and then access to correctly typed attribute from there. Only `timeout` is different as this was probably used from the outside (so from the processor)

```suggestion
            program = compile_expression(self.config.calc)
```

Also regarding the wording. This is rather a "parse" than a "compile", right?

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,583 @@
...
+    @abstractmethod
+    def complexity(self) -> int: ...
+
+    @abstractmethod
+    def optimize(self) -> "ASTNode": ...
+
+    def get_diagram(self) -> str:
+        diagram_render_context = DiagramRenderContext()
+        self.walk(diagram_render_context)
+        return diagram_render_context.get_graph_viz()
```
mhoff: Maybe rather have a utility function `node_to_diagram` which does exactly this but not as a method of the node class

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,583 @@
...
+        return True
+
+    def __repr__(self):
+        return f"<constant {self.value !r}>"
+
+    def evaluate(self, context):
+        return self.value
+
+    def optimize(self):
+        return type(self)(value=self.value)
```
mhoff, line 203: Do we need to copy? Can't we simply return `self` here? Might even be the default for `optimize` to return `self`

```suggestion
        return self
```

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,583 @@
...
+
+    def _preprocess_field_data(self, value: Any) -> Any:
+        return value
+
+    def evaluate(self, context):
+        value = get_dotted_field_value(context, self.path)
+        if value is None:
+            raise MissingValueError(f"Missing value for field {self.path!r}.")
+        value = self._preprocess_field_data(value)
+        return parse_value(value, self.output_type)
```
mhoff: It probably makes sense to exclude `None` for now. There is nothing too useful we can do with it. But `None` currently combines two cases: field does not exist and field does exist but is empty. There is also `get_dotted_field_value_with_missing` which makes this difference more explicit by providing `MISSING` and `None` separately. Might be useful for better error messages

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,583 @@
...
+    def __repr__(self):
+        return f"<variable {self.path !r}>"
+
+    def optimize(self):
+        return type(self)(path=self.path)
+
+
+class HexNumberVariableASTNode(VariableASTNode):
+    def _preprocess_field_data(self, value):
+        return read_hex_number(value)
```
mhoff: Another variant would be the below. I think I only thought about this because I initially had no intuitition what `_preprocess_field_data` is for. So maybe it is also simply a matter of wording or the missing pydoc.

```
class VariableASTNode:

    def _get_context_value(self, context):
      # get_dotted_field_value, None/MISSING check

    def evaluate(self, context):
        value = self._get_context_value(contet)
        return parse_value(value, self.output_type)
  
class HexNumberVariableASTNode:
    def _get_context_value(self, context):
      value = super()._get_context_value(self, context)
      return read_hex_number(value)
```

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,583 @@
...
+
+
+class CompositeASTNode(ASTNode):
+    input_type: ClassVar[ValueType]
+
+    def __init__(self, *children: ASTNode):
+        if not all(child.output_type.can_be_cast_to(self.input_type) for child in children):
```
mhoff, line 286: Is this the correct way around? I did not dive too deep here, but why should I cast an output value to an input value? Only for recursive scenarios I'd assume. Maybe reading further code clears this up

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,583 @@
...
+    output_type = ValueType.NUMBER
+
+    def __init__(self, inner: ASTNode):
+        super().__init__(inner)
+        self.inner = inner
+
+    def optimize(self):
+        optimized_inner = self.inner.optimize()
+        if optimized_inner.is_constant:
+            return ConstantNumberASTNode(-optimized_inner.evaluate({}))
```
mhoff: Maybe introduce a method do get the constant value (obviously failing, if not `is_constant`)?

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,583 @@
...
+            return ConstantNumberASTNode(
+                self.operation_fn(
+                    lhs_optimized.evaluate({}),
+                    rhs_optimized.evaluate({}),
+                )
+            )
+        if specific_optimization := self._operation_specific_optimizations(
+            lhs_optimized, rhs_optimized
+        ):
+            return specific_optimization
```
mhoff, line 387: Operation specific optimizations _could_ also be handled by overwriting optimize and doing something after the regular optimization. Obvious pre-condition would be to return `self` rather than a copy if nothing changed. Happy to talk about this. I see pros and cons on both sides.

tests/testdata/unit/calculator/ast/diagram.gv
```diff
@@ -0,0 +1,33 @@
...
+    n4 -> n8;
+    n5 -> n6;
+    n5 -> n7;
+    n9 -> n10;
+    n9 -> n11;
+    n11 -> n12;
+    n11 -> n13;
+    n13 -> n14;
+    n13 -> n15;
+}
```
mhoff: Not sure if we want this (at least as a dedicated file; can also be inlined in a test). Let's talk about this

mhoff (commented):
Many thanks, I like the changes a lot. I had a first look and added a few remarks of which some are really negligible and others we should have a quick talk about. I didn't spot anything blocking

logprep/processor/calculator/ast/node.py
krempel-4711: Optimize in the end should return a self contained new AST. Returning self would create a situation where the new ASTNode is part of the new AND the old AST. Evaluationwise this should be fine (from all i see now but might be problematic if someone extends the code in a funny way) but GC would be doomed to linger on the old AST even if I do not reference it anymore (now that I write those lines I am not sure if that assumption is really true).

logprep/processor/calculator/ast/node.py
krempel-4711: We are indeed in some kind of recursion while evaluating the AST. Let me phrase that line in other words: "The outputs of all children (i.e. oprands used or function parameters) must be castable to the input type expected by the parent node (that represents i.e. an operation or function call)"

I thing this is right as it is, though things like this can be easily confused so I apprichiate if you double check if my reasinging is correct here.

logprep/processor/calculator/ast/node.py
krempel-4711: Decided to stick with deepcopying.

doc/source/_templates/calculator/examples-renderer.tmpl
```diff
@@ -0,0 +1,12 @@
...
+{% macro render_case(expression, result) %}
+* :code:`{{expression}}` => :code:`{{result }}`
+{% endmacro %}
+
+{# --- actual template --- #}
+{% set primary = data.static_expression_example_test_cases %}
+
+{% for item in primary %}
+   {{- render_case(item[0], item[1]) }}
+{% endfor %}
```
mhoff: Maybe rather:

```suggestion
{% macro render_case(expression, result) %}
:code:`{{expression}}` => :code:`{{result }}`
{% endmacro %}

{% for item in data.static_expression_example_test_cases %}
   * {{- render_case(item[0], item[1]) }}
{% endfor %}
```

logprep/processor/calculator/ast/function_registry.py
```diff
@@ -0,0 +1,215 @@
...
+    ASTNode,
+    FunctionCallASTNode,
+    NotFunctionASTNode,
+    NumericFunctionCallASTNode,
+)
+
+ArgBounds: TypeAlias = tuple[int | EllipsisType, int | EllipsisType]
+
+
+@dataclass(frozen=True, kw_only=True)
```
mhoff: We are heavily employing attrs.define for other dataclass-like classes. I can see that user data (instance per event) and static data (like this) might be different use cases. I'd prefer to use attrs.define where applicable and only use something else if there is a clear criterium for doing so. Otherwise we might end up with unclear conventions and subtle errors

logprep/processor/calculator/ast/function_registry.py
```diff
@@ -0,0 +1,215 @@
...
+class Function:
+    """A registry entry for a function"""
+
+    names: Sequence[str]
+    """Names of the function can be invoked with"""
+    description: str
```
mhoff, line 30: Lately we have been inserting newlines between attribute and pydoc to make it very clear to the reader to which entity the description relates

```suggestion
    """Names of the function can be invoked with"""

    description: str
```

logprep/processor/calculator/ast/function_registry.py
```diff
@@ -0,0 +1,215 @@
...
+    Raises
+    ------
+    UnknownFunctionError
+        Raised if the requested function is unknown.
+    InvalidSyntaxError
+        Raised if the number of passed children does not match the number of
+        expected parameters.
+    """
+    if not (function_info := _FUNCTION_NAME_MAP.get(function_name.lower())):
+        raise UnknownFunctionError(f"Unknown function {function_name !r}.")
```
mhoff: I'm rather used to having no space between the variable name and "!r". Happy to discuss. We should teach the formatter to enforce what we decide

```suggestion
        raise UnknownFunctionError(f"Unknown function {function_name!r}.")
```

logprep/processor/calculator/ast/function_registry.py
```diff
@@ -0,0 +1,215 @@
...
+    min_param_count, max_param_count = function_info.arg_bounds
```
mhoff: I am open to use the walrus operator, but there are a few things going on in this line and I feel like it's a bit much. Walrus, truthy conversion (which can be a footgun), negation and two function calls.

```suggestion
    function_info = _FUNCTION_NAME_MAP.get(function_name.lower())
    if function_info is None:
        raise UnknownFunctionError(f"Unknown function {function_name !r}.")
```

logprep/processor/calculator/ast/function_registry.py
```diff
@@ -0,0 +1,215 @@
...
+        names=["multiply"],
+        description="Multiply two given values (legacy).",
+        function=operator.mul,
+        arg_bounds=(2, 2),
+    ),
+    NumericFunction(
+        names=["hypot"],
+        description="Multidimensional Euclidean distance from the origin to a point.",
+        function=math.hypot,
+        arg_bounds=(1, ...),
```
mhoff: I think I understand why chose to use `...` as an open boundary here. It's python native, there is already a proper type of it and for people who know `...` has a clear meaning.

However, looking at this line one might suspect that `(1, ...)` is a python language construct. I think explicit (optional)`min_arg_count` and `max_arg_count` would be better


```suggestion
        min_arg_count=1,
        max_arg_count=None, # could be omitted
```

logprep/processor/calculator/ast/function_registry.py
```diff
@@ -0,0 +1,215 @@
...
+class ProxyFunction(Function):
+    """Base class for a registry entries for proxy ProxyFunctionASTNodes"""
+
+    function: Callable[..., Any]
+    """The proxy function utilized by the ASTNode"""
+
+    @property
+    def additional_kwargs(self) -> dict[str, Any]:
+        """The keyword arguments needed for node init"""
+        return {"function": self.function}
```
mhoff, line 58: I wonder if we should remove `additional_kwargs` and do a switch on the type below. I know this sounds less clean at first, because one should typically avoid conditionals on static types. But I feel like we (at least currently) only use this dict to have a construction helper for the particular scenario of proxied/wrapped functions. And this is a responsibility of the factory function below and not of the dataclass.

So I'd suggest to instead do something like:
```
match function_info.node_type:
  case ProxyFunction():
    return function_info.node_type(function_name, *children, function=function_info.function)
  case _:
    return function_info.node_type(function_name, *children)
```

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+    def visit(self, node: "ASTNode", *children: "ASTNode") -> None:
+        """The callback for visiting a node.
+
+        Parameters
+        ----------
+        node : ASTNode
+            The visited node.
+        children : ASTNode
+            The children of the visited node.
+
```
mhoff:
```suggestion
```

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+    def visit(self, node: "ASTNode", *children: "ASTNode") -> None:
+        """The callback for visiting a node.
+
+        Parameters
+        ----------
+        node : ASTNode
+            The visited node.
+        children : ASTNode
+            The children of the visited node.
```
mhoff, line 37: To me, personally, this parameter description does not convey a lot of information. I rather believe it adds more noise than value this way and would - again personally - be fine with having only the single line as a description. Happy to discuss

This would apply to many docstrings below as well

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+    -------
+    str
+        The GraphViz code of the diagram visualizing the AST.
+    """
+    diagram_render_context = _DiagramRenderContext()
+    node.walk(diagram_render_context)
+    return diagram_render_context.get_graph_viz()
+
+
+EvaluationContext: TypeAlias = dict[str, FieldValue]
```
mhoff: `dict[str, FieldValue]` should actually be a re-used type in helper.py. We use it a lot as it is the event payload root construct. Then we _might_ skip the `EvaluationContext` - I am undecided if this is better or worse

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+        -------
+        ASTNode
+           An optimized version of this node. Will return a deepcopy if no
+           optimization is possible for consistency.
+
+        Raises
+        ------
+        DivisionByZeroError
+            Some optimizations might result in detecting a zero division.
+
```
mhoff:
```suggestion
```

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+        Will recursively call the walk function on child nodes.
+
+        Parameters
+        ----------
+        context : ASTWalkContext
+            The context object used for scanning the AST.
+        """
+
+    @abstractmethod
+    def evaluate(self, context: EvaluationContext) -> Any:
```
mhoff: Can you give me an impression on how expensive replacing `Any` with `int | float | bool` would be? I understand that full-blown type checking would be tricky and would need to be verified statically, but specializing `Any` to the set of generally valid value and return types should be fine, right?

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+
+    def walk(self, context):
+        return context.visit(self)
+
+
+class ConstantASTNode(TerminalASTNode):
+    """Base type for nodes representing constant values"""
+
+    def __init__(self, value: Any):
+        self.value = parse_value(value, self.output_type)
```
mhoff: I'd suggest to use `@attrs.define` for all nodes and declare the attributes this way.

```
@attrs.define(slots=True, frozen=True)
class ConstantASTNode(TerminalASTNode):
    """Base type for nodes representing constant values"""

    value: Any = field(converter=attrs.Converter(lambda v, self: parse_value(value, self.output_type), takes_self=True))
    """The constant value of this node"
```

Granted, this one is more complicated because the custom converter takes more effort than usual. But documenting public attributes is important and this way you can quickly read through the class doc and attributes doc. Otherwise you would need to screen the `__init__` for attribute docs and also use pydoc in statements (as used below).

The codebase makes heavy use of attrs and I think it makes sense to follow this pattern here. Or at least there should be a good reason not to.

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+        optimized_inner = self.inner.optimize()
+        if optimized_inner.is_constant:
+            return ConstantNumberASTNode(-_constant_value(optimized_inner))
+        return NegateASTNode(optimized_inner)
+
+    def evaluate(self, context):
+        return -self.inner.evaluate(context)
+
+    def __repr__(self) -> str:
+        return "<negate>"
```
mhoff, line 330: Normally I would have expected something like `f"neg({self.inner})"` as the string representation of a node (meaning I would have expected the string to contain the child representations as well). I think I understand why you did it this way - because the graph viz walker uses repr to produce individual labels. 

See also https://believemy.com/en/glossaries/python/repr

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+    output_type = ValueType.NUMBER
+
+
+class AddASTNode(ArithmeticASTNode):
+    """Node representing an addition operation"""
+
+    operator_symbol = "+"
+    operation_fn = operator.add
+
+    def _operation_specific_optimizations(self, lhs, rhs):
```
mhoff, line 419: Return type missing (also below)

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+
+    operator_symbol = "%"
+    operation_fn = operator.mod
+
+    def _operation_specific_optimizations(self, lhs, rhs):
+        if _is_constant_value(rhs, 0):
+            raise DivisionByZeroError("Expression resulted to a division by zero on optimization.")
+
+
+class PowASTNode(DivArithmeticASTNode):
```
mhoff, line 493: Is this correct? Can pow produce a div by zero like error?

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+
+    def evaluate(self, context):
+        value = self.value.evaluate(context)
+        lower_bound = self.lower_bound.evaluate(context)
+        op = operator.le if self.lower_bound_is_inclusive else operator.lt
+        if not op(lower_bound, value):
+            return False
+        upper_bound = self.upper_bound.evaluate(context)
+        op = operator.le if self.upper_bound_is_inclusive else operator.lt
+        return op(value, upper_bound)
```
mhoff: Not sure if this is quicker...

```
        # on init:
        self.op_lower = operator.le if self.lower_bound_is_inclusive else operator.lt
        self.op_upper = operator.le if self.upper_bound_is_inclusive else operator.lt
```

```suggestion
        self.upper_bound_is_inclusive = upper_bound_is_inclusive
        self.op_lower = operator.le if self.lower_bound_is_inclusive else operator.lt
        self.op_upper = operator.le if self.upper_bound_is_inclusive else operator.lt

    def optimize(self):
        if self.is_constant:
            return ConstantBooleanASTNode(_constant_value(self))
        return type(self)(
            self.lower_bound.optimize(),
            self.value.optimize(),
            self.upper_bound.optimize(),
            lower_bound_is_inclusive=self.lower_bound_is_inclusive,
            upper_bound_is_inclusive=self.upper_bound_is_inclusive,
        )

    def evaluate(self, context):
        value = self.value.evaluate(context)
        lower_bound = self.lower_bound.evaluate(context)
        if not self.op_lower(lower_bound, value):
            return False
        upper_bound = self.upper_bound.evaluate(context)
        return self.op_upper(value, upper_bound)
```

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+
+class ProxyFunctionCallASTNode(FunctionCallASTNode):
+    """Base class for FunctionCall nodes that utilize a callback function"""
+
+    def __init__(
+        self,
+        function_name: str,
+        *children: ASTNode,
+        function: Callable[..., Any],
+    ):
```
mhoff, line 653: The parameter order is a bit surprising IMHO. Why not put `function` before `*children`?

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,741 @@
...
+
+
+class NotFunctionASTNode(LogicFunctionASTNode):
+    """A node representing the 'not' function."""
+
+    def evaluate(self, context):
+        return not parse_value(
+            self.children[0].evaluate(context),
+            self.input_type,
+        )
```
mhoff, line 700: This is slightly inconsistent with how we dealt children in other parts of the ast node hierarchy. 


```suggestion
class NotFunctionASTNode(LogicFunctionASTNode):
    """A node representing the 'not' function."""

    def __init__(self, function_name: str, inner: ASTNode):
        super().__init__(function_name, inner)
        self.inner = inner

    def evaluate(self, context):
        return not parse_value(
            self.inner.evaluate(context),
            self.input_type,
        )
```

logprep/processor/calculator/ast/parse.py
```diff
@@ -0,0 +1,271 @@
+"""Functionality to parse expressions to an abstract syntax tree"""
+
+# pylint: disable=missing-docstring
```
mhoff: Why do we need this?

logprep/processor/calculator/ast/parse.py
```diff
@@ -0,0 +1,271 @@
...
+    VariableASTNode,
+)
+from logprep.processor.calculator.ast.util import read_hex_number
+from logprep.util.helper import VARIABLE_PATTERN
+
+
+def _build_constant(parsed: ParseResults) -> ASTNode:
+    assert len(parsed) == 1
+    assert isinstance(parsed[0], str)
+    return ConstantNumberASTNode(parsed[0])
```
mhoff, line 50: What do you think? This condenses the repetetive type related information a bit more and has the same function

```suggestion
    assert len(parsed) == 1 and isinstance(parsed[0], str)
    return ConstantNumberASTNode(parsed[0])
```

logprep/processor/calculator/ast/parse.py
```diff
@@ -0,0 +1,271 @@
...
+    assert len(parsed) >= 2
+    signs = parsed[:-1]
+    assert all(sign in ("+", "-") for sign in signs), signs
+    node = parsed[-1]
+    if isinstance(node, ParseResults):
+        assert len(node) == 1
+        node = node[0]
+    assert isinstance(node, ASTNode)
+    if len([s for s in signs if s == "-"]) % 2 == 1:
+        return NegateASTNode(node)
```
mhoff, line 92: Seems a bit less mixed up this way. First node, then signs. Instead of signs, node, signs.

```suggestion
    assert len(parsed) >= 2
    *signs, node = parsed
    if isinstance(node, ParseResults):
        assert len(node) == 1
        node = node[0]
    assert isinstance(node, ASTNode)
    assert all(sign in ("+", "-") for sign in signs), signs
    if len([s for s in signs if s == "-"]) % 2 == 1:
        return NegateASTNode(node)
```

logprep/processor/calculator/ast/parse.py
```diff
@@ -0,0 +1,271 @@
...
+    return VariableASTNode(parsed[0])
+
+
+def _build_hex_variable(parsed: ParseResults) -> ASTNode:
+    assert len(parsed) == 1
+    assert isinstance(parsed[0], str)
+    return HexNumberVariableASTNode(parsed[0])
+
+
+def _build_atom(parsed: ParseResults) -> ASTNode:
```
mhoff: I feel like this function needs a bit more context, comments or pydoc as it is hard to read (for me). The asserts already help, but it's still tricky to wrap you head around. Also consider the comment below in the same function body.

logprep/processor/calculator/ast/parse.py
```diff
@@ -0,0 +1,271 @@
...
+        assert len(node) == 1
+        node = node[0]
+    assert isinstance(node, ASTNode)
+    if len([s for s in signs if s == "-"]) % 2 == 1:
+        return NegateASTNode(node)
+    return node
+
+
+def _build_fn(parsed: ParseResults) -> ASTNode:
+    assert len(parsed) >= 1, parsed
```
mhoff: I like that we emit `parsed` as message, but this is inconsistent with the other asserts

logprep/processor/calculator/ast/parse.py
```diff
@@ -0,0 +1,271 @@
...
+    function_name = parsed[0]
+    assert isinstance(function_name, str)
+    assert all(isinstance(i, ParseResults) and len(i) == 1 for i in parsed[1:])
+    params = [i[0] for i in parsed[1:]]
+    assert all(isinstance(param, ASTNode) for param in params)
+    return try_create_function_node(function_name, *params)
```
mhoff: Maybe use `function_name, *parsed_params = parsed` here?

logprep/processor/calculator/ast/parse.py
```diff
@@ -0,0 +1,271 @@
...
+
+    if len(parsed) > 5:
+        raise InvalidSyntaxError("Comparisons can not be chained.")
+
+    if len(parsed) == 5:
+        lower_bound = parsed[0]
+        lower_op = parsed[1]
+        value = parsed[2]
+        upper_op = parsed[3]
+        upper_bound = parsed[4]
```
mhoff:
```suggestion
        lower_bound, lower_op, value, upper_op, upper_bound = parsed
```

logprep/processor/calculator/ast/util.py
```diff
@@ -0,0 +1,114 @@
...
+                return value
+            if isinstance(value, str):
+                if value.upper() == "PI":
+                    return math.pi
+                if value.upper() == "E":
+                    return math.e
+                try:
+                    return int(value)
+                except ValueError:
+                    ...
```
mhoff: That's a `pass` right? I see this kind of usage for the first time. Wouldn't raise from be more informative?

logprep/processor/calculator/rule.py
```diff
@@ -33,14 +33,46 @@
    :inherited-members:
    :no-undoc-members:
 
+Expression syntax for |PROCESSOR_NAME|:
+------------------------------------------------
```
mhoff: The doc build actually complains with a warning if the width is not exactly met...

```suggestion
Expression syntax for |PROCESSOR_NAME|:
---------------------------------------
```

logprep/processor/calculator/rule.py
```diff
@@ -74,77 +128,29 @@
...
-
-The calc expression is not whitespace sensitive.
+.. datatemplate:import-module:: tests.unit.processor.calculator.test_ast
+   :template: calculator/examples-renderer.tmpl
+
+The calc expression is not whitespace or case sensitive.
+
+
+Examples for |PROCESSOR_NAME|:
+------------------------------------------------
```
mhoff: Same width

logprep/processor/calculator/rule.py
```diff
@@ -50,18 +82,40 @@
...
 
-Only one comparison operator is allowed per expression. Chained comparisons such as
-:code:`1 < 2 < 3` or :code:`1 < 2 == 2` are not supported.
+The following functions are available, where the function names a case-insensitive:
 
-Boolean values are final results and cannot be reused as operands in arithmetic or comparison
-operations. Unary negation of boolean values is also not supported. This applies to boolean values
-returned by comparisons as well as boolean-returning functions such as :code:`all`.
+.. datatemplate:import-module:: logprep.processor.calculator.ast.function_registry
+   :template: calculator/function-renderer.tmpl
```
mhoff, line 117: I have mixed feelings about this. I like to reduce redundancy by wiring code definitions together with doc generation. However, here we spread docs in 2-3 files which makes it a bit hard to approach. I think at least simple (!) templates should inlined: https://sphinxcontribdatatemplates.readthedocs.io/en/latest/inline.html#an-inline-template

logprep/processor/calculator/rule.py
```diff
@@ -174,10 +180,17 @@ def __attrs_post_init__(self):
             self.source_fields = re.findall(FIELD_PATTERN, self.calc)
             super().__attrs_post_init__()
 
+    def __init__(self, filter_rule, config, processor_name):
+        super().__init__(filter_rule, config, processor_name)
+        assert isinstance(self._config, CalculatorRule.Config)
```
mhoff: Even though this might be better, we typically use a typed config property.

However, we can certainly switch to asserts as these are better suited for this type casting scenario. But we have to make sure to replace the old style with this new style and that -O is truly enabled to disable the asserts during runtime. 

Edit: I have to reposition myself. Using asserts is (generally) not good in this situation. See my argument below.

mhoff (commented):
Many thanks for the nice work. I have added a few remarks.

Also, please change the title to "PREFIX: optimize calculator", where PREFIX is e.g. "perf" or "refactor". I feel like both are fitting to some extent.

Also, please fill out the complete PR template. Changes (which are basically your changelog items) and how you tested the code (pytest).

logprep/processor/calculator/rule.py
krempel-4711: I strongly disagree when it comes to this specific Function table as its template code is >50 lines long and utilizes mutliple macros.

I changed it for the examples after I reduced the number of lines to 3 for that template. But I think everything that is more than a simple loop schould live in its own file.

logprep/processor/calculator/rule.py
krempel-4711: You suggested getting rid of the typed property:

https://github.com/fkie-cad/Logprep/pull/1018#discussion_r3922920715

mhoff: Yes, but I suggested to use a typed property for the config instead

> Nowadays we ensure there is one config property with the proper typing and then access to correctly typed attribute from there

Like this: https://github.com/fkie-cad/Logprep/blob/main/logprep/ng/processor/grokker/processor.py#L71

As soon as we have fixed the typing in the parallel type hierarchies, we can get rid of the property and all type assertions in this regard. For now we need to fix the typing for each config access and putting asserts everywhere (other processors access the config way more often from different methods) is very boilerplaty

logprep/processor/calculator/rule.py
mhoff: Yes, I agree. I wanted to make a more general point and this particular template is not simple but the other one is -- correct. Resolved

logprep/processor/calculator/ast/util.py
krempel-4711: Yes that is equivalent to writing `pass`. There seems to be no PEP concerning that. I would perfer using `...`over `pass` as I think it is more readable.

In this place a raise would not make sense, the logic is: cast to int, if that fails it tries to parse to float. In the original code there was a nested try block in the except block, would prerfer the non-nested version. But also the following try-except block raises no exception, that is also by design. The function will raise if it does not return anything. So any extra raise statement would be redundant (like an unneccesary else statement).

mhoff: Fair point regarding the raise from.

I did some quick research and `...` seems to be recommended for things which still need to be defined and `pass` is preferred for intentional no-ops. Some argue you could use either and it's completely up to personal preference. To me personally `pass` is more idiomatic for things which are excpliticly and intentionally skipped. `...` I'd use and typically see in abstract methods or typing stubs.

https://yannickbijl.github.io/posts/ellipses-instead-of-pass-keyword/
https://medium.com/@lyx_62906/what-the-mastering-pythons-ellipsis-c747623fc472
...

logprep/processor/calculator/ast/function_registry.py
krempel-4711: In this case those dataclasses are only used in the context of this factory function (except for documetation). So I would consider the dataclasses and the factory as on (kind of) closed, tightly coupled system.
So while I would dislike putting something like `additional_kwargs` on anything that does not meet the criteria stated above, here i think it is fine as it helps us to avoid switching on type later and thus have a generic init logic. One where where specific behaviour can be specified inplace in a rather declaritive way.

logprep/processor/calculator/ast/node.py
krempel-4711: Rephrased this example a bit. In this specific case I guess anyone wanting to implement that would already know what it is about i.e. form the docstring of the `ASTNode.walk` function.
In general I would prefer things to be handled consistently. Yes often it is rather hard to give a description, that adds value (if you gave parameters meaningfull names), BUT - besides consitency - there are two things to consider here:

1. Even if it is just a slight rephrasing that might help be helpfull for some people. (i.e. in this example the information that it is about the node that is *currently visted* might be the small piece of extra information otherwise misisng).
2. While the meaning of some params might be rather obvious, others might, then just covering specific parameters would be rather confusing (though this boils down to consistency).

Though even though I think it is an annyoing chore I think being complete about the documentation is important.

logprep/processor/calculator/ast/node.py
krempel-4711: Created the `JsonObject` alias for `dict[str,FieldValue]` but keept `EvaluationContext` as a type alias (now pointing to `JsonObject`) as it has an specific semantic meaning in this context. Also if one would ever want to extend/refactor that context object that would beneficial.

logprep/processor/calculator/ast/node.py
krempel-4711: Types are specificed on the corresponding method on the base class, usually i would not restate the types on a derived classes as it is redundant. The auto-completion doesn't do that by default and mypy won't complain.

logprep/processor/calculator/ast/node.py
krempel-4711: `0**-1`

logprep/processor/calculator/ast/node.py
krempel-4711: Factory `addtional_kwargs` approach discussed previous.

logprep/processor/calculator/ast/parse.py
krempel-4711: Renamed all `_build_x` functions to `__build_x` two underscores, that lead pylint to ignore the missing docstrings. But i would rather not add docstrings to those functions as they are internal.

logprep/processor/calculator/ast/parse.py
krempel-4711: I would always rather avoid grouping assertions, if you group multiple conditions you will have to find out wich part of the condition failed. If you have many specific single condition assertions this is a no brainer.
 
This part of the code has so many assertions because it was hard to figure out what the `ParseResult` passed to these callbacks will look like. So in this case: If anyone would ever change the syntax ( or write a specific testcase that catches an existing error) it is rather benificial to know immediatly if the problem is the length of the parsed or the type of the first entry.

logprep/processor/calculator/ast/parse.py
krempel-4711: Gave that specific function a more explicit name. Added an actual comment for all `__build_*` functions on top. Its a hard-to-write and hard-to-read internal. That results from the way `pyparsing` organizes those callbacks and that parsing is a quite abstract thing to do :))  
In this case, as everyting except `parse_expression` is an internal and somehow self contained i would stick with the comment on top.

logprep/processor/calculator/ast/node.py
krempel-4711: Acutally that was intentionally set to `Any` as the typing is superimposed by the `ValueType` and `parse_value` functions, and does not really reflect on the AST otherwise. 
Introduced  a new typealias `Value : TypeAlias = int | float | bool`, and applied that to some return values, can be narrowed down without making the code harder to extend in the future.

logprep/processor/calculator/ast/function_registry.py
krempel-4711: Moved the factory method to the registry entry. now the public facing factory function only dispatches towards the registry entry.

logprep/processor/calculator/ast/node.py
mhoff: The thing about pydoc is that it's expensive in terms of writing good/meaningful/value-adding pydoc but also consumes vertical screen estate and can quickly become outdated during development.

I know that this is kind of a controversial topic but I still think per use case differentiation > consistency makes sense as the downsides can be quite cumbersome.

In my own words, we roughly follow these strategies at the moment:
1. No pydoc at all: Only allowed for private methods and should only be used if methods are named explicitly enough that docs would be redundant.
2. Only one line (or 2-3) without params, return or exceptions: Used if the method name and parameter naming are obvious enough.
3. Full blown method pydoc (params, return, exceptions, ...): Used for truly public methods (like other developers using the logprep package) or very complex or very central functions

Compare the following two variants:

```
class ASTWalkContext(Protocol):
    """Protocol for scanning the abstract syntax tree by visitor pattern."""

    def visit(self, node: "ASTNode", *children: "ASTNode") -> None:
        """The callback for the currently visited node and its children"""

# ---

class ASTWalkContext(Protocol):
    """Protocol for scanning the abstract syntax tree by visitor pattern."""

    def visit(self, node: "ASTNode", *children: "ASTNode") -> None:
        """The callback for visiting a node.

        Parameters
        ----------
        node : ASTNode
            The visited node.
        children : ASTNode
            The children of the visited node. 
 
        """
```

Does the lower variant add any value in your opinion other than the consistency argument?

logprep/processor/calculator/ast/node.py
mhoff: Yes type checking is disabled if the signature is not typed: https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code

logprep/processor/calculator/ast/parse.py
mhoff: Simply setting the message to the object in question makes the problem plain by printing it for inspection (wouldn't it?):

```suggestion
    assert len(parsed) == 1 and isinstance(parsed[0], str), parsed
    return ConstantNumberASTNode(parsed[0])
```

logprep/processor/calculator/ast/parse.py
mhoff: Our static checking should only complain if public methods (so no "_" or "__" prefix) are undocumented. Could you double check this? Maybe it's your pylint config?

krempel-4711: does not complain on "_" so the disable was a leftover from some previous point. Though I wonder why this does not give me a warning for the unnecceary disable...

doc/source/_templates/calculator/function-renderer.tmpl
```diff
@@ -0,0 +1,47 @@
...
+   :header-rows: 1
+
+   * - Names
+     - Input Type
+     - Parameter Count
+     - Output Type
+     - Description
+{% for entry in data.FUNCTION_REGISTRY %}
+   {{- render_entry(entry) }}
+{% endfor %}
```
mhoff, line 43: Just a small suggestion. Wouldn't this be a tad clearer? The number and types of columents is immediately described above. The per-column macros are good and fine

```suggestion
.. list-table:: Available Functions
   :header-rows: 1

   * - Names
     - Input Type
     - Parameter Count
     - Output Type
     - Description
{% for entry in data.FUNCTION_REGISTRY %}
   * - {% for name in entry.names %}:code:`{{name}}`{%if not loop.last%},{% endif %} {% endfor %}
     - {{ render_input_type(entry)}}
     - {{ render_bounds(entry.min_args, entry.max_args) }}
     - {{ render_output_type(entry) }}
     - {{ entry.description -}}
{% endfor %}
```

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,671 @@
...
+    def __init__(self, lhs: ASTNode, rhs: ASTNode) -> None:
+        super().__init__(lhs, rhs)
+        self.lhs = lhs
+        """The left-hand-side operand of the operation"""
+
+        self.rhs = rhs
+        """The right-hand-side operand of the operation"""
+
+    def _operation_specific_optimizations(self, lhs: ASTNode, rhs: ASTNode) -> ASTNode | None:
+        # pylint: disable=unused-argument
```
mhoff: Better? Unfortunately calling with kwargs does not work then. Feel free to simply drop this suggestion

```suggestion
    def _operation_specific_optimizations(self, _lhs: ASTNode, _rhs: ASTNode) -> ASTNode | None:
```

logprep/processor/calculator/ast/function_registry.py
```diff
@@ -0,0 +1,225 @@
...
+    ------
+    UnknownFunctionError
+        Raised if the requested function is unknown.
+    InvalidSyntaxError
+        Raised if the number of passed children does not match the number of
+        expected parameters.
+    """
+
+    function_info = _FUNCTION_NAME_MAP.get(function_name.lower())
+    if not function_info:
```
mhoff: At some point in the past we decided to rather go with `is not None` than truthy/falsy. We can have talk about if we want to change this going forward. For now I would ask you to stick to `is not None` or let's have the discussion before this PR gets merged.

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,671 @@
...
+                        _constant_value(lhs_optimized),
+                        _constant_value(rhs_optimized),
+                    )
+                )
+            except ZeroDivisionError as error:
+                raise DivisionByZeroError("Zero division error on optimization") from error
+        if specific_optimization := self._operation_specific_optimizations(
+            lhs_optimized, rhs_optimized
+        ):
+            return specific_optimization
```
mhoff: As mentioned in [another comment](https://github.com/fkie-cad/Logprep/pull/1018/changes#r4036535931) we generally try to avoid truthy/falsy checks. Also (minor thing) maybe call the variables `lhs_opt` and `rhs_opt` which makes this a bit shorter and not less readable in my opinion. But only a matter of personal preference

logprep/processor/calculator/ast/node.py
mhoff: Good catch!

logprep/processor/calculator/ast/node.py
```diff
@@ -0,0 +1,671 @@
...
+
+    def __init__(
+        self,
+        function_name: str,
+        function: Callable[..., Value],
+        *children: ASTNode,
+    ) -> None:
+        super().__init__(function_name, *children)
+        self.function = function
+        """The actual function to call"""
```
mhoff: Feel free to leave it as it is, but just 2 sidenotes:
1) Why not make these attributes private?
2) I cannot remember having seen these kind of attribute docs in the codebase. I think that's the case because then mostly properties were used, they were not documented at all or attrs fields were being employed. We just have to watch out to not have too many different styles of implementing certain things

logprep/processor/calculator/ast/parse.py
```diff
@@ -0,0 +1,276 @@
...
+    NegateASTNode,
+    RangeCheckASTNode,
+    VariableASTNode,
+)
+from logprep.processor.calculator.ast.util import read_hex_number
+from logprep.util.helper import VARIABLE_PATTERN
+
+# All __build_* functions below serve as callbacks for the  pyparsing based
+# syntax parser. They take a parsing result and convert it to an ASTNode.
+# Pyparse will wrap those nodes in an PraseResults instance before further
```
mhoff:
```suggestion
# Pyparse will wrap those nodes in an ParseResults instance before further
```

Not `ParserElement`?

Maybe rather move this docs to the module pydoc?

logprep/processor/calculator/ast/util.py
```diff
@@ -0,0 +1,72 @@
...
+
+class ValueType(Enum):
+    """A enum specifying the type used in an expression.
+    The AST emulates its own typing system through this enum.
+    """
+
+    BOOLEAN = auto()
+    """A boolean type"""
+    NUMBER = auto()
+    """A numeric type (integer or float)"""
```
mhoff, line 22:
```suggestion
    BOOLEAN = auto()
    """A boolean type"""

    NUMBER = auto()
    """A numeric type (integer or float)"""
```

logprep/processor/calculator/rule.py
```diff
@@ -50,18 +82,40 @@
...
+    * :code:`1 < 2 == 2`
+    * :code:`-(1 < 2)`
+    * :code:`(1 < 2) + 1`
+    * :code:`(1 < 2) == (2 < 3)`
+    * :code:`all(1, 1) * 2`
+
 
-Only one comparison operator is allowed per expression. Chained comparisons such as
-:code:`1 < 2 < 3` or :code:`1 < 2 == 2` are not supported.
+The following functions are available, where the function names a case-insensitive:
```
mhoff:
```suggestion
The following functions are available, where the function names are case-insensitive:
```

logprep/processor/calculator/rule.py
```diff
@@ -50,18 +82,40 @@
...
+
+Where :code:`a`, :code:`b` and :code:`c` are numeric values and the yielded
+result is a boolean.
+
+Arithmetic expression are evaluated before comparisons and range checks.
+
+.. warning::
+    Comparisons, range checks and some functions return boolean values.
+    Because unary minus, operators, range checks and most functions can only
+    accept numbers not booleans the following examples would result in
```
mhoff: Maybe?

```suggestion
    accept numbers not booleans, the following examples would result in
```

logprep/processor/calculator/rule.py
```diff
@@ -74,77 +128,35 @@
...
-The calc expression is not whitespace sensitive.
+.. datatemplate:import-module::
+   :source: tests.unit.processor.calculator.test_ast
+
+   {% for item in data.static_expression_example_test_cases %}
+   * :code:`{{item[0]}}` => :code:`{{item[1]}}`
+   {% endfor %}
+
+
+The calc expression is not whitespace or case sensitive.
```
mhoff: Above you wrote "case-insensitive"

So...

```suggestion
The calc expression is not whitespace or case-sensitive.
```

or the other way around?

logprep/processor/calculator/ast/function_registry.py
mhoff: Looks good, many thanks!

mhoff (commented):
Did a second iteration. Only very minor issues left

mhoff (dismissed):
Very good first PR. Thank you!

mhoff (approved):

PR #1019, opened 2026-08-26
fix: mixed-type scenarios in lucene range expressions

mhoff:
## Description

* filter: field values are now automatically coerced in range matches
* filter: allow mixed numeric range boundaries and type coercion for range matching
* filter: unquoted asterisks (`*`) in range expressions are now considered open boundaries
* filter: add support for `*` (open boundary) in range expressions
* filter: fix lucene range expressions not matching on mixed-type scenarios
* filter: treat `inf`/`nan` as string values instead of numeric range boundaries

## Assignee

### Documentation

### Code Quality

* pytest

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--1019.org.readthedocs.build/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 ### Breaking
 * restrict UNIX timestamp normalization to seconds, milliseconds, microseconds, and nanoseconds.
 * handle non-string `concatenator` source fields with `ProcessingWarning` instead of `ProcessingCriticalError`
+* filter: rework range expressions to allow mixed types and type coercion
+* filter: drop support for `inf`/`nan` in range expressions
+* filter: add support for `*` (open boundary) in range expressions
 
 ### Features
 * add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,8 @@
 ### Breaking
 * restrict UNIX timestamp normalization to seconds, milliseconds, microseconds, and nanoseconds.
 * handle non-string `concatenator` source fields with `ProcessingWarning` instead of `ProcessingCriticalError`
-* filter: rework range expressions to allow mixed types and type coercion
-* filter: drop support for `inf`/`nan` in range expressions
-* filter: add support for `*` (open boundary) in range expressions
+* filter: allow mixed numeric range boundaries and type coercion for range matching
+* filter: unquoted asterisks (`*`) in range expressions are now considered open boundaries
 
 ### Features
 * add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
@@ -12,6 +11,7 @@
 * generic_adder: add support for templated http urls & content_field
 * generic_resolver: add content_field support
 * field_name_replacer: add new `field_name_replacer` processor to replace occurences of strings in key names
+* filter: add support for `*` (open boundary) in range expressions
 
 ### Improvements
 * docs: enable pydoc placeholders for facilitating component reuse through inheritance
@@ -37,6 +37,7 @@
 * generic_adder: allow `None` as valid input via `add`
 * grokker: allow fallback matches without named fields
 * filter: fix lucene range expressions not matching on mixed-type scenarios
+* filter: treat `inf`/`nan` as string values instead of numeric range boundaries
 
 ## 20.0.0
 ### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@
 ### Breaking
 * restrict UNIX timestamp normalization to seconds, milliseconds, microseconds, and nanoseconds.
 * handle non-string `concatenator` source fields with `ProcessingWarning` instead of `ProcessingCriticalError`
-* filter: allow mixed numeric range boundaries and type coercion for range matching
-* filter: unquoted asterisks (`*`) in range expressions are now considered open boundaries
+* filter: field values are now automatically coerced in range matches
 
 ### Features
 * add support for fractional UNIX timestamps in the `timestamper` processor while preserving supported integer timestamp normalization.
@@ -12,6 +11,7 @@
 * generic_resolver: add content_field support
 * field_name_replacer: add new `field_name_replacer` processor to replace occurences of strings in key names
 * filter: add support for `*` (open boundary) in range expressions
+* filter: allow mixed numeric range boundaries and type coercion for range matching
 
 ### Improvements
 * docs: enable pydoc placeholders for facilitating component reuse through inheritance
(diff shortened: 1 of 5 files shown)


tests/unit/filter/test_filter_expression.py
```diff
@@ -241,45 +241,102 @@ def test_does_match_when_value_is_identical(self):
...
     def test_string_representation(self):
         assert str(self.filter) == "key1.key2:[23 TO 42]"
 
+    def test_string_representation_with_float_bounds(self):
+        numeric_range = NumericRangeFilterExpression(["key1", "key2"], 23.0, 42.0)
+        assert str(numeric_range) == "key1.key2:[23.0 TO 42.0]"
+
+    def test_string_representation_with_open_bound(self):
+        open_upper = NumericRangeFilterExpression(["key1", "key2"], 23, None)
+        assert str(open_upper) == "key1.key2:[23 TO *]"
```
krempel-4711: consider using mark.parametrize

logprep/filter/expression/filter_expression.py
```diff
@@ -297,83 +297,109 @@ class RangeBasedFilterExpression(KeyBasedFilterExpression, Generic[RangeBoundary
...
+    """Range based filter expression that matches numeric values."""
 
-    def does_match(self, document: dict) -> bool:
+    def does_match(self, document: dict[str, FieldValue]) -> bool:
         value = self._get_value(self.key, document)
 
-        if not isinstance(value, float) or isinstance(value, bool):
+        if isinstance(value, str):
+            try:
+                value = float(value)
```
kaya-david: This coercion is not lossless for large integers. For example, `float("9007199254740993")` becomes `9007199254740992.0`.

Could we preserve integer precision here, e.g. by trying `int()` before `float()`?

The following test reproduces the issue and currently fails with the current implementation:

```python
def test_does_match_when_value_is_large_numeric_string(self):
    value = 2**53 + 1  # 9007199254740993, float(value) becomes 9007199254740992.0
    numeric_range = NumericRangeFilterExpression(["key1", "key2"], value, value)

    assert numeric_range.matches({"key1": {"key2": str(value)}})
```

logprep/filter/lucene_filter.py
```diff
@@ -403,122 +457,83 @@ def _parse_tree(self, tree: luqum.tree) -> FilterExpression:
...
+            parsed_value = int(value)
         except ValueError:
-            return None
+            try:
+                parsed_value = float(value)
+            except ValueError as exc:
+                raise ValueError(f"failed to parse '{value}' as int/float") from exc
 
-        LuceneTransformer._validate_range_boundaries(lower_bound, upper_bound, expr)
+        if not math.isfinite(parsed_value):
```
kaya-david: `math.isfinite()` can raise an `OverflowError` for valid Python integers that are too large to be represented as a float.

This is a regression from the previous integer range handling, where arbitrary-size integers were parsed with `int()` and compared without converting them to floats.

The following test reproduces the issue and fails with the current implementation:

```python
def test_created_numeric_range_supports_integer_beyond_float_range(self, range_query):
    value = 10**400  # valid Python int, but too large to convert to float

    lucene_filter = LuceneFilter.create(range_query(f"[{value} TO {value}]"))

    assert lucene_filter.matches({"key": value})
```

It currently fails with `OverflowError: int too large to convert to float`.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 ### Breaking
 * restrict UNIX timestamp normalization to seconds, milliseconds, microseconds, and nanoseconds.
 * handle non-string `concatenator` source fields with `ProcessingWarning` instead of `ProcessingCriticalError`
+* filter: rework range expressions to allow mixed types and type coercion
+* filter: drop support for `inf`/`nan` in range expressions
```
kaya-david:
```suggestion
* filter: treat `inf`/`nan` as string values instead of numeric range boundaries
```

and I'm not sure whether `inf`/`nan` should be listed under **Breaking** at all.

Previously, these expressions were explicitly rejected as invalid with `Range boundaries must be finite numbers`. With this change, they are accepted and fall back to lexicographic string comparison instead.

So the current changelog entry `drop support for inf/nan` seems misleading. Support isn't being dropped; previously invalid expressions are becoming valid with different semantics.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 ### Breaking
 * restrict UNIX timestamp normalization to seconds, milliseconds, microseconds, and nanoseconds.
 * handle non-string `concatenator` source fields with `ProcessingWarning` instead of `ProcessingCriticalError`
+* filter: rework range expressions to allow mixed types and type coercion
```
kaya-david:
```suggestion
* filter: allow mixed numeric range boundaries and type coercion for range matching
```

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
...
+* filter: drop support for `inf`/`nan` in range expressions
+* filter: add support for `*` (open boundary) in range expressions
```
kaya-david: Breaking or Feature?

logprep/filter/lucene_filter.py
```diff
@@ -65,11 +64,23 @@
...
-The lower and upper boundaries of a range must have the same type. Mixed ranges
-such as integer-to-float, integer-to-string, or float-to-string are not
-supported.
+Numeric range boundaries may freely mix integers and floats, and match integer,
+floating-point, and numeric-looking string values in documents (e.g. a field
+holding :code:`"24"` matches :code:`[18 TO 65]`).
+A range mixing a numeric and a non-numeric boundary, such as :code:`[1 TO bar]`,
+falls back to a lexicographic string range instead.
+
+String range matches integer and floating-point field values too,
```
kaya-david:
```suggestion
String ranges match integer and floating-point field values too,
```

kaya-david (commented):
Thanks, left some comments

CHANGELOG.md
mhoff: Fair enough. So it's rather a bugfix as these are implementation details which leaked the abstraction before

CHANGELOG.md
```diff
@@ -2,13 +2,16 @@
 ### Breaking
 * restrict UNIX timestamp normalization to seconds, milliseconds, microseconds, and nanoseconds.
 * handle non-string `concatenator` source fields with `ProcessingWarning` instead of `ProcessingCriticalError`
+* filter: allow mixed numeric range boundaries and type coercion for range matching
+* filter: unquoted asterisks (`*`) in range expressions are now considered open boundaries
```
kaya-david: `* filter: unquoted asterisks (`*`) in range expressions are now considered open boundaries`

I'don't think this belongs under **Breaking** and can be removed. Previously, open `*` boundaries were rejected, while this change adds support for them. That seems like a new feature rather than a breaking change, and it is already covered under **Features** below.

kaya-david (commented):
Thanks. Just one small point left

kaya-david (approved):
Thanks. LGTM

PR #1021, opened 2026-08-31
feat: add domain resolver timeout cache, prune domain map and rework cache

ppcad:
## Description

* add separate cache for timeouts in `domain_resolver` to allow different lifetimes than the regular domain cache
* add `lifetime` parameter to `domain_resolver` in addition to `timeout` parameter
* add metrics for cached timeouts and successfully resolved domains to `domain_resolver`
* prune domains from `domain_resolver` mapping that are not in the cache
* prune domains from `domain_resolver` mapping that are not in the cache 

## Assignee

### Documentation

### Code Quality

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--1021.org.readthedocs.build/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@
 ### Improvements
 * add a separate cache for timeouts in `domain_resolver` to allow different lifetimes than the regular domain cache
 * add `lifetime` parameter to `domain_resolver` in addition to `timeout` parameter
-* add metrics for cached timeouts and successfully resolved domains to `domain_resolver`
+* add metrics for cached timed-out domains and successfully resolved domains to `domain_resolver`
 * prune domains from `domain_resolver` mapping that are not in the cache
 * docs: enable pydoc placeholders for facilitating component reuse through inheritance
 * docs: change processor natural naming to capital cased with whitespace (e.g. "Generic Resolver")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@
 * grokker: improve performance by stopping on first matching expression
 
 ### Bugfix
-* Prune domains from `domain_resolver` mapping that are not in the cache 
+* prune domains from `domain_resolver` mapping that are not in the cache
 * chart: fix command handling
 * ng: fix input timeout to also accept int parameters
 * ng: fix `http_input` `collect_meta` leading to shared dicts between events
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@
 * filter: allow mixed numeric range boundaries and type coercion for range matching
 
 ### Improvements
+* add separate cache for timeouts in `domain_resolver` to allow different lifetimes than the regular domain cache
+* add `lifetime` parameter to `domain_resolver` in addition to `timeout` parameter
+* add metrics for cached timeouts and successfully resolved domains to `domain_resolver`
+* prune domains from `domain_resolver` mapping that are not in the cache
 * docs: enable pydoc placeholders for facilitating component reuse through inheritance
 * docs: change processor natural naming to capital cased with whitespace (e.g. "Generic Resolver")
 * docs: use processor name placeholders for most usages
@@ -41,6 +45,7 @@
 * grokker: improve performance by stopping on first matching expression
 
 ### Bugfix
+* Prune domains from `domain_resolver` mapping that are not in the cache 
 * chart: fix command handling
 * ng: fix input timeout to also accept int parameters
 * ng: fix `http_input` `collect_meta` leading to shared dicts between events
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
 * filter: allow mixed numeric range boundaries and type coercion for range matching
 
 ### Improvements
-* add separate cache for timeouts in `domain_resolver` to allow different lifetimes than the regular domain cache
+* add a separate cache for timeouts in `domain_resolver` to allow different lifetimes than the regular domain cache
 * add `lifetime` parameter to `domain_resolver` in addition to `timeout` parameter
 * add metrics for cached timeouts and successfully resolved domains to `domain_resolver`
 * prune domains from `domain_resolver` mapping that are not in the cache
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,15 @@
 ## Upcoming Changes
 ### Breaking
+* add `lifetime` parameter to `domain_resolver` in addition to `timeout` parameter
+* add a separate cache for timeouts in `domain_resolver` to allow different lifetimes than the regular domain cache using the `timeout_block_time` parameter
 
 ### Features
 * timestamper: support generating timestamps from the current time when no `source_fields` are configured
 
 ### Improvements
 * timestamper: validate `source_format` and `source_timezone` according to the configured `source_fields`
+* add metrics for cached timed-out domains and successfully resolved domains to `domain_resolver`
+* make cache hold a payload in addition to timestamps and adapt `domain_resolver` accordingly
 
 ### Bugfix
 
@@ -25,10 +29,6 @@
 * filter: allow mixed numeric range boundaries and type coercion for range matching
 
 ### Improvements
-* add a separate cache for timeouts in `domain_resolver` to allow different lifetimes than the regular domain cache
-* add `lifetime` parameter to `domain_resolver` in addition to `timeout` parameter
-* add metrics for cached timed-out domains and successfully resolved domains to `domain_resolver`
-* prune domains from `domain_resolver` mapping that are not in the cache
 * docs: enable pydoc placeholders for facilitating component reuse through inheritance
 * docs: change processor natural naming to capital cased with whitespace (e.g. "Generic Resolver")
 * docs: use processor name placeholders for most usages
@@ -45,7 +45,6 @@
 * grokker: improve performance by stopping on first matching expression
 
 ### Bugfix
-* prune domains from `domain_resolver` mapping that are not in the cache
 * chart: fix command handling
 * ng: fix input timeout to also accept int parameters
 * ng: fix `http_input` `collect_meta` leading to shared dicts between events
--- a/logprep/ng/processor/domain_resolver/processor.py
+++ b/logprep/ng/processor/domain_resolver/processor.py
@@ -97,9 +97,7 @@ class DomainResolver(Processor):
            Ensure to set this to a reasonable value to avoid DOS attacks by malicious domains in
            your logs. The default is set to 0.5 seconds.
         """
-        lifetime: float = field(
-            default=1.0, validator=validators.optional(validators.instance_of(float))
-        )
+        lifetime: float = field(default=1.0, validator=validators.instance_of(float))
         """Total timeout for resolving of domains including multiple attempts."""
         max_cached_domains: int = field(validator=validators.instance_of(int))
         """The maximum number of cached domains. One cache entry requires ~250 Byte, thus 10
@@ -194,7 +192,7 @@ class DomainResolver(Processor):
 
     __slots__ = ["_domain_ip_map"]
 
-    _domain_ip_map: dict[str, Optional[SuccessResult | FailedResult]]
+    _domain_ip_map: dict[str, SuccessResult | FailedResult]
 
     rule_class = DomainResolverRule
 
@@ -251,7 +249,7 @@ class DomainResolver(Processor):
         if not domain_or_url_str:
             return
 
-        url = urlsplit(domain_or_url_str)
+        url = urlsplit(str(domain_or_url_str))
         domain = url.hostname
         if url.scheme == "":
             domain = url.path
--- a/logprep/ng/processor/domain_resolver/processor.py
+++ b/logprep/ng/processor/domain_resolver/processor.py
@@ -14,11 +14,11 @@ Processor Configuration
         rules:
             - tests/testdata/rules/rules
         timeout: 0.5
+        lifetime: 1.0
         max_cached_domains: 20000
         max_caching_days: 1
         hash_salt: secure_salt
         cache_enabled: true
-        debug_cache: false
 
 .. autoclass:: logprep.processor.domain_resolver.processor.DomainResolver.Config
    :members:
@@ -31,38 +31,50 @@ Processor Configuration
 
 import datetime
 import logging
-import socket
 import typing
 from enum import IntEnum
 from functools import cached_property
-from multiprocessing import context
-from multiprocessing.pool import ThreadPool
+from typing import Optional
 from urllib.parse import urlsplit
 
-from attrs import define, field, validators
+from attr import define, field, validators
+from dns.exception import Timeout, FormError, SyntaxError as DNSSyntaxError, TooBig
+from dns.resolver import Resolver, NXDOMAIN, LifetimeTimeout, NoAnswer, NoNameservers
 
+from logprep.abc.processor import Processor
 from logprep.metrics.metrics import CounterMetric
-from logprep.ng.abc.processor import Processor
-from logprep.processor.base.rule import Rule
 from logprep.processor.domain_resolver.rule import DomainResolverRule
-from logprep.util.cache import Cache
+from logprep.util.cache import Cache, Timer
 from logprep.util.hasher import SHA256Hasher
-from logprep.util.helper import FieldValue, add_fields_to, get_dotted_field_value
+from logprep.util.helper import get_dotted_field_value
 
 logger = logging.getLogger("DomainResolver")
 
 
-class ResolveStatus(IntEnum):
+class FailureType(IntEnum):
     """Status of resolving domains"""
 
-    SUCCESS = 0
-    """Resolving the domain was successful"""
-    TIMEOUT = 1
+    TIMEOUT = 0
     """Domain resolver timeout while trying to resolve the domain (this is not a socket timeout)"""
-    INVALID = 2
+    INVALID = 1
     """The resolved domain was invalid and thus not resolved"""
-    UNKNOWN = 3
+    UNKNOWN = 2
     """Tried to resolve the domain, but the domain is unknown"""
+    NO_ANSWER = 3
+    """The resolved domain was valid, but returned no data"""
+    NO_NAMESERVERS = 4
+    """Nameservers do not exist or timed out"""
+
+
+@define
+class SuccessResult:
+    resolved_ip: str
+
+
+@define
+class FailedResult:
+    failure_type: FailureType
+    error: Exception | None = None
 
 
 class DomainResolver(Processor):
@@ -70,24 +82,47 @@ class DomainResolver(Processor):
 
     @define(kw_only=True)
     class Config(Processor.Config):
-        """DomainResolver config"""
+        """Config for |PROCESSOR|"""
 
         timeout: float = field(
             default=0.5,
             validator=validators.instance_of(float),
             converter=float,
         )
-        """Timeout for resolving of domains."""
+        """Timeout for resolving of domains.
+
+        .. security-best-practice::
+           :title: |PROCESSOR| - Timeout
+
+           Ensure to set this to a reasonable value to avoid DOS attacks by malicious domains in
+           your logs. The default is set to 0.5 seconds.
+        """
+        lifetime: float = field(
+            default=1.0, validator=validators.optional(validators.instance_of(float))
+        )
+        """Total timeout for resolving of domains including multiple attempts."""
         max_cached_domains: int = field(validator=validators.instance_of(int))
         """The maximum number of cached domains. One cache entry requires ~250 Byte, thus 10
         million elements would require about 2.3 GB RAM. The cache is not persisted. Restarting
-        Logprep does therefore clear the cache."""
+        Logprep does therefore clear the cache.
+
+        .. security-best-practice::
+           :title: |PROCESSOR| - Max Cached Domains
+
+           Ensure to set this to a reasonable value to avoid excessive memory usage
+           and OOM situations by the domain resolver cache.
+
+        """
+        timeout_block_time: float = field(default=5.0, validator=validators.instance_of(float))
+        """Minutes after which a timed out domain can be resolved again."""
+        cache_prune_interval: float = field(default=3600.0, validator=validators.instance_of(float))
+        """Seconds after which decayed domains are pruned from caches."""
         max_caching_days: int = field(validator=validators.instance_of(int))
         """Number of days a domains is cached after the last time it appeared.
         This caching reduces the CPU load of Logprep (no demanding encryption must be performed
         repeatedly) and the load on subsequent components (i.e. Logstash or Opensearch).
         Setting the caching days to Null deactivates the caching. In case the cache size has been
-        exceeded (see `domain_resolver.max_cached_domains`),the oldest cached pseudonyms will
+        exceeded (see `domain_resolver.max_cached_domains`),the oldest cached resolved domains will
         be discarded first.Thus, it is possible that a domain is re-added to the cache before
         max_caching_days has elapsed if it was discarded due to the size limit."""
         hash_salt: str = field(validator=validators.instance_of(str))
@@ -95,9 +130,6 @@ class DomainResolver(Processor):
         cache_enabled: bool = field(default=True, validator=validators.instance_of(bool))
         """If enabled activates a cache such that already seen domains do not need to be resolved
         again."""
-        debug_cache: bool = field(default=False, validator=validators.instance_of(bool))
-        """If enabled adds debug information to the current event, for example if the event
-        was retrieved from the cache or newly resolved, as well as the cache size."""
 
     @define(kw_only=True)
     class Metrics(Processor.Metrics):
@@ -124,6 +156,13 @@ class DomainResolver(Processor):
             )
         )
         """Number of urls that were resolved from cache"""
+        resolved_domains: CounterMetric = field(
+            factory=lambda: CounterMetric(
+                description="Number of domains that were successfully resolved",
+                name="domain_resolver_resolved_domains",
+            )
+        )
+        """Number of domains that were successfully resolved"""
         timeouts: CounterMetric = field(
             factory=lambda: CounterMetric(
                 description="Number of timeouts that occurred while resolving a url",
@@ -131,6 +170,13 @@ class DomainResolver(Processor):
             )
         )
         """Number of timeouts that occurred while resolving a url"""
+        timeouts_cached: CounterMetric = field(
+            factory=lambda: CounterMetric(
+                description="Number of timeouts from the timeout cache for a url",
+                name="domain_resolver_timeouts_cached",
+            )
+        )
+        """Number of timeouts from the timeout cache for a url"""
         invalid_domains: CounterMetric = field(
             factory=lambda: CounterMetric(
                 description="Number of invalid domains",
@@ -148,11 +194,11 @@ class DomainResolver(Processor):
 
     __slots__ = ["_domain_ip_map"]
 
-    _domain_ip_map: dict[str, str | None]
+    _domain_ip_map: dict[str, Optional[SuccessResult | FailedResult]]
 
     rule_class = DomainResolverRule
 
-    def __init__(self, name: str, configuration: Processor.Config) -> None:
+    def __init__(self, name: str, configuration: Processor.Config):
         super().__init__(name, configuration)
         self._domain_ip_map = {}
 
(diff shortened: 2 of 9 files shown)


logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -31,73 +31,102 @@
...
     """Domain resolver timeout while trying to resolve the domain (this is not a socket timeout)"""
-    INVALID = 2
+    INVALID = 1
     """The resolved domain was invalid and thus not resolved"""
-    UNKNOWN = 3
+    UNKNOWN = 2
     """Tried to resolve the domain, but the domain is unknown"""
+    NO_ANSWER = 3
+    """The resolved domain was valid, but returned no data"""
+    NO_NAMESERVERS = 4
```
mhoff: Use `= auto()` instead of static numbers?

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -31,73 +31,102 @@
...
+        .. security-best-practice::
+           :title: |PROCESSOR| - Max Cached Domains
+
+           Ensure to set this to a reasonable value to avoid excessive memory usage
+           and OOM situations by the domain resolver cache.
+
+        """
+        timeout_block_time: float = field(default=5.0, validator=validators.instance_of(float))
+        """Minutes after which a timed out domain can be resolved again."""
+        cache_prune_interval: float = field(default=3600.0, validator=validators.instance_of(float))
```
mhoff: Having a line break makes it a bit easier to read, because there is no ambiguity whether the text belongs to the entry above or below

```suggestion
        """Minutes after which a timed out domain can be resolved again."""

        cache_prune_interval: float = field(default=3600.0, validator=validators.instance_of(float))
```

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -148,11 +191,11 @@ class Metrics(Processor.Metrics):
 
     __slots__ = ["_domain_ip_map"]
 
-    _domain_ip_map: dict[str, str | None]
+    _domain_ip_map: dict[str, SuccessResult | FailedResult]
 
     rule_class = DomainResolverRule
 
-    def __init__(self, name: str, configuration: Processor.Config) -> None:
```
mhoff, line 155: Why remove the type annotation? I think this way the body is not checked properly

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -162,88 +205,133 @@ def config(self) -> Config:
...
 
     @cached_property
-    def _cache(self) -> Cache:
+    def _dns_resolver(self) -> Resolver:
+        dns_resolver = Resolver()
+        dns_resolver.timeout = self.config.timeout
+        dns_resolver.lifetime = self.config.lifetime
+        return dns_resolver
+
+    @cached_property
```
mhoff: Rather use `setup()` to instantiate the properties. This makes the attributes of an object and what exactly happens on init/setup way clearer.

On a related note: aren't we missing `__dict__` in `__slots__` for `cached_property` to work properly?

logprep/util/cache.py
```diff
@@ -23,13 +47,52 @@ def requires_storing(self, item: int | str) -> bool:
...
+        last_stored = self.get(item)
+        if last_stored is None:
+            return False
+
+        if datetime.datetime.now() - last_stored > self._max_timedelta:
+            self.pop(item)
+            return False
+        return True
+
+    def add(self, item: Union[int, str]):
```
mhoff:
```suggestion
    def add(self, item: int | str):
```

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -162,88 +205,133 @@ def config(self) -> Config:
...
-    def _thread_pool(self) -> ThreadPool:
-        return ThreadPool(processes=1)
+    def _hasher(self) -> SHA256Hasher:
+        return SHA256Hasher()
 
-    async def _apply_rules(self, event: dict[str, FieldValue], rule: Rule) -> None:
-        rule = typing.cast(DomainResolverRule, rule)
+    def _apply_rules(self, event: dict[str, typing.Any], rule: DomainResolverRule):
+        self._timeout_cache.prune_decayed()
+        self._domain_cache.prune_decayed()
```
mhoff: Why not use `Processor._schedule_task` to register the cache prune? It seems expensive to me to do this check three times for each event. Assuming 1000 EPS, that is once every 1ms.

Sidenote: `Component.run_pending_tasks` is currently missing in the ng-Architecture. Would need to be added here (https://github.com/fkie-cad/Logprep/blob/main/logprep/ng/runner.py#L52)

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -162,88 +205,133 @@ def config(self) -> Config:
...
+        self._prune_domain_ip_map()
         source_field = rule.source_fields[0]
         domain_or_url_str = get_dotted_field_value(event, source_field)
         if not domain_or_url_str:
             return
-        if not isinstance(domain_or_url_str, str):
-            raise ValueError("expected source_field to be a string")
 
-        url = urlsplit(domain_or_url_str)
+        url = urlsplit(str(domain_or_url_str))
```
mhoff: We are silently converting `dict`, `list`, `None`, `bool`, `int`, `float` to a `str` here which never seems to make sense

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -162,88 +205,133 @@ def config(self) -> Config:
...
+
+        result = self._resolve_ip(domain)
+        if isinstance(result, FailedResult) and self._is_timeout(result):
+            self._timeout_cache.add(hash_string)
+        return result
 
-        if self.config.debug_cache:
-            self._store_debug_infos(event, requires_storing)
+    def _resolve_with_cache(self, domain: str) -> SuccessResult | FailedResult:
+        hash_string = self._hasher.hash_str(domain, salt=self.config.hash_salt)
```
mhoff, line 270: We are calculating the same hash twice. Once in `_resolve_with_timeout_check` and once here

logprep/util/cache.py
```diff
@@ -1,20 +1,44 @@
...
+    def __init__(
+        self, max_items=1000000, max_timedelta=datetime.timedelta(days=90.0), prune_interval=5
+    ):
         self._max_items = max_items
         self._max_timedelta = max_timedelta
+        self._prune_timer = Timer(prune_interval)
         super().__init__()
 
-    def requires_storing(self, item: int | str) -> bool:
+    def is_cached(self, item: Union[int, str]) -> bool:
```
mhoff:
```suggestion
    def is_cached(self, item: int | str) -> bool:
```

logprep/util/cache.py
```diff
@@ -23,13 +47,52 @@ def requires_storing(self, item: int | str) -> bool:
...
-        self.move_to_end(item)
-        if self._max_timedelta.total_seconds() == 0.0 or now - last_stored > self._max_timedelta:
-            self[item] = now
-            if len(self) > self._max_items:
-                self.popitem(last=False)
+        last_stored = self.get(item)
+        if last_stored is None:
+            return False
+
+        if datetime.datetime.now() - last_stored > self._max_timedelta:
```
mhoff: Is `datetime.datetime.now()` the quickest way to get a monotonic counter? I think we are using other methods as well. `Timer` uses `time.time()`

logprep/util/cache.py
```diff
@@ -23,13 +47,52 @@ def requires_storing(self, item: int | str) -> bool:
...
+            Item whose timestamp to update in the cache.
+
+        """
+        last_stored = self.get(item)
+        if last_stored is not None:
+            self[item] = datetime.datetime.now()
             return True
         return False
+
+    def prune_decayed(self):
```
mhoff: We have two pruning approaches. One per item (`is_cached`) and one here. Would it be viable to rely on only one instead?

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -162,88 +205,133 @@ def config(self) -> Config:
...
+            self._domain_cache.update_cache(hash_string)
+            result = self._domain_ip_map[hash_string]
+            self.metrics.resolved_cached += 1
+            return result
+
+        result = self._resolve_with_timeout_check(domain)
+        match result:
+            case SuccessResult(_):
+                self._domain_cache.add(hash_string)
+                self._domain_ip_map.update({hash_string: result})
```
mhoff: Shouldn't this rather be a combined structure? Like a subclass of `Cache` which holds tuples of timestamp and payload? Also isn't there a standard implementation for this. It surprises me that we need to do these things ourselves

CHANGELOG.md
```diff
@@ -26,6 +30,7 @@
 * ci: introduce umbrella job for enforcing merge status checks with GitHub
 
 ### Bugfix
+* Prune domains from `domain_resolver` mapping that are not in the cache
```
```suggestion
* prune domains from `domain_resolver` mapping that are not in the cache 
```

pyproject.toml
```diff
@@ -70,6 +70,7 @@ dependencies = [
   "uvloop<1",
   "httptools<1",
   "more-itertools>=11.1.0",
+  "dnspython==2.8.0",
```
mhoff:
```suggestion
  "dnspython>=2.8.0",
```

mhoff (commented):
Hello @ppcad, many thanks for the work. I did a first review pass and added a few remarks

logprep/ng/processor/domain_resolver/processor.py
ppcad: I've added newlines between the parameters.

logprep/util/cache.py
ppcad: I'm using time() for performance now. timedelta() is still used for large numbers for better readability, since it is only used in tests and on initialization.

logprep/ng/processor/domain_resolver/processor.py
ppcad: I've refactored it.

logprep/ng/processor/domain_resolver/processor.py
ppcad: I've changed it.

logprep/ng/processor/domain_resolver/processor.py
ppcad: I have made cache entries contain a timestamp and a payload to remove `_domain_up_map`. We do not have to keep it in sync now. We have our own implementation, since the standard cache has no TTL feature.

logprep/util/cache.py
ppcad: I removed pruning. It might leave the cache larger, but it makes it easier to maintain.

logprep/ng/processor/domain_resolver/processor.py
ppcad: I did not use it at first, since it had a bug in my Logprep version, which seems to be fixed now. It used it's own timer before to prevent it from being to costly. However, I ended up not using pruning at all now, so it's no longer needed.

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -29,75 +29,112 @@
...
         million elements would require about 2.3 GB RAM. The cache is not persisted. Restarting
-        Logprep does therefore clear the cache."""
+        Logprep does therefore clear the cache.
+
+        .. security-best-practice::
+           :title: |PROCESSOR| - Max Cached Domains
+
+           Ensure to set this to a reasonable value to avoid excessive memory usage
+           and OOM situations by the domain resolver cache.
+
```
mhoff, line 121:
```suggestion
```

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -29,75 +29,112 @@
...
-    INVALID = 2
+    INVALID = auto()
     """The resolved domain was invalid and thus not resolved"""
-    UNKNOWN = 3
+    UNKNOWN = auto()
     """Tried to resolve the domain, but the domain is unknown"""
+    NO_ANSWER = auto()
+    """The resolved domain was valid, but returned no data"""
+    NO_NAMESERVERS = auto()
+    """Nameservers do not exist or timed out"""
```
mhoff, line 69:
```suggestion
    NO_ANSWER = auto()
    """The resolved domain was valid, but returned no data"""

    NO_NAMESERVERS = auto()
    """Nameservers do not exist or timed out"""
```

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
 ## Upcoming Changes
 ### Breaking
+* add `lifetime` parameter to `domain_resolver` in addition to `timeout` parameter
+* add a separate cache for timeouts in `domain_resolver` to allow different lifetimes than the regular domain cache using the `timeout_block_time` parameter
```
mhoff: remove `debug_cache`

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -124,13 +161,27 @@ class Metrics(Processor.Metrics):
...
         )
         """Number of urls that were resolved from cache"""
+        resolved_domains: CounterMetric = field(
+            factory=lambda: CounterMetric(
+                description="Number of domains that were successfully resolved",
+                name="domain_resolver_resolved_domains",
+            )
+        )
+        """Number of domains that were successfully resolved"""
         timeouts: CounterMetric = field(
```
mhoff, line 179:
```suggestion
        """Number of domains that were successfully resolved"""

timeouts: CounterMetric = field(
```

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -124,13 +161,27 @@ class Metrics(Processor.Metrics):
...
             factory=lambda: CounterMetric(
                 description="Number of timeouts that occurred while resolving a url",
                 name="domain_resolver_timeouts",
             )
         )
         """Number of timeouts that occurred while resolving a url"""
+        timeouts_cached: CounterMetric = field(
```
mhoff, line 187:
```suggestion
"""Number of timeouts that occurred while resolving a url"""

        timeouts_cached: CounterMetric = field(
```

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -146,104 +197,119 @@ class Metrics(Processor.Metrics):
...
 
     rule_class = DomainResolverRule
 
-    def __init__(self, name: str, configuration: Processor.Config) -> None:
-        super().__init__(name, configuration)
-        self._domain_ip_map = {}
-
     @property
     def config(self) -> Config:
         """Provides the properly typed rule configuration object"""
```
mhoff:
```suggestion
        """Provides the properly typed configuration object"""
```

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -146,104 +197,119 @@ class Metrics(Processor.Metrics):
...
 
-    @cached_property
-    def _cache(self) -> Cache:
-        cache_max_timedelta = datetime.timedelta(days=self.config.max_caching_days)
-        return Cache(max_items=self.config.max_cached_domains, max_timedelta=cache_max_timedelta)
+    async def setup(self):
+        await super().setup()
+        self._dns_resolver = Resolver()
+        self._dns_resolver.timeout = self.config.timeout
+        self._dns_resolver.lifetime = self.config.lifetime
```
mhoff, line 234: It was surprising to me that you need to overwrite attributes after init. Others might have the same question. I think a comment makes sense here

```suggestion
        self._dns_resolver = Resolver()
        # attributes are meant to be overwritten after init
        self._dns_resolver.timeout = self.config.timeout
        self._dns_resolver.lifetime = self.config.lifetime
```

logprep/util/cache.py
```diff
@@ -1,35 +1,87 @@
...
-import datetime
+import time
+from dataclasses import dataclass
+from datetime import timedelta
+
 from collections import OrderedDict
+from typing import Any
+
+
+@dataclass
```
mhoff, line 11: Rather a slotted attrs class?

logprep/util/cache.py
```diff
@@ -1,35 +1,87 @@
...
+class CacheEntry:
```
mhoff, line 12: Rather a private class I suppose

```suggestion
class _CacheEntry:
```

logprep/util/cache.py
```diff
@@ -1,35 +1,87 @@
...
 class Cache(OrderedDict):
     """Caches items along with a timestamp of when they were last stored."""
 
-    EPOCH = datetime.datetime(1970, 1, 1)
-
-    def __init__(self, max_items=1000000, max_timedelta=datetime.timedelta(days=90.0)):
+    def __init__(
+        self,
+        max_items=1000000,
+        max_timedelta=timedelta(days=90).total_seconds(),
```
mhoff, line 25: The type of init param changes here. Please use type annotations and are other users of `max_timedelta` affected by this breaking change?

logprep/util/cache.py
```diff
@@ -1,35 +1,87 @@
...
         super().__init__()
 
-    def requires_storing(self, item: int | str) -> bool:
-        """Check if the item was stored within the last timedelta.
+    def __getitem__(self, key) -> CacheEntry:
+        return super().__getitem__(key)
+
+    def get(self, *args, **kwargs) -> CacheEntry | None:
+        """Get a cached item with the type CacheEntry."""
+        return super().get(*args, **kwargs)
```
mhoff, line 36: Do we need this as a public method? Should a using component care about our cache entry implementation? I think we should have a method which returns the value inside `CacheEntry`

logprep/util/cache.py
```diff
@@ -1,35 +1,87 @@
...
+
+    def is_cached(self, key: str) -> bool:
+        """Check if the item has exceeded its time to live.
 
         Parameters
         ----------
-        item : str
+        key : str
             Name of item to check for in the cache.
```
mhoff:
```suggestion
```

logprep/util/cache.py
```diff
@@ -1,35 +1,87 @@
...
+    def add(self, key: str, value: Any = None):
+        """Add the item into the cache or refresh its time to live.
+
+        Parameters
+        ----------
+        key : str
+            Key for item to add into the cache.
+        value : Any
+            Value of item to add into the cache.
+
```
mhoff:
```suggestion
```

logprep/util/cache.py
```diff
@@ -1,35 +1,87 @@
...
 from collections import OrderedDict
+from typing import Any
+
+
+@dataclass
+class CacheEntry:
+    """Entry for cache that consists of value and insertion time"""
+
+    value: Any
```
mhoff, line 15: We should use a generic type parameter to avoid `Any`

logprep/util/cache.py
```diff
@@ -1,35 +1,87 @@
...
+        last_stored = self.get(key)
+        if last_stored is None:
+            return False
+
+        if time.time() - last_stored.insertion_time > self._max_timedelta:
+            self.pop(key)
+            return False
+        return True
+
+    def add(self, key: str, value: Any = None):
```
mhoff, line 49: I understand that you want the `value = None` scenario for timeout caching. However, that's a bit surprising for a general cache interface. Maybe at least remove the default parameter

logprep/util/cache.py
```diff
@@ -1,35 +1,87 @@
...
+            self.popitem(last=False)
+
+    def refresh_time_to_live(self, key) -> bool:
+        """Update the items timestamp inside the cache.
+
+        Parameters
+        ----------
+        key : str
+            Item whose timestamp to update in the cache.
+
```
mhoff:
```suggestion
```

logprep/util/cache.py
```diff
@@ -1,35 +1,87 @@
...
+        value : Any
+            Value of item to add into the cache.
+
+        """
+        if self.refresh_time_to_live(key):
+            return
+
+        self[key] = CacheEntry(value, time.time())
+        if len(self) > self._max_items:
+            self.popitem(last=False)
```
mhoff, line 56: The LRU example for `OrderedDict` calls `move_to_end` when retrieving an entry: https://docs.python.org/3/library/collections.html#ordereddict-examples-and-recipes

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -146,104 +197,119 @@ class Metrics(Processor.Metrics):
...
+        )
 
-    @cached_property
-    def _hasher(self) -> SHA256Hasher:
-        return SHA256Hasher()
+        cache_max_timedelta = timedelta(days=self.config.max_caching_days).total_seconds()
+        self._domain_cache = Cache(
+            max_items=self.config.max_cached_domains,
+            max_timedelta=cache_max_timedelta,
+        )
```
mhoff, line 246:
```suggestion
        self._timeout_cache = Cache(
            max_items=self.config.max_cached_domains,
            max_timedelta=timedelta(minutes=self.config.timeout_block_time).total_seconds(),
        )

        self._domain_cache = Cache(
            max_items=self.config.max_cached_domains,
            max_timedelta=timedelta(days=self.config.max_caching_days).total_seconds(),
        )
```

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -146,104 +197,119 @@ class Metrics(Processor.Metrics):
...
-            self.metrics.resolved_new.inc(1)
-        else:
-            resolved_ip = self._domain_ip_map.get(hash_string)
-            self.metrics.resolved_cached.inc(1)
-        self._add_resolve_infos_to_event(event, rule, resolved_ip)
+        match result:
+            case SuccessResult(resolved_ip) if resolved_ip:
+                self._add_resolve_infos_to_event(event, rule, resolved_ip)
+            case FailedResult(_, error) if error:
+                self._handle_warning_error(event, rule, error)
```
mhoff, line 278: What happens in either scenario if the if condition does not hold true. Well, nothing obviously, but also no log or no incremented metric. Maybe we should emit an info here

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -146,104 +197,119 @@ class Metrics(Processor.Metrics):
...
+            return self._domain_cache[hash_string].value
+
+        if self._timeout_cache.is_cached(hash_string):
+            self.metrics.timeouts_cached.inc(1)
+            return FailedResult(FailureType.TIMEOUT)
+
+        result = self._resolve_ip(domain)
+        match result:
+            case FailedResult(FailureType.TIMEOUT | FailureType.NO_NAMESERVERS):
+                self._timeout_cache.add(hash_string)
```
mhoff, line 295: What do we do on `FailureType.UNKNOWN` or maybe other additional failure types?

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -146,104 +197,119 @@ class Metrics(Processor.Metrics):
...
+            case _:
+                self._domain_cache.add(hash_string, result)
+        self.metrics.resolved_new.inc(1)
```
mhoff, line 298: This should not happen for the failure path

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -146,104 +197,119 @@ class Metrics(Processor.Metrics):
...
 
         Assumes socket default timeout is None and relies on threading to create a timeout.
         """
         try:
-            result = self._thread_pool.apply_async(socket.gethostbyname, (domain,))
-            resolved_ip = result.get(timeout=self.config.timeout)
-            return resolved_ip, ResolveStatus.SUCCESS
-        except ValueError:  # Makes no connection so does not need to be cached
+            result = self._dns_resolver.resolve(domain, "A")
+            self.metrics.resolved_domains.inc(1)
```
mhoff, line 312: What is the difference between `resolved_new` and `resolved_domains`?

logprep/ng/processor/domain_resolver/processor.py
```diff
@@ -146,104 +197,119 @@ class Metrics(Processor.Metrics):
...
-        event_dbg = {
-            "resolved_ip_debug": {
-                "obtained_from_cache": not requires_storing,
-                "cache_size": len(self._domain_ip_map.keys()),
-            }
-        }
-        add_fields_to(event, event_dbg, overwrite_target=True)
+            return FailedResult(FailureType.UNKNOWN)
+        except NoAnswer:
+            return FailedResult(FailureType.NO_ANSWER)
```
mhoff, line 324: Why do we only communicate the error via `FailedResult` for `NO_NAMESERVERS` and not the others?

tests/unit/processor/domain_resolver/test_domain_resolver.py
```diff
@@ -1,20 +1,30 @@
...
 from logprep.factory import Factory
-from logprep.processor.base.exceptions import FieldExistsWarning
-from logprep.processor.domain_resolver.processor import ResolveStatus
 from tests.unit.processor.base import BaseProcessorTestCase
 
-REL_TLD_LIST_PATH = "tests/testdata/external/public_suffix_list.dat"
-TLD_LIST = f"file://{Path().absolute().joinpath(REL_TLD_LIST_PATH).as_posix()}"
-
 
 class TestDomainResolver(BaseProcessorTestCase):
```
mhoff, line 23: `test_cases` and `example_test_cases` would be great to give the users proper (tested) usage examples. See the `test_generic_resolver.py` for reference. The happy path usage is pretty straight-forward, so this should be rather a small addendum

mhoff (commented):
Hello @ppcad, many thanks! I have added a few remarks

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
...
 
 ### Features
 * timestamper: support generating timestamps from the current time when no `source_fields` are configured
 
 ### Improvements
 * timestamper: validate `source_format` and `source_timezone` according to the configured `source_fields`
+* add metrics for cached timed-out domains and successfully resolved domains to `domain_resolver`
+* make cache hold a payload in addition to timestamps and adapt `domain_resolver` accordingly
```
mhoff: Please consider starting your changes with `domain_resolver: ` as you can see it below. This is a pattern which we slowly adopt currently. Also, please put a prefix in front of the PR title "feat: "

logprep/ng/processor/domain_resolver/processor.py
ppcad: Do you want to remove just this part or the whole security best practices section?

mhoff: Only the single line. I have added this kind of remark at other places as well. I always meant only the single empty line

logprep/ng/processor/domain_resolver/processor.py
ppcad: `resolved_domains` are successfully resolved domains not from the cache. `resolved_new` are domains for which the resolve method was called

logprep/ng/processor/domain_resolver/processor.py
ppcad: Right now it is working how it has worked before the recent chages. `resolved_new` are basically attempts to resolve domains irrespective of the result. It could have been renamed to `resolve_attempts`, but I did not want to change old metric names

logprep/ng/processor/domain_resolver/processor.py
ppcad: In those cases it is cached in the domain cache, since it is assumed to not be a temporary error. Temporary errors will be put inside the `timeout_cache`.

logprep/ng/processor/domain_resolver/processor.py
ppcad: The other exception can commonly happen when resolving domains and I would not like them to spam the logs. The `NO_NAMESERVERS` is rarer and could also mean something is misconfigured, so I would like to see it as a warning.

logprep/ng/processor/domain_resolver/processor.py
ppcad: I have changed it, but for some reason it's not showing as outdated.

logprep/ng/processor/domain_resolver/processor.py
ppcad: I've changed it, but it's not showing up as outdated for some reason.

logprep/ng/processor/domain_resolver/processor.py
mhoff: `cache_max_timedelta` is still set outside, but I guess that's personal preference. Resolved

logprep/ng/processor/domain_resolver/processor.py
mhoff: I see

logprep/ng/processor/domain_resolver/processor.py
ppcad: The only missing case is `FailedResult(_)` without `error`, which I delibratly omited, since those can happen frequently (timeout, unknown domain) and they have metrics.

logprep/util/cache.py
ppcad: I've changed it.

logprep/util/cache.py
ppcad: I've changed it.

logprep/util/cache.py
ppcad: I've changed it.

logprep/ng/processor/domain_resolver/processor.py
ppcad: I've changed it.

logprep/util/cache.py
ppcad: I've changed it and did not find anything else that is affected.

tests/unit/processor/domain_resolver/test_domain_resolver.py
ppcad: I have added it and I have also ensured that `source_fields` must be a list with exactly one element.

PR #1026, opened 2026-09-07
feat: make deduplication controllable

Pablu23:
## Description

Make deduplication possible for field_manager and inheriting processors that can actually use it

## Assignee

### Documentation

### Code Quality

Run tests

## Reviewer

---
<a name="readthedocs-preview"></a>The rendered docs for this PR can be found [here](https://logprep--1026.org.readthedocs.build/).


logprep/ng/processor/decoder/processor.py
```diff
@@ -66,6 +66,10 @@ def transform_values(
         decoder = DECODERS[decoder_rule.source_format]
         return self._decode(event, decoder_rule, decoder, source_field_values)
 
+    @override
+    def is_processable_value(self, value: FieldValue) -> bool:
+        return super().is_processable_value(value) and value != "" and value != []
```
krempel-4711: What about `{}`? Also consider adding a utility for this kind of check in the utils.

logprep/ng/processor/field_manager/processor.py
```diff
@@ -44,117 +43,149 @@
...
         duplicates = []
         ordered_flatten_list = []
         flat_source_fields = self._get_flatten_source_fields(source_fields_values)
         for field_value in flat_source_fields:
-            if field_value not in duplicates:
-                duplicates.append(field_value)
-                ordered_flatten_list.append(field_value)
+            if deduplicate and field_value in duplicates:
+                continue
+            duplicates.append(field_value)
```
krempel-4711: Name of variable is misleading should be something like `present_keys` (they are only duplicates if they are encoutered multiple times) also this should be a set for performance reasons.

logprep/ng/processor/field_manager/processor.py
```diff
@@ -44,117 +43,149 @@
...
                 error,
-                failure_tags=[f"_{self.rule_class.rule_type}_missing_field_warning"],
+                failure_tags=[f"_{rule.rule_type}_missing_field_warning"],
             )
             return True
         return False
 
     @staticmethod
-    def _get_field_values(event, source: Iterable) -> list:
+    def _get_field_values(event: dict[str, FieldValue], source: list[str]) -> list[FieldValue]:
```
krempel-4711: Could use `Iterable[str]` here as a parameter type as any iterable will do.

logprep/ng/processor/field_manager/processor.py
```diff
@@ -44,117 +43,149 @@
...
         return [get_dotted_field_value(event, source_field) for source_field in source]
 
-    def _get_missing_fields_error(
-        self, source_fields: Iterable, field_values: Iterable
-    ) -> Exception:
+    def _get_missing_fields_error(self, source_fields, field_values):
```
krempel-4711: Missing type annotations

logprep/ng/processor/field_manager/processor.py
```diff
@@ -44,117 +43,149 @@
...
         missing_fields = [key for key, value in zip(source_fields, field_values) if value is None]
         return Exception(f"{self.name}: missing source_fields: {missing_fields}")
 
     @staticmethod
-    def _get_flatten_source_fields(source_fields_values: Iterable) -> list:
+    def _get_flatten_source_fields(source_fields_values: list[FieldValue]) -> list[FieldValue]:
```
krempel-4711: See https://github.com/fkie-cad/Logprep/pull/1026/changes#r4023534128

logprep/processor/decoder/processor.py
```diff
@@ -66,6 +66,10 @@ def transform_values(
         decoder = DECODERS[decoder_rule.source_format]
         return self._decode(event, decoder_rule, decoder, source_field_values)
 
+    @override
+    def is_processable_value(self, value: FieldValue) -> bool:
+        return super().is_processable_value(value) and value != "" and value != []
```
krempel-4711: see https://github.com/fkie-cad/Logprep/pull/1026/changes#r4023452822

logprep/processor/key_checker/processor.py
```diff
@@ -24,36 +24,30 @@
...
             return
 
-        output_value = get_dotted_field_value(event, rule.target_field)
+        existing_value: FieldValue = get_dotted_field_value(event, rule.target_field)
 
-        if isinstance(output_value, Iterable):
-            output_value = list({*not_existing_fields, *output_value})
-        else:
-            output_value = not_existing_fields
+        if isinstance(existing_value, list) and is_list_of(existing_value, str):
```
krempel-4711, line 50: General question couldn't we move the `isinstance(x, list)` check into the `is_list_of` utility? I know this is not in scope for this PR, but wanted to mention.

krempel-4711 (commented):
Thanks for the PR. Here are some remarks.

logprep/ng/processor/decoder/processor.py
Pablu23: Good question, should probably be filtered aswell, adding this as a utility kind of defeats the purpose of making this configurable per processor whats allowed and what not, Id say. But please do give me counter arguments :)

logprep/ng/processor/field_manager/processor.py
Pablu23: Will rename, making this a set, would not preserve the duplicates though so thats not possible or do you mean something else?

logprep/ng/processor/field_manager/processor.py
Pablu23: I think there was a reason why I made it into a list, but ill check it again

logprep/processor/key_checker/processor.py
Pablu23: we probably could, on the otherhand, if we already know we got a list on our hands, we can skip that isinstance check, which we couldnt if we put it into the helper method itself

logprep/ng/processor/field_manager/processor.py
krempel-4711: currently `duplicate`is only used to check if a `field_value` already occured. The duplicate check is just checking if it contains a `field_value` has been seen before. It is not relevent how often the value appeard before (it is also not tracked currently). As a set it can still serve this purpuse while the contains check would have less complexity.

logprep/ng/processor/decoder/processor.py
krempel-4711: If `value != "" and value != [] and value != {}` is a common thing to check making a helper for this would be semantically clearer and the code less redunant. If this is only ever relevent in this place (and the corresponding non-ng) forget about the helper.

logprep/ng/processor/field_manager/processor.py
krempel-4711: from the usage i can see here `Iterable[str]` should be sufficient. When it comes to the `list` type using that as a parameter type can become tricky when it comes to inheritence as list are invariant considering their item type. Probably not to relevant here.

logprep/processor/key_checker/processor.py
krempel-4711: Check that, the util got 4 usages where on 2 the list is already known. So I guess one could introduce an additional util and name the current one to `has_item_type`, but then again the usage count is so low we could skip that for now.

logprep/ng/processor/field_manager/processor.py
Pablu23: I tried this and we actually cant make this a set, because dict is a valid FieldValue, and dict is not hashable, so it cant go into a set / get checked if it is in a set. We could do a bigger code change here, but I dont think the performance gains are worth it

krempel-4711: Thought this was Strings only.

krempel-4711 (approved):

